PR #11, opened 2017-05-10
Change time_per_step from Duration to Seconds

### LINKED ISSUE #10: time_per_step() should return f64, not Duration.
Currently, the `Estimate::time_per_step` returns `Duration`. This limits the precision and range that time per step can represent – `Duration` always stores multiples of nanosecods. Moreover, if your step takes less than nanosecond, time per step would be 0, and the ETA would be always 00:00:00. This is how I discovered the bug.

TLDR: Here is list of speeds that indicatif currently supports: :)
…, ¼GB/s, ⅓GB/s, ½GB/s, 1GB/s, ∞YB/s.

krdln:
Fixes #10 
Works for me. I hope it haven't introduced any bugs..

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -9,7 +9,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use parking_lot::RwLock;
 
 use console::{Term, Style, measure_text_width};
-use utils::{expand_template, Estimate, secs_to_duration};
+use utils::{expand_template, Estimate, Seconds};
 use format::{FormattedDuration, HumanDuration, HumanBytes};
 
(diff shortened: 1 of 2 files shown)


krdln:
Now I realised it's a pub function, but the `ProgressState` is not exported anywere, so it doesn't breaks the API. Am I right?

Anyway, the docs and function name should indicate that it returns seconds. Would changing the name to `avg_secs_per_state` be ok, or is that weird? Or maybe add `struct Seconds(pub f64)` to utils?

krdln:
I took an attemt to create a `Seconds` newtype. I wan't be sad if you won't like this idea, though :)

PR #51, opened 2018-05-16
adds wrapper for reader

pacman82:
Should help with issue #8. Sorry about the formatting changes, I had format on save enabled. If you mind I can redo the PR. Looking forward to your feedback!

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -702,6 +702,24 @@ impl ProgressBar {
         ProgressBarIter { bar: self, it: it }
     }
 
+    /// Wraps a Reader with the progress bar.
+    ///
+    /// ```rust,norun
+    /// # use std::fs::File;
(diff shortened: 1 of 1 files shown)


otavio:
This is indeed nice. I wonder if it belongs here or not. I did basically the same thing[1] outside the library code.

1. https://github.com/otavio/rsget/blob/98513cec30f068f0704ea9e5bff9c4088aabb764/src/main.rs#L28

It would be nice if @mitsuhiko could comment here to say if a wrapper is desired or not.

masklinn:
@pacman82 it looks like you put the file through rustfmt or something, leading to a much noisier diff than necessary. Maybe reverting all the non-Read stuff would help?

pacman82:
Cleaned up the formatting noise, and rebased the commit.

PR #141, opened 2019-12-15
Don't panic when the ProgressDrawTargetKind::Remote channel is unavai…

### LINKED ISSUE #140: Panic when ProgressBar outlives MultiProgress
The code below panics due to `pb.clone()` at the end. Using just `pb` doesn't trigger the issue.
```rust
fn main() {
    mpb.add(pb.clone());
}
```

The underlying reason is in `apply_draw_state` (called by ProgressState's Drop impl):
```rust
ProgressDrawTargetKind::Remote(idx, ref chan) => {
    chan.lock().unwrap().send((idx, draw_state)).unwrap();
}
```
The code above works without any issues when a ProgressBar gets dropped before a MultiProgress, but changing that order makes things break because then `send` returns an error as the receiver - the now dropped MultiProgress - is unavailable.

mibac138:
…lable

Includes a minor formatting change in a test (rustfmt).

Fixes #140

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -165,7 +165,11 @@ impl ProgressDrawTarget {
                 }
             }
             ProgressDrawTargetKind::Remote(idx, ref chan) => {
-                chan.lock().unwrap().send((idx, draw_state)).unwrap();
+                return chan
+                    .lock()
+                    .unwrap()
+                    .send((idx, draw_state))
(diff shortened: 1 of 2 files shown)


mmstick:
Ran across this issue today, and this does fix the core dump when adding progress bars dynamically to a multibar that's joining in the background.

PR #160, opened 2020-03-24
Steady tick should check if the update interval has changed

### LINKED ISSUE #159: enable_steady_tick doesn't update already running thread
When calling `enable_steady_tick` and then calling it again, the update interval doesn't change. The example below updates only every 3s despite calling `enable_steady_tick(1000)`.

```rust

fn main() {
    let pb = ProgressBar::new(1).with_style(ProgressStyle::default_bar().template("{elapsed_precise}"));
    pb.enable_steady_tick(3000);
    pb.enable_steady_tick(1000);
    loop {}
}
```

mibac138:
Fixes #159


djc:
Thanks!

PR #162, opened 2020-04-03
Move Instant from ProgressDrawState to ProgressDrawTargetKind::Term

mibac138:
I believe it makes more sense for `Instant` to be stored in `ProgressDrawTargetKind::Term`. The `Instant` is used for checking the rate of drawing only. Now, when it is stored in the draw target it is more accurate (as it actually checks the time of the last draw instead of the time when an update was crated) and reduces memory usage.


djc:
This looks nice, would you mind rebasing it?

djc:
Thanks!

PR #169, opened 2020-04-16
Reduce ProgressStyle size

mibac138:
Changed ProgressStyle's fields from Vec to Box<[]> as they aren't
modified anyway.


mibac138:
Implemented in #174

PR #178, opened 2020-06-05
Implement `insert` method for `MultiProgress`

nlinker:
Fixes https://github.com/mitsuhiko/indicatif/issues/177

out

### DIFF
--- a/examples/multi-tree.rs
+++ b/examples/multi-tree.rs
@@ -1,0 +1,151 @@
+use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
+use lazy_static::lazy_static;
+use std::sync::{Arc, Mutex};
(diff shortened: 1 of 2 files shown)


examples/multi-tree.rs
```diff
@@ -0,0 +1,151 @@
...
+use std::time::Duration;
+
+use std::fmt::Debug;
+
+#[derive(Clone, Debug)]
+struct Rng(u64);
+
+impl Default for Rng {
+    fn default() -> Self {
+        Self(14312312512314u64)
```
ip1981, line 14: Huh?

nlinker: This is just to avoid `rand` dependency. You propose to get ms since epoch, right?

PR #181, opened 2020-06-25
Remove unsafe sync on MultiProcess

gshuflin:
This commit introduces some uses of `Mutex` similar to https://github.com/mitsuhiko/indicatif/pull/25 , in order to make it possible to remove the `unsafe sync` implementation on the `MultiProcess` type.


djc:
There's no longer any unsafe in the code base (and the code has changed quite a lot).

PR #183, opened 2020-07-07
Fix bars with fine detail progress

dnaka91:
I noticed that the `finebars` example was broken due to some precision cutoff in v0.15.0. This is visible as not seeing any fine animations anymore and instead only seeing the very first character being used in the animation.

This rolls back the cast `fill as usize` to fix the display of any bars with fine progress animation.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -148,11 +148,10 @@ impl ProgressStyle {
         let width = width / state.style.char_width;
         let pct = state.fraction();
         let fill = pct * width as f32;
-        let fill = fill as usize;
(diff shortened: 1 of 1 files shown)


djc:
Would you mind rebasing this?

dnaka91:
Hey @djc thanks for looking into this, I thought it's been forgotten. I just checked out the main branch and it seems the latest changes have already fixed this.

So this change is actually not needed anymore 🙇

djc:
Thanks for checking!

PR #186, opened 2020-07-13
Minor improvement of ProgressStyle::template docs

mibac138:


src/style.rs
```diff
@@ -110,7 +110,8 @@ impl ProgressStyle {
         self
     }
 
-    /// Sets the template string for the progress bar.
+    /// Sets the template string for the progress bar. List of placeholders is available at crate
+    /// root docs.
```
djc, line 114: I find the term `placeholders` pretty confusing for this. The documentation seems to call them `keys`. Also, the second string here should be split out onto a separate line, with an empty line between the first and last lines.

djc:
Thanks!

Issue #187, opened 2020-07-21
Even if all ProgressBar are drop, MultiProgress doesn't join


Stargateur:
Like multiprogress must contain a strong arc to progress bar, if my code panic, I don't call finish and so I deadlock.

Is there a way to fix this ? I will do a wrapper to call finish on drop but that could be nice to have it in this crate PrograssBarFinish. Or better multiprogress could see that all progress bar have been drop.

mibac138:
Are you sure you aren't cloning any `ProgressBar`? If you are, can you provide a code sample demonstrating this issue? As long as you don't, the `ProgressBar` will finish on drop unless you've set `draw_delta` to anything other than `0` and during dropping the position hasn't advanced enough since the last draw. You can see the relevant code [here](https://github.com/mitsuhiko/indicatif/blob/main/src/progress.rs#L834-L846). The `draw_state` function call *will* notify the `MultiProgress` it's been dropped which the `MultiProgress` then takes into account when determining if it should continue running.

Stargateur:
that indeed the case I use `set_draw_delta(count / 100)` there is a panic at 0 so yeah didn't progress much. So this will deadlock if we use `set_draw_delta` not cool.

I did my wrapper work great:

```rust
#[cfg(feature = "indicatif")]
pub struct ProgressBarFinish(ProgressBar);

#[cfg(feature = "indicatif")]
impl ProgressBarFinish {
    fn new(count: u64) -> Self {
        Self(ProgressBar::new(count))
    }
}

#[cfg(feature = "indicatif")]
impl Drop for ProgressBarFinish {
    fn drop(&mut self) {
        self.finish()
    }
}

#[cfg(feature = "indicatif")]
impl Deref for ProgressBarFinish {
    type Target = ProgressBar;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

mibac138:
I *think* ignoring `draw_delta` when dropping should be okay as it's purpose is to limit the `ProgressBar`'s performance overhead in situations when the `ProgressBar` is very frequently updated (see #59 for motivation). One last draw call when dropping should be insignificant. I will submit a PR to fix this, possibly surprising, behavior.

Stargateur:
I agree too, thank for your help.

djc:
#188 has been merged, I think this is fixed now.

mibac138:
@djc I believe this can be closed now, yes.

PR #188, opened 2020-07-21
ProgressBar drop should ignore draw delta

mibac138:
See #187 for context
cc @Stargateur


djc:
Would be happy to merge this if you can rebase it.

djc:
Thanks!

PR #185, opened 2020-07-13
Implement `remove` method for `MultiProgress`

nlinker:
out

Fixes https://github.com/mitsuhiko/indicatif/issues/184

PR #192, opened 2020-08-09
Fix `ProgressBar::finish_and_clear` messing up other progress bars order

### LINKED ISSUE #191: finish_and_clear sometimes messes up the order of other progress bars
Changing line [`mp2.remove(pb)`](https://github.com/mitsuhiko/indicatif/pull/185/files#diff-12b2905be9b9da2900dd186c69e4e82eR140) to `pb.finish_and_clear()` messes up the order of the progress bars. In my opinion `MultiProgress::remove` should work the same as `ProgressBar::finish_and_clear`.

mibac138:
Fixes #191 

I've failed to create a test case, however rebasing this PR onto #185 and changing [this line](https://github.com/mitsuhiko/indicatif/pull/185/files#diff-12b2905be9b9da2900dd186c69e4e82eR140) to `pb.finish_and_clear` can be used to test this change.


src/progress.rs
```diff
@@ -1056,6 +1056,12 @@ impl MultiProgress {
             let mut state = self.state.write().unwrap();
             if draw_state.finished {
                 state.objects[idx].done = true;
+                if draw_state.lines.is_empty() {
+                    // `finish_and_clear` was called
+                    if let Some(order_idx) = state.ordering.iter().position(|x| *x == idx) {
+                        state.ordering.remove(order_idx);
+                    }
```
djc, line 1063: We should probably use `state.ordering.retain()` here. Would you be willing to rebase this?

PR #198, opened 2020-08-24
fix var match

darkskygit:
this pr fix var match in progress style like this:
before:
image
after:
image


mibac138:
I believe you're supposed to escape curly braces using double curly braces. So if you want the template to literally output `{` you should write `{{` in the template, so your example would become: `{{ pos: {pos}, len: {len} }}`. I feel like this was documented somewhere but I can't find it now. Regardless, I think this is an improvement as there aren't any variables using braces internally and I don't expect there to be any.

djc:
Would be good to add a test for this, which doesn't seem too onerous. Can you do that?

djc:
Thanks!

PR #200, opened 2020-09-25
Implement duration (eta + elapsed) for formatting

akx:
This seemed like an obvious, small improvement – tqdm shows the total extrapolated duration, why couldn't indicatif?


djc:
Makes sense, thanks!

PR #202, opened 2020-09-26
Add indexed parallel iterator support

drivasperez:
Hi,

Rayon exposes a supertrait of `ParallelIterator`, `IndexedParallelIterator`, which has a `len` method. This PR adds a new progress trait, `IndexedParallelProgressIterator`, that wraps `IndexedParallelIterator` and makes the `progress` method work in the same way `ProgressIterator` does with `Iterator`'s `size_hint` method, so we don't have to rely on `progress_count`.

I'm pretty new to working with Rust, and I'm not sure this is the best way of going about this. I tried to add an impl of `ParallelProgressIterator` for `IndexedParallelIterator`, but the compiler complained that the implementations were ambiguous because an `IndexedParallelIterator` is also an `ParallelIterator`. Maybe the extra trait makes this API worse than just getting people to use `progress_count`.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -56,8 +56,10 @@ impl<S, T: Iterator<Item = S>> ProgressIterator for T {
 pub mod rayon_support {
     use super::*;
     use rayon::iter::{
-        plumbing::Consumer, plumbing::Folder, plumbing::UnindexedConsumer, ParallelIterator,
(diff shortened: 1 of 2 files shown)


LucaCappelletti94:
Hi @drivasperez, could you possibly add a usage example? Thanks!

djc:
Makes sense to me, thanks.

PR #213, opened 2020-11-15
Let println() take ToString instead of Into<String>

Beiri22:
It allows for arguments of type Display, str and String -> like normal println() would.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -517,10 +517,10 @@ impl ProgressBar {
     /// Note that if the progress bar is hidden (which by default happens if
     /// the progress bar is redirected into a file) println will not do
     /// anything either.
-    pub fn println<I: Into<String>>(&self, msg: I) {
+    pub fn println<I: ToString>(&self, msg: I) {
         let mut state = self.state.write().unwrap();
 
-        let mut lines: Vec<String> = msg.into().lines().map(Into::into).collect();
+        let mut lines: Vec<String> = msg.to_string().lines().map(Into::into).collect();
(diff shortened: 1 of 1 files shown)


vallentin:
In this case it's actually better to have it remain as `Into<String>` instead of changing it to `ToString`. Because internally `ProgressDrawState` needs to own the strings (`lines: Vec<String>`).

If you call `pb.println(String::from("Hello World"))`, then with `Into<String>` then the `String` is moved. However if you change it to `ToString` then the string is needlessly cloned.

While yes, `ToString` does allow passing types which `impl Display`, but I think it isn't worth the needless clone.

Now, what would be better, would be to change it to `AsRef<str>`. Because `msg.into()` also results in an additional needless clone for `pb.println("Hello World")`, as it gets converted into a `String` which is immediately discarded after `collect()`.

Beiri22:
What about a second method for easy printing of Display-objects?

vallentin:
Just to be clear, I'm not a maintainer of this library, I just wanted to share my 2 cents.

Personally, I'd say yeah. Having a `println`/`println_with_display` (or `println_str`/`println`) combo would be nice.

Though it might be easier just to implement that locally with an extension trait, e.g.:

```rust
trait ProgressBarExt {
    fn println_with_display<I: ToString>(&self, msg: I):
}

impl ProgressBarExt for ProgressBar {
    fn println_with_display<I: ToString>(&self, msg: I) {
        self.println(msg.to_string());
    }
}
```

PR #221, opened 2020-12-21
Fix example download-speed on 80 width terminal

pickfire:
Screenshot_20201221_224824


pickfire:
cc @kpcyrd

kpcyrd (approved):

djc:
Makes sense, thanks!

PR #229, opened 2021-01-02
Refactor to remove a bit of duplicated logic from MultiProgress

marienz:
Although implemented differently, the logic for setting "finished" in join_impl
is identical to that in is_done.

Move is_done from MultiProgress to MultiProgressState so we can reuse it when
setting "finished", keeping the implementation that is easier to follow.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -956,6 +956,20 @@ struct MultiProgressState {
     move_cursor: bool,
 }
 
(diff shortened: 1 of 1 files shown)


djc:
If you want to rebase this, I'll merge it!

marienz:
Done!

djc:
Thanks!

Issue #205, opened 2020-10-09
Add println method to MultiProgress


mdekstrand:
Right now, the ProgressBar has a println object, but MultiProgress does not.

I have created the [`happylog`](https://crates.io/crates/happylog) package that provides a backend to `log` for console-focused logging, and allows the program to reroute logging output to a progress bar's `println` to keep log output and progress updates from interfering with each other. In this use case, it would fit better for `MultiProgress` users to be able to reroute logging to the entire `MultiProgress`, instead of needing to pick one of the progress bars to route it to.

marienz:
I just found indicatif and thought I needed println() on MultiProgress for much the same reason you do, and took a stab at implementing it. See https://github.com/marienz/indicatif/tree/multiprogress-println (try `cargo run --example multi-tree` to see it in action, and note the diff is much smaller with whitespace differences ignored).

There are some gotchas in using this, which is why I haven't sent a pull request yet:

1. Calling `println()` before calling `join()` doesn't print (it doesn't block, the message gets buffered and printed once `join()` is called). This is really the same caveat that already applies to progress bars, so maybe not too much of an issue: it means the MultiProgress should be registered as a logging backend immediately before calling `join()`. Note the warnings about "will deadlock your program" in the docstrings are overstating things, as far as I can tell, I'll probably tweak that wording separately.
2. Calling `join()` before adding any progress bars returns immediately. This isn't new, but it does mean any code running into `println()` not working before `join()` can't necessarily fix it by moving `join()` earlier.
3. Calling `println()` after the last progress bar finishes drops messages (they're buffered but never read). I think this is a problem for use cases like yours: I don't think there's a race-free way to unregister the MultiProgress as a logging backend after it has stopped drawing, but before it might have buffered a message. I've added a method to retrieve buffered messages that can be used to deal with this (`buffered_messages()`, also added to the multi-tree example).

I don't see a good (that is: not racy) way of making `println()` work before or after `join()`. It's fairly straightforward to make `join()` block until at least one progress bar is added, but that would be a change in behavior for code not using the new `println()` so I don't think it's a good idea. These two caveats combined do mean this new functionality is not quite as useful for happylog as you might have hoped: your users will still need to add at least one progress bar and call  `join()` close to where they reroute logging (they can't just create a MultiProgress at the start of their program and reroute logging without actually using the MultiProgress yet).

[...]

PR #231, opened 2021-01-03
[RFC] Remove MultiProgress::join()

marienz:
This should not be merged yet, but I'd like to know if this looks like something worth pursuing.

This change moves drawing of MultiProgress from the join() thread to the child ProgressBar updates.

This has the following advantages:

- It addresses #125 (async MultiProgress), as long as users are willing to assume the writes to the screen do not block (and they do not add enough child progress bars for lock contention on the MultiProgress state to be an issue). It also makes it trivial to use MultiProgress without any threads or async runtime at all.
- It helps with #205 (add println to MultiProgress) because println on MultiProgress can now be made to work while the MultiProgress does not have any unfinished progress bars / join() hasn't been called yet.
- Arguably the code is simpler (and it removes the one use of `unsafe` in indicatif: MultiProgress is now automatically `Sync`).

It also has disadvantages, most of which I think can be mitigated:

- It removes grouping, which regresses the problem fixed by #166. I think it is possible to largely mitigate this by implementing a draw target that uses a timer (via an async runtime or a thread) to schedule a draw `rate - last_draw.elapsed()` in the future if we don't want to draw immediately. I haven't written the code for this yet, though, so it might not be as easy as I think...
- It is backwards-incompatible. This could be mitigated by implementing this as a new type (and a new `ProgressDrawTargetKind`, which is fine as those are not public) if the old behavior is deemed useful enough to keep around and/or we want to avoid the semver bump.
- It removes the ability to use the MultiProgress as a synchronization point. For the examples, this was very easy to fix by joining the worker threads instead. I don't know if this is also true for "real" code.
- It moves work from the join() thread to the threads updating the progress bars. I suspect this is not a problem because that work is already fast enough (relative to rendering the child progress bar to a ProgressDrawState, which was already happening on the worker threads), but I haven't benchmarked it. It might be a problem for code that has at least one progress bar that is updated at a very high rate.

I think the DrawTarget-with-a-timer is useful anyway: it should mitigate a problem very similar to #166 when doing something like:
```rust
use std::thread;
use std::time::Duration;

use indicatif::ProgressBar;

fn main() {
    let p = ProgressBar::new_spinner();
    for _ in 0..10 {
        p.set_message("performing fast work (you should rarely see this)");
        // Just to demonstrate they don't need to be back-to-back. Removing this sleep does not change behavior.
        thread::sleep(Duration::from_millis(5));
        p.set_message("performing slow work (you should see this)");
    }
}
```
This currently only shows "you should not see this" for the same reason the MultiProgress example didn't stay in sync. With a DrawTarget-with-a-timer, it would show "you should not see this" for only one update interval. So I'll probably see if I can get that to work even if this change isn't deemed useful.

Please let me know what you think!

### DIFF
--- a/examples/finebars.rs
+++ b/examples/finebars.rs
@@ -15,24 +15,29 @@ fn main() {
 
     let m = MultiProgress::new();
 
-    for s in styles.iter() {
-        let pb = m.add(ProgressBar::new(512));
-        pb.set_style(
-            ProgressStyle::default_bar()
-                .template(&format!("{{prefix:.bold}}▕{{bar:.{}}}▏{{msg}}", s.2))
-                .progress_chars(s.1),
-        );
-        pb.set_prefix(s.0);
-        let wait = Duration::from_millis(thread_rng().gen_range(10, 30));
-        thread::spawn(move || {
-            for i in 0..512 {
-                pb.inc(1);
-                pb.set_message(&format!("{:3}%", 100 * i / 512));
-                thread::sleep(wait);
-            }
-            pb.finish_with_message("100%");
-        });
-    }
+    let handles: Vec<_> = styles
+        .iter()
+        .map(|s| {
+            let pb = m.add(ProgressBar::new(512));
+            pb.set_style(
+                ProgressStyle::default_bar()
+                    .template(&format!("{{prefix:.bold}}▕{{bar:.{}}}▏{{msg}}", s.2))
+                    .progress_chars(s.1),
+            );
+            pb.set_prefix(s.0);
+            let wait = Duration::from_millis(thread_rng().gen_range(10, 30));
+            thread::spawn(move || {
+                for i in 0..512 {
+                    pb.inc(1);
+                    pb.set_message(&format!("{:3}%", 100 * i / 512));
+                    thread::sleep(wait);
+                }
+                pb.finish_with_message("100%");
+            })
+        })
+        .collect();
 
-    m.join().unwrap();
+    for h in handles {
+        let _ = h.join();
+    }
 }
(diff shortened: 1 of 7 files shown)


mdekstrand:
My main question on this direction is whether it creates new problems for scoping of the MultiProgress and threads. I don't know that it does, but also don't currently know that it doesn't.

marienz:
I'm not sure if I understand the question about scoping. Can you give an example you're concerned about?

To port to the new code, all you should have to do is remove the call to `join()` and make sure this does not cause the program to exit too soon (you may have to wait for worker threads explicitly to do so). It is ok if dropping the call to `join()` causes the MultiProgress to be dropped before the worker threads are done with their ProgressBars: any ProgressBars already added to the MultiProgress before it was dropped will continue to work correctly, you just can't add any additional ones without a reference to the MultiProgress.

marienz:
#234 adds the promised DrawTarget-with-a-timer.

jackinloadup:
This branch solved a problem I was having with my application structure. Thanks for the work!

djc:
Sorry for taking so long to get to this. I do think this idea makes sense. Would you be willing to rebase it on top of current main? It looks like some of the structural changes that you made are already part of main now, but there have been quite a few changes, so the rebase might be a little painful.

aj-bagwell:
I like this too, I have rebased @marienz's changes onto main and added a leaky bucket to the rate limiting to fix the regression on #166 in [aj-bagwell:multiprogress-unjoined](https://github.com/mitsuhiko/indicatif/compare/main...aj-bagwell:multiprogress-unjoined) if that helps keep this moving.

djc:
Sounds great, would you mind submitting a separate PR for this?

aj-bagwell:
@djc sure, new pull request is #284

djc:
Okay, I'll close this for now.

PR #232, opened 2021-01-07
Add the ability to set a redraw rate on progress bar

ithinuel:
I was having performance issues where a 20sec process without progress bar would take above 1h with it.
`ProgressDrawTarget::stderr_with_hz(25)` did not seem to have any effect.

`ProgressBar::set_draw_delta` showed a significant impact but my input can vary from few Kb to several Gb and picking an arbitrary % for the update isn't an option. 

This is what solves the issue for me.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -230,6 +230,7 @@ pub(crate) struct ProgressState {
     message: String,
     prefix: String,
     draw_delta: u64,
+    draw_rate: u64,
     draw_next: u64,
     status: Status,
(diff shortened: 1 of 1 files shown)


djc:
LGTM, thanks!

mibac138:
@ithinuel Have you used `progress.println` in your code?

ithinuel:
I did not at the time. Since I moved to a “lazy processing” so I no longer need a progress bar/spinner in my app.

PR #234, opened 2021-01-09
[RFC] Threaded draw

marienz:
This PR could be merged by itself, but it makes more sense when combined with #231, which removes batching from MultiProgress.

The new code is slower for code that updates very frequently, because the channel send is not cheap enough. I'm working on some optimizations of format_state to balance that out.

I'm not sure how much the change in behavior for code that does not finish its progress bars affects real code.

### DIFF
--- a/examples/multi-synced.rs
+++ b/examples/multi-synced.rs
@@ -1,0 +1,27 @@
+// From https://github.com/mitsuhiko/indicatif/pull/166
+
+use std::thread;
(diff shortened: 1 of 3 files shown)


chris-laplante:
@djc, @marienz indicatif multi no longer uses channels, so I think this PR may be obsolete?

djc:
Yes, I think we no longer need this.

Issue #219, opened 2020-12-17
Difference between rayon and non-rayon progress bar, rayon erases progress bar at the end


mj-64:
Consider this example:

```
use indicatif::{ParallelProgressIterator, ProgressBar, ProgressIterator, ProgressStyle};
use rayon::prelude::*;

fn task(i: isize) -> isize {
    std::thread::sleep(std::time::Duration::from_millis(50));
    i * i
}

fn main() {
    let progress_bar = ProgressBar::new(0);
    progress_bar.set_style(
        ProgressStyle::default_bar()
            .template(
                "{spinner:.green} [{elapsed_precise:.green}] [{eta_precise:.cyan}] {msg:.magenta} ({percent:.bold}%) [{bar:30.cyan/blue}]",
            )
            .progress_chars("█░"),
    );
    progress_bar.set_length(1000);
    let r: Vec<isize> = (0..1000isize).into_par_iter().progress_with(progress_bar).map(task).collect();
    println!("\n{}", r.into_iter().sum::<isize>());
}
```

The example with ```...into_par_iter().progress_with(...``` erases (removes) the progress bar when finished.
When I change it to ```...into_iter().progress_with(...```, the progress bar remains visible.

PR #240, opened 2021-02-19
Allow ProgressBarIter to use a progress bar without consuming it

benkay86:
_Edited to use `Borrow` instead of `Deref`._

The `ProgressIterator` trait and corresponding `ProgressBarIter` struct are very useful for adding a progress bar to an iterator.  The currently implementation of `ProgressIterator::progress_with()` takes ownership of/consumes the progress bar.  This limits control over the progress bar after it has been added to the iterator.  For example, in the code below we can't call `pb.finish_and_clear()` to change the default finish behavior with the current implementation.

(In theory we can still access the public field `ProgressBarIter::progress`, but in practice the `ProgressBarIter` will probably be consumed by another chained iterator at which point the field will no longer be accessible.)

Maintaining backwards-compatibility with the current API, I propose makeing `ProgressIterator::progress_with()` generic over `std::borrow::Borrow<ProgressBar>` so that it can consume a `ProgressBar` **or** take a reference `&ProgressBar`, `Box<ProgressBar>`, `Arc<ProgressBar>`, etc.  The user then retains ownership (if using `&ProgressBar`) or a reference/clonable reference (with `Box` or `Arc`) which can be used to modify the behavior of the progress bar even after it has been added to the iterator.  For example, the default behavior is for the progress bar to be `finish()`ed when the iterator is exhausted, but using the generic `progress_with()` below we can call `finish_and_clear()` to clear the bar after we are done with it.

```rust
use indicatif::{MultiProgress, ProgressBar, ProgressIterator, ProgressStyle};
use std::sync::Arc;
use std::thread::sleep;
use std::time::Duration;

fn main() {
    // Create a MultiProgress.
    let multi = Arc::new(MultiProgress::new());

    // Add an overall progress indicator.
    let main_pb = multi.add(ProgressBar::new(5).with_style(
        ProgressStyle::default_bar()
        .template("{msg} [{wide_bar}] {pos}/{len}")
    ));
    main_pb.set_message("Overall");
    main_pb.tick();

    // Spawn the MultiProgress animation thread.
    let multi_animation = {
        let multi = multi.clone();
        std::thread::spawn(move || {
            multi.join()
        })
    };

    // 100 ms delay.
    const DELAY: Duration = Duration::from_millis(100);

    // Create five sequential tasks with progress indicators.
    (0..5).progress_with(main_pb).for_each(|x| {
        // Progress indicator with 10 steps.
        let pb = multi.add(ProgressBar::new(10).with_style(
            ProgressStyle::default_bar()
            .template("{msg} [{wide_bar}] {pos}/{len}")
        ));
        let msg = format!("Task {:<2}", x);
        pb.set_message(&msg);

        // Wrap task in progress bar.  Note passing &pb instead of pb.
        (0..10).progress_with(&pb).for_each(|_| sleep(DELAY));

        // Clear progress bar.
        // Not possible before!!!
        pb.finish_and_clear();
    });

    // Wait for animation to finish.
    multi_animation.join().unwrap().unwrap();
}
```

If you think this kind of change is consistent with the design goals of indicatif then I would be happy to add it to the rayon parallel progress iterators too.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,4 +1,5 @@
 use crate::progress::ProgressBar;
+use std::borrow::Borrow;
 
 /// Wraps an iterator to display its progress.
 pub trait ProgressIterator
@@ -7,7 +8,7 @@ where
 {
     /// Wrap an iterator with default styling. Attempt to guess iterator
     /// length using `Iterator::size_hint`.
-    fn progress(self) -> ProgressBarIter<Self> {
+    fn progress(self) -> ProgressBarIter<Self, ProgressBar> {
         let n = match self.size_hint() {
             (_, Some(n)) => n as u64,
             _ => 0,
@@ -16,30 +17,30 @@ where
     }
 
     /// Wrap an iterator with an explicit element count.
-    fn progress_count(self, len: u64) -> ProgressBarIter<Self> {
+    fn progress_count(self, len: u64) -> ProgressBarIter<Self, ProgressBar> {
         self.progress_with(ProgressBar::new(len))
     }
 
     /// Wrap an iterator with a custom progress bar.
-    fn progress_with(self, progress: ProgressBar) -> ProgressBarIter<Self>;
+    fn progress_with<P: Borrow<ProgressBar>>(self, progress: P) -> ProgressBarIter<Self, P>;
 }
 
 /// Wraps an iterator to display its progress.
-pub struct ProgressBarIter<T> {
+pub struct ProgressBarIter<T, P: Borrow<ProgressBar>> {
     it: T,
-    pub progress: ProgressBar,
+    pub progress: P,
 }
 
-impl<S, T: Iterator<Item = S>> Iterator for ProgressBarIter<T> {
+impl<S, T: Iterator<Item = S>, P: Borrow<ProgressBar>> Iterator for ProgressBarIter<T, P> {
     type Item = S;
 
     fn next(&mut self) -> Option<Self::Item> {
         let next = self.it.next();
 
         if next.is_some() {
-            self.progress.inc(1);
+            self.progress.borrow().inc(1);
         } else {
-            self.progress.finish();
+            self.progress.borrow().finish();
         }
 
         next
@@ -47,7 +48,7 @@ impl<S, T: Iterator<Item = S>> Iterator for ProgressBarIter<T> {
 }
 
 impl<S, T: Iterator<Item = S>> ProgressIterator for T {
-    fn progress_with(self, progress: ProgressBar) -> ProgressBarIter<Self> {
+    fn progress_with<P: Borrow<ProgressBar>>(self, progress: P) -> ProgressBarIter<Self, P> {
         ProgressBarIter { it: self, progress }
     }
 }
@@ -234,7 +235,7 @@ mod test {
     #[test]
     fn it_can_wrap_an_iterator() {
         let v = vec![1, 2, 3];
-        let wrap = |it: ProgressBarIter<_>| {
+        let wrap = |it: ProgressBarIter<_,_>| {
             assert_eq!(it.map(|x| x * 2).collect::<Vec<_>>(), vec![2, 4, 6]);
         };
 
@@ -244,5 +245,11 @@ mod test {
             let pb = ProgressBar::new(v.len() as u64);
             v.iter().progress_with(pb)
         });
+        let pb = ProgressBar::new(v.len() as u64);
+        let it = v.iter().progress_with(&pb);
+        assert_eq!(it.map(|x| x * 2).collect::<Vec<_>>(), vec![2, 4, 6]);
+        let pb = Box::new(ProgressBar::new(v.len() as u64));
+        let it = v.iter().progress_with(pb);
+        assert_eq!(it.map(|x| x * 2).collect::<Vec<_>>(), vec![2, 4, 6]);
     }
 }


djc:
This may not be super obvious, but a `ProgressBar` internally has an `Arc<RwLock<>>` around the actual state, so passing ownership is actually not super meaningful in this case. I have also run into the problem of not being able to pick the right finish behavior, but I think we might this want a differently scoped fix for that problem. (Maybe related to #219?)

benkay86:
> This may not be super obvious, but a ProgressBar internally has an Arc<RwLock<>> around the actual state, so passing ownership is actually not super meaningful in this case.

I had long suspected that `ProgressBar` uses some form of interior mutability for its methods taking `&self` and mutating state, but as you say this `Arc` is _internal_.  I may be overlooking some way of leveraging this in `ProgressBar`'s API, but since `ProgressIterator::progress_with()` takes ownership of `ProgressBar` (not `Arc<ProgressBar>`) I don't see any way to manually tweak the bar after it has been wrapped in a `ProgresBarIter` in a meaningful way, for example, to change the progress style if a warning/error occurs during iteration, to modify the finish behavior, etc.

```rust
pb = ProgressBar::new();
piter = (1..10).progress_with(pb);
// technically can still access pb like this
piter.progress.set_style(/* whatever */);
piter.for_each(|x|
    // can't access pb from in here
);
```

With the proposed backwards-compatible change:

```rust
pb = ProgressBar::new();
piter = (1..10).progress_with(&pb);
// now can access pb directly or piter.progress if we want
piter.for_each(|x|
    // now we can access pb from in here
);
// and even out here
```

> I have also run into the problem of not being able to pick the right finish behavior, but I think we might this want a differently scoped fix for that problem.

Agreed, I had noticed that the `finish()` behavior for `ProgressBarIter` is different from the `finish_and_clear()` behavior in `ProgressBar`'s `Drop` trait as well.  I was going to address that in a different PR, probably by proposing an enum to specify the default finish method on `Drop` or completion of an iterator.

djc:
You can just clone the `ProgressBar` and use the clone to mutate it after setting up the iterator?

djc:
As for the finish behaviour, I'm inclined to think we should have a behavior-agnostic `finish()` method that users like the iter wrappers use, and APIs to configure the desired finishing behavior. That seems more in line with how the rest of the APIs work.

benkay86:
> You can just clone the ProgressBar and use the clone to mutate it after setting up the iterator?

Ah, I see.  `ProgressBar` derives `Clone` and `ProgressBar`'s only member, `state`, is an `Arc`, so a cloned `ProgressBar` actually references the same `ProgressBar` that is was cloned from.  This makes my pull request largely redundant since you could just do the following with the minimal expense of an unnecessary clone in some cases:

```rust
let pb = ProgressBar::new();
let cloned_pb = pb.clone();
(0..10).progress_with(pb).for_each(|x|
    pb_cloned.do_something();
);
pb_cloned.finish_and_clear();
```

Even though this behavior is clearly documented for the struct-level doc comments on `ProgressBar`, it was totally not obvious to me because `ProgressBar::state` is not public, and moreover I had seen examples elsewhere of people using `Arc<ProgressBar>`.  We don't have [arbitrary self types](https://github.com/rust-lang/rust/issues/44874) yet, but I wonder if it would be practical to refactor like this to make the `Arc` explicit in the API?

```rust
pub struct ProgressBar {
    state: RwLock<ProgressState>,
}
impl ProgressBar {
    pub fn new(len: u64) -> Arc<ProgressBar> {
        Arc::new(ProgressBar { state: /* ... */ })
    }
    pub fn tick(self: &Arc<ProgressBar>) {
        self.state.update_and_draw( /* ... */ );
    }
}
```

Alternatively you could override `clone()` like this to make the behavior even more explicit in the documentation:
```rust
impl ProgressBar {
    /// Clone this `ProgressBar`.  The clone references the same bar
    /// as the original.  This is similar to invoking `clone()` on [`std::sync::Arc`].
    pub fn clone(&self) -> Self {
        <Self as Clone>::clone(self)
        // Or drop the #[derive(Clone)] and:
        // ProgressBar { state: self.state.clone() }
    }
}
```

As for this pull request, I don't mind if you close it without merging.  Merging would offer a little more flexibility than the current API without any performance cost, but it would make the code a little more complicated to read.  If you don't merge, it's still possible to do what I want using the special behavior of `clone()`.

djc:
So how would you solve the finish behavior issue with the current `clone()` approach? I was under the impression it wasn't possible.

benkay86:
My apologies, it is _mostly_ possible using the current clone approach, and this PR doesn't substantially change the behavior.  I am so daft I did not realize that `ProgressBar::clone()` has the same clone-by-reference behavior as `Arc::clone()` (as opposed to the behavior of `String::clone()`, which would create a totally new string).  So returning to the very first example in the PR you can do:

```rust
// Create the progress bar.
let pb: ProgressBar = multi.add(...);
// Clone the progress bar.
// The cloned bar actually controls the same bar as `pb`!
let pb_clone = pb.clone();
// Move `pb` into the `ProgressBarIter` struct,
// and do something with the resultant iterator.
(0..10).progress_with(pb).for_each(|_| sleep(DELAY));
// Here `ProgressBarIter` has called `pb.finish()`.
// Then the `ProgressBarIter is dropped so `pb` is also dropped.
// But we still have `pb_clone`, so we can do this:
pb_clone.finish_and_clear();
```

Now the point of this PR was to make `ProgressIterator::progress_with()` receive the progress bar by value/move _or by reference_.  But since `ProgressBar` already has clone-by-reference behavior, this change turns out to be unnecessary.

Now, as a separate PR it might be nice to introduce some API into `ProgressStyle` or else something like `ProgressBar::set_finish(behavior: FinishBehavior)` where `FinishBehavior` is some enum that lets the user choose between finish with message, finish and clear, finish at current position, etc.  And then we could modify the `ProgressIterator` implementation and the `Drop` implementation for `ProgressBar` to do what is says in this enum instead of some fixed, default behavior.  That would save the user from having to keep track of a clone of the progress bar after it has been wrapped in an iterator.

djc:
Right, that's what I was wondering about. Okay, I'm going to close this -- I actually think hiding the `Arc` inside `ProgressBar` is nice, although I'd be happy to take more patches documenting the behavior in places you think should talk about it that currently don't. Would also like to take a patch to set the finish behavior with an API along the lines you describe (sounds good so far).

PR #241, opened 2021-02-27
Properly query the width of a MultiProgress draw target.

stuhood:
The `MultiProgress` draw target was not actually reporting the width of the underlying `ProgressDrawTarget`, and would instead always report the width of `Term::stderr()`.

Because `Term` does not support writing to arbitrary file handles (yet... https://github.com/mitsuhiko/console/pull/93), this issue would have been relatively hard to detect: a `MultiProgress` would need to have been started against `stdout` with a _separate_ TTY on `stderr`. But when used against an arbitrary file handle, the result was that the width fell back to a hardcoded value.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -36,7 +36,11 @@ enum Status {
 
 enum ProgressDrawTargetKind {
     Term(Term, Option<ProgressDrawState>, Option<Duration>),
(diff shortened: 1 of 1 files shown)


djc:
This makes sense to me, thanks!

PR #244, opened 2021-03-10
Add ProgressFinish API

benkay86:
As discussed in a [tangentially related PR](https://github.com/mitsuhiko/indicatif/pull/240), it would be nice to have an API for specifying how a progress bar should be `finish()`ed by default when it is dropped or when an iterator it is wrapping has finished.  This PR proposes such an API by adding a `ProgressFinish` enum to the `ProgressStyle`, which can be used like this:

```rust
let pb = ProgressBar::new().with_style(
    ProgressStyle::default_bar().on_finish(ProgressFinish::AndClear)
);
```

In this example `pb` will automatically have `finish_and_clear()` called on it when it is dropped, unless the user manually calls one of the other `finish()` methods before dropping.

This is especially useful for set-it and forget-it style progress bars used in iterators.  For example:

```rust
const DELAY: Duration = Duration::from_millis(100);
(0..10).progress_with(pb).for_each(|_| sleep(DELAY));
// Don't have to clone pb and call finish_and_clear()
```

Unfortunately I am not familiar with Rayon-style parallel iterators and so have ***not*** attempted to modify the Rayon parallel iterator API to use the new API (I am not even sure if this is possible).  I also have ***not*** attempted to modify `ProgressBarWrap` to automatically finish its progress bar since the `std::io::Read` and `Write` APIs do not provide any unambiguous way to know whether the underlying reader or writer is "finished."

---

Under the hood, this PR:

1. Adds an enum `ProgressFinish` whose variants correspond to the methods `finish()`, `finish_and_clear()`, etc.
2. Adds a field `ProgressStyle::on_finish` to store the desired `ProgressFinish` behavior.
3. Adds `ProgressState::update_and_draw()` and moves most of the logic from `ProgressBar::update_and_draw()` there.
4. Adds `ProgressState::finish()` and related methods, and makes the `ProgressBar::finish()` methods delegate to them.
5. Modifies the `Drop` trait for `ProgressState` to use the behavior specified by the `ProgressStyle::on_finish`.
6. Modifies the `Iterator` implementation for `ProgressBarIter` to use the `ProgressStyle::on_finish` behavior when the iterator completes (i.e. yields `None`).
7. Also adds a `Drop` trait to `ProgressBarIter` which makes sure the bar gets finished even if the iterator doesn't complete.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -50,8 +50,8 @@ impl<S, T: Iterator<Item = S>> Iterator for ProgressBarIter<T> {
 
         if item.is_some() {
             self.progress.inc(1);
-        } else {
-            self.progress.finish();
+        } else if !self.progress.is_finished() {
+            self.progress.finish_using_style();
         }
 
         item
@@ -70,8 +70,8 @@ impl<T: DoubleEndedIterator> DoubleEndedIterator for ProgressBarIter<T> {
 
         if item.is_some() {
             self.progress.inc(1);
-        } else {
-            self.progress.finish();
+        } else if !self.progress.is_finished() {
+            self.progress.finish_using_style();
         }
 
         item
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -444,6 +444,33 @@ impl ProgressState {
             state.status = Status::DoneVisible;
         });
     }
+
+    /// Finishes the progress bar using the [`ProgressFinish`] behavior stored
+    /// in the [`ProgressStyle`].
+    pub fn finish_using_style(&mut self) {
+        if let Some(on_finish) = std::mem::take(&mut self.style.on_finish) {
+            match on_finish {
+                ProgressFinish::Default => {
+                    self.finish();
+                }
+                ProgressFinish::AtCurrentPos => {
+                    self.finish_at_current_pos();
+                }
+                ProgressFinish::WithMessage(msg) => {
+                    self.finish_with_message(&msg);
+                }
+                ProgressFinish::AndClear => {
+                    self.finish_and_clear();
+                }
+                ProgressFinish::Abandon => {
+                    self.abandon();
+                }
+                ProgressFinish::AbandonWithMessage(msg) => {
+                    self.abandon_with_message(&msg);
+                }
+            }
+        }
+    }
 }
 
 /// A progress bar or spinner.
@@ -827,28 +854,7 @@ impl ProgressBar {
     /// Finishes the progress bar using the [`ProgressFinish`] behavior stored
     /// in the [`ProgressStyle`].
     pub fn finish_using_style(&self) {
-        let on_finish = self.state.lock().unwrap().style.get_on_finish().clone();
-        match on_finish {
-            ProgressFinish::Default => {
-                self.finish();
-            }
-            ProgressFinish::AtCurrentPos => {
-                self.finish_at_current_pos();
-            }
-            ProgressFinish::WithMessage(msg) => {
-                self.finish_with_message(&msg);
-            }
-            ProgressFinish::AndClear => {
-                self.finish_and_clear();
-            }
-            ProgressFinish::Abandon => {
-                self.abandon();
-            }
-            ProgressFinish::AbandonWithMessage(msg) => {
-                self.abandon_with_message(&msg);
-            }
-            ProgressFinish::None => { /* do nothing */ }
-        }
+        self.state.lock().unwrap().finish_using_style();
     }
 
     /// Sets a different draw target for the progress bar.
@@ -985,26 +991,28 @@ impl Drop for ProgressState {
 
         // How should we finish the bar?
         match std::mem::take(&mut self.style.on_finish) {
-            ProgressFinish::Default => {
-                self.finish();
-            }
-            ProgressFinish::AtCurrentPos => {
-                self.finish_at_current_pos();
-            }
-            ProgressFinish::WithMessage(msg) => {
-                self.finish_with_message(&msg);
-            }
-            ProgressFinish::AndClear => {
-                self.finish_and_clear();
-            }
-            ProgressFinish::Abandon => {
-                self.abandon();
-            }
-            ProgressFinish::AbandonWithMessage(msg) => {
-                self.abandon_with_message(&msg);
-            }
+            Some(on_finish) => match on_finish {
+                ProgressFinish::Default => {
+                    self.finish();
+                }
+                ProgressFinish::AtCurrentPos => {
+                    self.finish_at_current_pos();
+                }
+                ProgressFinish::WithMessage(msg) => {
+                    self.finish_with_message(&msg);
+                }
+                ProgressFinish::AndClear => {
+                    self.finish_and_clear();
+                }
+                ProgressFinish::Abandon => {
+                    self.abandon();
+                }
+                ProgressFinish::AbandonWithMessage(msg) => {
+                    self.abandon_with_message(&msg);
+                }
+            },
             // Fallback to original drop behavior for bars that are not finished.
-            ProgressFinish::None => {
+            None => {
                 self.status = Status::DoneHidden;
                 draw_state(self).ok();
             }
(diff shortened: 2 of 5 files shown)


src/iter.rs
```diff
@@ -39,13 +39,27 @@ impl<S, T: Iterator<Item = S>> Iterator for ProgressBarIter<T> {
         if next.is_some() {
             self.progress.inc(1);
         } else {
-            self.progress.finish();
+            // Iterator is exhausted.
```
djc, line 42: Please merge the below `if` with the `else` above.

src/progress.rs
```diff
@@ -348,6 +349,84 @@ impl ProgressState {
             (1_000_000_000 / avg_time) as u64
         }
     }
+
+    /// Call the provied `FnOnce` to update the state.  Then redraw the
```
djc, line 353: Would you mind isolating the pure refactoring (moving the finish updating code into the state into a separate commit or PR? That way I can more carefully review the other changes you're making here.

src/progress.rs
```diff
@@ -1246,6 +1308,13 @@ impl<I: Iterator> Iterator for ProgressBarIter<I> {
         if item.is_some() {
             self.bar.inc(1);
         }
+        else {
+            // Iterator is exhausted.
```
djc, line 1312: Inline the `if` into the `else` here, please.

src/style.rs
```diff
@@ -4,18 +4,49 @@ use std::cell::RefCell;
 use console::{measure_text_width, Style};
 
 use crate::format::{BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, HumanDuration};
-use crate::progress::ProgressState;
+use crate::progress::*;
```
djc, line 7: I'd prefer not to have wildcard imports.

src/style.rs
```diff
@@ -4,18 +4,49 @@ use std::cell::RefCell;
...
+    /// Finishes the progress bar and leaves the current message and progress.
+    /// Same behavior as calling [`ProgressBar::abandon()`].
+    Abandon,
+    /// Finishes the progress bar and sets a message, and leaves the current progress.
+    /// Same behavior as calling [`ProgressBar::abandon_with_message()`].
+    AbandonWithMessage(Cow<'static, str>),
+    /// Do nothing.  Leave the progres bar unfinished.
+    None,
+}
+impl Default for ProgressFinish {
```
djc, line 39: Add an empty line before this, please.

djc (commented):
This looks good, thanks! I'm afraid I merged some other changes that mean you'll need to rebase, and some comments below. In general if you find yourself enumerating the list of things the PR changes, I feel like separate commits might be better (although some of your items are definitely tightly coupled).

benkay86:
@djc, thank you for your review.  I have rebased And split the PR into 4 separate, logically-organized commits.  I think I addressed all of your formatting lints, although I may have missed something due to rebasing.  Please let me know what you think!

src/progress.rs
```diff
@@ -366,6 +366,84 @@ impl ProgressState {
             (1_000_000_000 / avg_time) as u64
         }
     }
+
+    /// Call the provied `FnOnce` to update the state.  Then redraw the
```
djc, line 370: Can you fix the typo here? `provied` -> `provided`.

src/progress.rs
```diff
@@ -824,6 +824,33 @@ impl ProgressBar {
         self.state.lock().unwrap().abandon_with_message(msg);
     }
 
+    /// Finishes the progress bar using the [`ProgressFinish`] behavior stored
+    /// in the [`ProgressStyle`].
+    pub fn finish_using_style(&self) {
+        let on_finish = self.state.lock().unwrap().style.get_on_finish().clone();
```
djc, line 830: Do we need the `.clone()` here?

Style nits: let's just inline the value of `on_finish` in the `match` statement below. I also think the match arms here could just be `self.finish(),` rather than `{\nself.finish();\n}`, can you see if that works? Additionally let's just skip the `/* do nothing */` comment, I don't think it adds value here.

src/style.rs
```diff
@@ -3,16 +3,51 @@ use console::{measure_text_width, Style};
...
+    }
+}
+
 /// Controls the rendering style of progress bars.
 #[derive(Clone, Debug)]
 pub struct ProgressStyle {
     tick_strings: Vec<Box<str>>,
     progress_chars: Vec<Box<str>>,
     template: Box<str>,
+    pub(crate) on_finish: ProgressFinish,
```
djc, line 50: It looks like the crate-level visibility isn't actually used in this commit. Maybe remove it?

src/progress.rs
```diff
@@ -978,12 +978,37 @@ impl WeakProgressBar {
...
     fn drop(&mut self) {
+        // Progress bar is already finished.  Do not need to do anything.
         if self.is_finished() {
             return;
         }
 
-        self.status = Status::DoneHidden;
-        draw_state(self).ok();
+        // How should we finish the bar?
+        match std::mem::take(&mut self.style.on_finish) {
```
djc, line 987: I don't understand why you use `take()` here?

src/iter.rs
```diff
@@ -51,7 +51,10 @@ impl<S, T: Iterator<Item = S>> Iterator for ProgressBarIter<T> {
         if item.is_some() {
             self.progress.inc(1);
         } else {
-            self.progress.finish();
+            // Finish progress bar immediately.
```
djc, line 54: As requested previously, please inline the `if` in the `else` above.

src/iter.rs
```diff
@@ -71,7 +74,10 @@ impl<T: DoubleEndedIterator> DoubleEndedIterator for ProgressBarIter<T> {
...
```
djc, line 77: As above.

djc (commented):
Thanks for splitting up the commits, that makes it much easier to follow!

benkay86:
I agree it would look nicer to inline the `match` arms, but doing so triggers a lint from rustfmt and fails Github's CI checks.

I refactored to avoid the clone, however the take, clone, and publicity issues are related.  Within a code block the Rust borrow checker tracks borrowing/moving disjoint fields.  However, when calling methods which receive `self`, the borrow checker assumes the method will use *all* the members of `self`.  The borrow checker cannot track disjoint fields across function calls.  This leads to compiler errors (see comments) in the following scenarios:

```rust
pub fn finish_using_style(&mut self) {
        if let Some(on_finish) = self.style.on_finish { // on_finish moved out of self here
            match on_finish {
                ProgressFinish::Default => {
                    self.finish();                      // attempt to borrow self here after partial move
                }
```

```rust
pub fn finish_using_style(&mut self) {
        if let Some(on_finish) = &self.style.on_finish { // self borrowed immutably here
            match on_finish {
                ProgressFinish::Default => {
                    self.finish();                       // attempt to borrow self mutably here
                }
```

Without splitting up `ProgressState` into multiple sub-structures, there are two ways to deal with this.  One is to `clone()`.  In this particular case the cost of cloning is no greater than the cost of cloning a `Cow` (but not as trivial as a `Copy`).

```rust
pub fn finish_using_style(&mut self) {
        let on_finish = self.style.on_finish.clone();
        if let Some(on_finish) = on_finish { // moved from clone on stack
            match on_finish {
                ProgressFinish::Default => {
                    self.finish();           // no conflict with self here
                }
```

Another, slightly more performant, workaround is to `take()`.  This requires mutable access to the field we are taking.  To give `ProgressState` mutable access to the `ProgressStyle::on_finish` field we need some kind of visibility specifier like `pub (crate)` or in this case just `pub (super)`.

```rust
pub fn finish_using_style(&mut self) {
        // take() moves on_finish from self onto stack and replaces
        // it with None so that self remains in a consistent state
        let on_finish = std::mem::take(&mut self.style.on_finish);
        if let Some(on_finish) = on_finish { // moved from stack
            match on_finish {
                ProgressFinish::Default => {
                    self.finish();           // no conflict with self here
                }
```

djc:
Okay, that makes sense then! Thanks for the thorough explanation, I will take a final look later today.

djc:
Thanks!

PR #246, opened 2021-03-12
Remove ProgressState::width

mibac138:
This variable was always `None` and used in only one place. It was
probably created to cache the target's width, however it's not too
expensive to check and can create problems after resizing a terminal (a
`wide_xyz` placeholder will use wrong width)

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -154,10 +154,7 @@ impl ProgressDrawTarget {
         match self.kind {
             ProgressDrawTargetKind::Term { ref term, .. } => term.size().1 as usize,
             ProgressDrawTargetKind::Remote { ref state, .. } => state.read().unwrap().width(),
-            ProgressDrawTargetKind::Hidden => {
-                // TODO: Expose `console::term::DEFAULT_WIDTH`?
-                79
-            }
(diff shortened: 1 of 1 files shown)


src/progress.rs
```diff
@@ -145,10 +145,7 @@ impl ProgressDrawTarget {
         match self.kind {
             ProgressDrawTargetKind::Term(ref term, ..) => term.size().1 as usize,
             ProgressDrawTargetKind::Remote(ref state, ..) => state.read().unwrap().width(),
-            ProgressDrawTargetKind::Hidden => {
-                // TODO: Expose `console::term::DEFAULT_WIDTH`?
-                79
-            }
+            ProgressDrawTargetKind::Hidden => 0,
```
djc, line 148: Could we make this `unreachable!()` or something similar, or is it still called?

djc (commented):
I think this makes sense...

src/progress.rs
mibac138: It was called in `ProgressBar::println` which I have now changed to only format state when it's not hidden. Meaning now it can be safely made unreachable.

djc:
Great, thanks!

PR #247, opened 2021-03-14
Allow usage of deprecated Term::is_term

mibac138:
cc https://github.com/mitsuhiko/console/issues/94


djc:
It looks like console 0.14.1 has been published, so we might not need this anymore?

mibac138:
Now that console is released the only case when this would be useful is if you update indicatif but not console, which seems unlikely.

PR #250, opened 2021-03-15
Switch custom Rng implementation for rand dependency

mibac138:
This shortens the example. The original implementation used a custom rng
implementation to [avoid rand dependency](https://github.com/mitsuhiko/indicatif/pull/178#discussion_r436044780), however it's already used as a dev-dependency, so it shouldn't matter.


djc:
Nice, thanks!

PR #252, opened 2021-03-15
Add back clone required to trigger a bug

mibac138:
This clone call is actually necessary. Without it a now-fixed bug will not be triggered. To test this change `ProgressDrawTarget::apply_draw_state` (line 165) from:
```rust
ProgressDrawTargetKind::Remote { idx, ref chan, .. } => {
    return chan
        .lock()
        .unwrap()
        .send((idx, draw_state))
        .map_err(|e| io::Error::new(io::ErrorKind::Other, e));
}
```
to:
```rust
ProgressDrawTargetKind::Remote { idx, ref chan, .. } => {
    chan
        .lock()
        .unwrap()
        .send((idx, draw_state))
        .map_err(|e| io::Error::new(io::ErrorKind::Other, e)).unwrap();
}
```

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -1333,7 +1333,10 @@ mod tests {
     fn late_pb_drop() {
         let pb = ProgressBar::new(10);
         let mpb = MultiProgress::new();
-        mpb.add(pb);
+        // This clone call is required to trigger a now fixed bug.
+        // See <https://github.com/mitsuhiko/indicatif/pull/141> for context
+        #[allow(clippy::redundant_clone)]
+        mpb.add(pb.clone());
     }
 
     #[test]


src/progress.rs
```diff
@@ -1333,7 +1333,9 @@ mod tests {
     fn late_pb_drop() {
         let pb = ProgressBar::new(10);
         let mpb = MultiProgress::new();
-        mpb.add(pb);
+        // This clone call is required to trigger a now fixed bug.
```
djc, line 1336: Can you add a suppression of the clippy lint?

mibac138: I think it's now suppressed, but I couldn't make clippy emit the warning so I'm not sure. Can you share how you run clippy to make the warning appear?

djc:
Thanks for the quick response and reviewing my changes!

PR #254, opened 2021-03-15
Use Mutex instead of RwLock in ProgressBar

mibac138:
Nearly every call to RwLock was write() defying it's advantage over
Mutex. Additionally, Mutex<T> implements Sync for T: Send, contrary
to RwLock<T> which implements Sync for T: Sync, which is a bit harder to
satisfy and thus less versatile.


djc:
Makes sense, thanks!

mibac138:
Thanks for your quick responses! :)

PR #255, opened 2021-03-15
Compress Estimate struct

mibac138:
Size reduced from 72 bytes to 48 bytes.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -16,42 +16,55 @@ pub fn secs_to_duration(s: f64) -> Duration {
 }
 
 pub struct Estimate {
-    buf: Vec<f64>,
-    buf_cap: usize,
-    last_idx: usize,
+    buf: Box<[f64; 15]>,
+    data: u8,
     started: Option<(Instant, u64)>,
 }
 
 impl Estimate {
-    pub fn new() -> Estimate {
-        Estimate {
-            buf: vec![],
-            buf_cap: 10,
-            last_idx: 0,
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -16,64 +16,77 @@ pub fn secs_to_duration(s: f64) -> Duration {
 }
 
 pub struct Estimate {
-    buf: Vec<f64>,
-    buf_cap: usize,
-    last_idx: usize,
+    buf: Box<[f64; 15]>,
```
djc, line 19: The mechanisms you're using here could use a bunch of comments -- it's not at all clear to me what the approach is. Also, do you have an actual use case where this matters or do you just like squeezing out performance? :)

mibac138: I do like squeezing out performance and that's the main reason, however there were a few issues related to performance and this is a small stepping stone in improving it.

src/utils.rs
```diff
@@ -15,8 +15,17 @@ pub fn secs_to_duration(s: f64) -> Duration {
     Duration::new(secs, nanos)
 }
 
+/// Ring buffer with constant capacity
```
djc, line 18: Would you mind adding some notes about how/where this is used?

src/utils.rs
djc: I'm curious what kind of scenarios you're seeing where issues like these impact performance. Can you say more about that?

djc:
Thanks!

PR #256, opened 2021-03-15
Move ProgressState::tick to ProgressStyle

mibac138:
And change it from a `u64` to `u8` which is still more than enough

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::cell::Cell;
 
 use console::{measure_text_width, Style};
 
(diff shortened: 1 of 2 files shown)


src/style.rs
```diff
@@ -13,6 +14,7 @@ use unicode_segmentation::UnicodeSegmentation;
 #[derive(Clone, Debug)]
 pub struct ProgressStyle {
     tick_strings: Vec<Box<str>>,
+    tick: Cell<u8>,
```
djc, line 17: Could use some comments. What is `tick`? Why does it need a `Cell` wrapped around it?

djc:
Hi @mibac138, would you be willing to rebase this on top of current main?

djc:
Going to close this for now, feel free to reopen if you want to work on this again.

PR #257, opened 2021-03-15
Change ProgressStyle::template from Cow<'static, str> to Box<str>

mibac138:
This reduces ProgressStyle's size by 16 bytes.

Cow is useful when occasionally mutating the data inside, however
it was never mutated, and only read from. The only, negligible, benefit
is when reusing a single static str as a template across multiple
progress bars (as the template would be stored in memory only once;
as an example the template in download-continued is 106 bytes),
however the current API doesn't even expose such a functionality despite
it being easily implementable.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -1,5 +1,3 @@
-use std::borrow::Cow;
-
 use console::{measure_text_width, Style};
 
 use crate::format::{BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, HumanDuration};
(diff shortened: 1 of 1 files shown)


src/style.rs
```diff
@@ -1,10 +1,7 @@
-use std::borrow::Cow;
-
-use console::{measure_text_width, Style};
-
 use crate::format::{BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, HumanDuration};
 use crate::progress::ProgressState;
 use crate::utils::{expand_template, pad_str};
+use console::{measure_text_width, Style};
```
djc, line 4: Please keep the third-party crate exports above the local crate imports, in a separate block.

src/style.rs
```diff
@@ -68,7 +65,7 @@ impl ProgressStyle {
                 .collect(),
             progress_chars,
             char_width,
-            template: Cow::Borrowed("{wide_bar} {pos}/{len}"),
+            template: Box::<str>::from("{wide_bar} {pos}/{len}"),
```
djc, line 68: I'd prefer simplifying this to `"{wide_bar} {pos}/{len}".into()` (and the same for the other cases).

djc:
Thanks!

PR #258, opened 2021-03-15
Implement `remove` method for `MultiProgress`

nlinker:
out

Fixes https://github.com/mitsuhiko/indicatif/issues/184

Redone https://github.com/mitsuhiko/indicatif/pull/185 against other branch and rebased on the fresh main

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -14,6 +14,7 @@ use crate::state::{
 use crate::style::ProgressStyle;
 use crate::utils::Estimate;
 use crate::{ProgressBarIter, ProgressIterator};
+use std::collections::HashSet;
 
 /// A progress bar or spinner.
 ///
@@ -523,7 +524,8 @@ impl MultiProgress {
         let (tx, rx) = channel();
         MultiProgress {
             state: Arc::new(RwLock::new(MultiProgressState {
-                objects: vec![],
+                objects: Vec::new(),
+                free_set: HashSet::new(),
                 ordering: vec![],
                 draw_target,
                 move_cursor: false,
@@ -556,16 +558,12 @@ impl MultiProgress {
     /// object overriding custom `ProgressDrawTarget` settings.
     pub fn add(&self, pb: ProgressBar) -> ProgressBar {
         let mut state = self.state.write().unwrap();
-        let idx = state.objects.len();
-        state.objects.push(MultiObject {
-            done: false,
-            draw_state: None,
-        });
-        state.ordering.push(idx);
+        let object_idx = MultiProgress::extend_state(&mut state);
+        state.ordering.push(object_idx);
         pb.set_draw_target(ProgressDrawTarget {
             kind: ProgressDrawTargetKind::Remote {
                 state: self.state.clone(),
-                idx,
+                idx: object_idx,
                 chan: Mutex::new(self.tx.clone()),
             },
         });
@@ -582,11 +580,7 @@ impl MultiProgress {
     /// is added to the end of the list.
     pub fn insert(&self, index: usize, pb: ProgressBar) -> ProgressBar {
         let mut state = self.state.write().unwrap();
-        let object_idx = state.objects.len();
-        state.objects.push(MultiObject {
-            done: false,
-            draw_state: None,
-        });
+        let object_idx = MultiProgress::extend_state(&mut state);
         if index > state.ordering.len() {
             state.ordering.push(object_idx);
         } else {
@@ -602,6 +596,32 @@ impl MultiProgress {
         pb
     }
 
+    /// Removes a progress bar.
+    ///
+    /// The progress bar is removed only if it was previously inserted or added
+    /// by the methods `MultiProgress::insert` or `MultiProgress::add`.
+    /// If the passed progress bar does not satisfy the condition above,
+    /// the `remove` method does nothing.
+    pub fn remove(&self, pb: &ProgressBar) {
+        let mut state = self.state.write().unwrap();
+        match &pb.state.lock().unwrap().draw_target.kind {
+            ProgressDrawTargetKind::Term { .. } => {}
+            ProgressDrawTargetKind::Remote { idx, .. } => {
+                state.objects[*idx] = None;
+                state.free_set.insert(*idx);
+                if let Some(idx) = state.ordering.iter().position(|x| *x == *idx) {
+                    state.ordering.remove(idx);
+                }
+            }
+            ProgressDrawTargetKind::Hidden => {}
+        }
+    }
+
+    pub fn count(&self) -> usize {
+        let state = self.state.read().unwrap();
+        state.objects.len()
+    }
+
     /// Waits for all progress bars to report that they are finished.
     ///
     /// You need to call this as this will request the draw instructions
@@ -614,6 +634,26 @@ impl MultiProgress {
     /// Works like `join` but clears the progress bar in the end.
     pub fn join_and_clear(&self) -> io::Result<()> {
         self.join_impl(true)
+    }
+
+    /// Modify the state on inserting/adding a progress bar
+    fn extend_state(state: &mut MultiProgressState) -> usize {
+        let object_idx;
+        if let Some(i) = state.free_set.iter().next().cloned() {
+            state.objects[i] = Some(MultiObject {
+                done: false,
+                draw_state: None,
+            });
+            state.free_set.remove(&i);
+            object_idx = i;
+        } else {
+            state.objects.push(Some(MultiObject {
+                done: false,
+                draw_state: None,
+            }));
+            object_idx = state.objects.len() - 1;
+        }
+        object_idx
     }
 
     fn join_impl(&self, clear: bool) -> io::Result<()> {
@@ -641,7 +681,9 @@ impl MultiProgress {
 
             let mut state = self.state.write().unwrap();
             if draw_state.finished {
-                state.objects[idx].done = true;
+                if let Some(ref mut obj) = &mut state.objects[idx] {
+                    obj.done = true;
+                }
             }
 
             // Split orphan lines out of the draw state, if any
@@ -659,7 +701,9 @@ impl MultiProgress {
                 ..draw_state
             };
 
-            state.objects[idx].draw_state = Some(draw_state);
+            if let Some(ref mut obj) = &mut state.objects[idx] {
+                obj.draw_state = Some(draw_state);
+            }
 
             // the rest from here is only drawing, we can skip it.
             if state.draw_target.is_hidden() {
@@ -688,9 +732,10 @@ impl MultiProgress {
             lines.append(&mut orphan_lines);
 
             for index in state.ordering.iter() {
-                let obj = &state.objects[*index];
-                if let Some(ref draw_state) = obj.draw_state {
-                    lines.extend_from_slice(&draw_state.lines[..]);
+                if let Some(obj) = &state.objects[*index] {
+                    if let Some(ref draw_state) = obj.draw_state {
+                        lines.extend_from_slice(&draw_state.lines[..]);
+                    }
                 }
             }
 
@@ -842,5 +887,52 @@ mod tests {
         let _: Box<dyn Send> = Box::new(ProgressBar::new(1));
         let _: Box<dyn Sync> = Box::new(MultiProgress::new());
         let _: Box<dyn Send> = Box::new(MultiProgress::new());
+    }
+
+    #[test]
+    fn multi_progress_modifications() {
+        let mp = MultiProgress::new();
+        let p0 = mp.add(ProgressBar::new(1));
+        let p1 = mp.add(ProgressBar::new(1));
+        let p2 = mp.add(ProgressBar::new(1));
+        let p3 = mp.add(ProgressBar::new(1));
+        mp.remove(&p2);
+        mp.remove(&p1);
+        let p4 = mp.insert(1, ProgressBar::new(1));
+
+        let state = mp.state.read().unwrap();
+        // the removed place for p1 is reused
+        assert_eq!(state.objects.len(), 4);
+        assert_eq!(
+            state
+                .objects
+                .iter()
+                .map(|o| if o.is_some() { 1 } else { 0 })
+                .sum::<usize>(),
+            3
+        );
+        // free_set may contain 1 or 2
+        let set_1 = vec![1_usize].into_iter().collect();
+        let set_2 = vec![2_usize].into_iter().collect();
+        if state.free_set == set_1 {
+            assert_eq!(state.ordering, vec![0, 2, 3]);
+            assert_eq!(extract_index(&p4), 2);
+        } else if state.free_set == set_2 {
+            assert_eq!(state.ordering, vec![0, 1, 3]);
+            assert_eq!(extract_index(&p4), 1);
+        } else {
+            unreachable!();
+        }
+        assert_eq!(extract_index(&p0), 0);
+        assert_eq!(extract_index(&p1), 1);
+        assert_eq!(extract_index(&p2), 2);
+        assert_eq!(extract_index(&p3), 3);
+    }
+
+    fn extract_index(pb: &ProgressBar) -> usize {
+        match pb.state.lock().unwrap().draw_target.kind {
+            ProgressDrawTargetKind::Remote { idx, .. } => idx,
+            _ => unreachable!(),
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


nlinker:
I guess AppVeyor build has not been setup yet and the CI failure is expected.

djc:
Yeah, you can ignore the AppVeyor stuff.

djc (commented):
Okay, that seems like a lot of changes. I'd like to see a minimal version of this first, one that doesn't change any examples (except insofar as needed to make them compile) and makes minimal changes to the internal data structures. Are you okay with this?>

nlinker:
> Okay, that seems like a lot of changes. I'd like to see a minimal version of this first, one that doesn't change any examples (except insofar as needed to make them compile) and makes minimal changes to the internal data structures. Are you okay with this?

Thank you for feedback. Yes, will do. This change is very needed to show progressing process' tree for https://github.com/whamcloud/integrated-manager-for-lustre/pull/1975 (and the continuation in the private fork). Let me describe briefly the changes made:

1. `MultiProgress` is essentially the collection of `ProgressBar`-s with the external objects, that refer to the `ProgressBar`. Each external object has the index of the `ProgressBar` and can call the methods on it (progress bar) when some event occurs.

2. Before the current PR, the collection was vector. The `remove` method, however, will shift the indices of the vector, which means the external objects are no longer able to find the correct progress bar to update. This is the reason, why the collection type is changed from `Vec` to `HashMap` – the keys stay the same after the modifications.

3. Since external objects now keep not indices, but keys, the type of key is changed to `u32`, to explicitly show that it is not an index somewhere and one cannot treat the number as index. It is more type safe as well.

4. Hence we need to have a way to generate the keys and make sure they are unique for any new `ProgressBar`. There are two choices for the generation: random and sequential. Random forces us to bring a new dependency (and the keys can be theoretically non-unique), sequential forces us to bring some counter. I chose to have a counter for each `MultiProgress`.

5. The `multi-tree` example is changed accordingly to show the removal of the temporary progress bar. This makes the example changed substantially. I'm ok to hold the changes to it for now.

djc:
I think I would prefer keeping a `Vec`, just changing it to have `Option` in it. That way we optimize for the common case (where removing doesn't happen) -- hash maps are relatively slow/memory-hungry compared to `Vec`.

@mibac138 any opinions on if/how we should do this?

nlinker:
@djc @mibac138 Let me add my 2 cents.

Tombstones (i.e. `None` placeholders in the places where the elements were removed) have the huge disadvantage with the pattern, when elements frequently is added and then removed. This usage pattern will blow up the underlying storage with lots of `None` elements in it.

Let's consider the performance of `HashMap`s. It is not as slow as our intuition says about it:
```
running 6 tests
test bench_hm_10    ... bench:          82 ns/iter (+/- 17)
test bench_hm_100   ... bench:         788 ns/iter (+/- 180)
test bench_hm_1000  ... bench:       7,625 ns/iter (+/- 1,631)
test bench_vec_10   ... bench:           3 ns/iter (+/- 0)
test bench_vec_100  ... bench:          23 ns/iter (+/- 1)
test bench_vec_1000 ... bench:         356 ns/iter (+/- 96)
```
The benchmark shows that in Rust iteration through 1000-element takes _7.6 microseconds_ which IMO way fast enough, if we want to render with 1000 FPS, we can allow us to do it  `1000 / 7.6 = 131` times slower.

The code of the benchmark is here: 
```rust
#![feature(test)]
extern crate test;

use indicatif::MultiObject;
use test::Bencher;

fn create_hm(n: usize) -> HashMap<u32, MultiObject> {
    let mut hm = HashMap::new();
    for j in 0..n {
        hm.insert(j as u32, MultiObject {
            done: j % 2 == 0,
            draw_state: None,
        });
    }
    hm
}

fn create_vec(n: usize) -> Vec<Option<MultiObject>> {
    let mut v = Vec::with_capacity(n);
    for j in 0..n {
        v.push(Some(MultiObject {
            done: j % 2 == 0,
            draw_state: None,
        }));
    }
    v
}

fn run_hm(xs: &HashMap<u32, MultiObject>) -> usize {
    let m = xs.len();
    let mut count = 0;
    for x in 0..m as u32 {
        if let Some(v) = xs.get(&x) {
            if v.done {
                count += 1;
            }
        }
    }
    count
}

fn run_vec(xs: &Vec<Option<MultiObject>>) -> usize {
    let m = xs.len();
    let mut count = 0;
    for x in 0..m {
        if let Some(v) = &xs[x] {
            if v.done {
                count += 1;
            }
        }
    }
    count
}

fn run_bench_vec(b: &mut Bencher, m: usize) {
    let xs = create_vec(m);
    b.iter(|| {
        let i = run_vec(&xs);
        assert_eq!(i, m/2);
    })
}

fn run_bench_hm(b: &mut Bencher, m: usize) {
    let xs = create_hm(m);
    b.iter(|| {
        let i = run_hm(&xs);
        assert_eq!(i, m/2);
    })
}

#[bench]
fn bench_vec_10(b: &mut Bencher) {
    run_bench_vec(b, 10);
}
#[bench]
fn bench_hm_10(b: &mut Bencher) {
    run_bench_hm(b, 10);
}

#[bench]
fn bench_vec_100(b: &mut Bencher) {
    run_bench_vec(b, 100);
}
#[bench]
fn bench_hm_100(b: &mut Bencher) {
    run_bench_hm(b, 100);
}

#[bench]
fn bench_vec_1000(b: &mut Bencher) {
    run_bench_vec(b, 1000);
}
#[bench]
fn bench_hm_1000(b: &mut Bencher) {
    run_bench_hm(b, 1000);
}
```

nlinker:
> Is your use case actually affected by the disadvantages of tombstones, or are you just reasoning about this in terms of abstract complexities?

This is the loop going through the collection in the rendering part.

What about making it `Vec<Option<MultiObject>>`, but I can improve `insert` function so that it finds the first tombstone to insert a new `ProgressBar` instead of appending it to the end? There could be some ineffective patterns, but the straightforward alternating insert-then-remove will be efficient.

@djc, does it sound as a good plan for you?

djc:
That sounds good! I do think we should then have something like a free list, so we don't have to walk the list to check for tombstones. Or maybe use something like slotmap?

src/progress.rs
```diff
@@ -1080,19 +1081,25 @@ impl MultiProgress {
     /// object overriding custom `ProgressDrawTarget` settings.
     pub fn add(&self, pb: ProgressBar) -> ProgressBar {
         let mut state = self.state.write().unwrap();
-        let object_id = self.id_counter.fetch_add(1, Ordering::SeqCst);
-        state.objects.insert(
-            object_id,
-            MultiObject {
+        let object_idx;
+        if let Some(i) = state.objects.iter().position(|x| x.is_none()) {
```
djc, line 1085: I think I suggested a free list in my previous commit, did you miss that?

nlinker: I tried, however I was unsure about the code complexity added. Given your suggestion I've made the changes in the followup commit.

nlinker:
Hi, any questions/thoughts on the change?

djc:
Haven't had time to carefully review it yet. Hope to do that soon.

src/progress_bar.rs
```diff
@@ -555,16 +557,12 @@ impl MultiProgress {
     /// object overriding custom `ProgressDrawTarget` settings.
     pub fn add(&self, pb: ProgressBar) -> ProgressBar {
         let mut state = self.state.write().unwrap();
-        let idx = state.objects.len();
-        state.objects.push(MultiObject {
-            done: false,
-            draw_state: None,
-        });
-        state.ordering.push(idx);
+        let object_idx = MultiProgress::extend_state(&mut state);
```
djc, line 560: Let's just use the `idx` name here, please.

src/progress_bar.rs
```diff
@@ -522,7 +523,8 @@ impl MultiProgress {
         let (tx, rx) = channel();
         MultiProgress {
             state: Arc::new(RwLock::new(MultiProgressState {
-                objects: vec![],
+                objects: Vec::new(),
+                free_set: HashSet::new(),
```
djc, line 527: I think we can use a `BinaryHeap` for the free set, or do you see an actual risk of duplicates?

src/progress_bar.rs
```diff
@@ -615,6 +635,26 @@ impl MultiProgress {
...
+            state.free_set.remove(&i);
+            object_idx = i;
+        } else {
+            state.objects.push(Some(MultiObject {
+                done: false,
+                draw_state: None,
+            }));
+            object_idx = state.objects.len() - 1;
+        }
+        object_idx
```
djc, line 655: Let's extract something like `let new = Some(MultiObject { .. })` from the `if/else` block, and assign the `object_idx` from the block directly (so `let object_idx = if let .. { } else { .. };`)?

src/state.rs
```diff
@@ -274,9 +275,15 @@ impl Drop for ProgressState {
 }
 
 pub(crate) struct MultiProgressState {
-    pub(crate) objects: Vec<MultiObject>,
+    /// The collection of states corresponding to progress bars
```
djc, line 278: Thanks for adding these comments!

djc (commented):
Sorry for the delay, but this is looking good. Some minor nits left...

src/progress_bar.rs
nlinker: Sure, will do it.

src/progress_bar.rs
nlinker: I was guided by the performance of `pop`, which in case of `HashSet` is O(1), it is just
```rust
let Some(i) = state.free_set.iter().next().cloned()
```
and in case of other data structures not that fast (e.g. `BinHeap`'s `pop` operation is O(log). Fortunately we don't have to choose the minimum tombstone, just some of them if any, since the visual order is kept maintained by `ordering` array.

The duplication can be done by misuse, with e.g.:
```
let pb = mp.add(ProgressBar::new(10));
mp.remove(&pb);
mp.remove(&pb);
```
After the second `remove` the `free_set` would contain 2 zeros, if it allowed duplications.
I think I add the test to control the correct behavior

djc: We can prevent the duplication by checking at remove time whether something is present in the slot (`is_some()`), and only adding it to the free set if it was previously occupied.

I'm not sure the `O(log)` complexity of `BinaryHeap::pop()` should be guiding? In your project, how does the number of progress bar additions relate to the number of removals? If `pop()` is cheaper but `push()` is substantially more expensive (which I think would be the case for `HashSet` compared to `BinaryHeap`), the `HashSet` doesn't work to our advantage (even ignoring the added memory usage)

nlinker: I can imagine the number of progress bars like 100-150 in worst cases. I think `BinaryHeap` works for our case too, I will make the change.

src/progress_bar.rs
```diff
@@ -615,6 +635,26 @@ impl MultiProgress {
...
 
+    /// Modify the state on inserting/adding a progress bar
+    fn extend_state(state: &mut MultiProgressState) -> usize {
+        let object_idx;
+        if let Some(i) = state.free_set.iter().next().cloned() {
+            state.objects[i] = Some(MultiObject {
+                done: false,
+                draw_state: None,
+            });
+            state.free_set.remove(&i);
```
nlinker, line 646: Oops, this operation is unsupported for heaps.

src/progress_bar.rs
```diff
@@ -607,10 +607,12 @@ impl MultiProgress {
         match &pb.state.lock().unwrap().draw_target.kind {
             ProgressDrawTargetKind::Term { .. } => {}
             ProgressDrawTargetKind::Remote { idx, .. } => {
-                state.objects[*idx] = None;
-                state.free_set.insert(*idx);
-                if let Some(idx) = state.ordering.iter().position(|x| *x == *idx) {
-                    state.ordering.remove(idx);
+                if state.objects[*idx].is_some() {
```
djc, line 610: Let's use `take()` here.

src/progress_bar.rs
djc: Hmm, I think `BTreeSet` could also be a decent option here, what do you think?

nlinker: To reduce memory footprint? Ok, let's try.

src/progress_bar.rs
djc: That's the number, but I meant are additions/removals generally 1:1? Or do you expect that there will be more like 100 additions and 25 removals?

src/progress_bar.rs
```diff
@@ -638,22 +640,18 @@ impl MultiProgress {
 
     /// Modify the state on inserting/adding a progress bar
     fn extend_state(state: &mut MultiProgressState) -> usize {
-        let object_idx;
+        let new_obj = Some(MultiObject {
+            done: false,
+            draw_state: None,
+        });
         if let Some(i) = state.free_set.iter().next().cloned() {
```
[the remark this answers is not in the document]
nlinker: Hm, otherwise types aren't line up and we have to do like this:
```
        *(if let Some(i) = state.free_set.iter().next() {
            state.objects[*i] = new_obj;
            state.free_set.remove(i);
            i
        } else {
            state.objects.push(new_obj);
            state.objects.len() - 1
        })
```

src/progress_bar.rs
nlinker: I'm not sure I understand. `take`  returns us `Option<MultiObject>`, which we don't need.

src/progress_bar.rs
nlinker: Let me show
image
This is all complete state, but in reality many of tree nodes are working and the progress bars show their statuses.

Each collapse/expand removes/adds a bunch of progress bars at once, the extreme case is to collapse/expand the root node, the operation removes and adds almost all progress bars.
100 lines on the terminal we can do using smaller fonts, I think 150 is the upper bound on the `MultiProgress` size, in case of big trees we will implement some workarounds (scrolls or so).

src/progress_bar.rs
```diff
@@ -582,26 +580,50 @@ impl MultiProgress {
...
+        let mut state = self.state.write().unwrap();
+        match &pb.state.lock().unwrap().draw_target.kind {
+            ProgressDrawTargetKind::Term { .. } => {}
+            ProgressDrawTargetKind::Remote { idx, .. } => {
+                if state.objects[*idx].is_some() {
+                    state.objects[*idx] = None;
+                    state.free_set.insert(*idx);
+                    if let Some(idx) = state.ordering.iter().position(|x| *x == *idx) {
+                        state.ordering.remove(idx);
+                    }
```
djc, line 615:
```suggestion
                if let Some(obj) = state.objects[*idx].take() {
                    state.free_set.insert(*idx);
                    state.ordering.retain(|&x| x != *idx);
```

src/progress_bar.rs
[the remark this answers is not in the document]
djc: So worst case we can use `copied()` here instead of `cloned()`. But the whole `let i = ..iter().next(); ...remove(i);` pattern seems pretty ugly if we can also pick a data type that has a natural `pop()` method. Since `BTreeSet` doesn't have that, I still feel like `Vec` or `VecDeque` or `BinaryHeap` are more attractive.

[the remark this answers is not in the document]
nlinker: Nit: `cloned` for `Copy` types works exactly like `copied`: https://godbolt.org/z/1e3YaMnhx
Although clippy kept silent here despite on https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy

src/progress_bar.rs
nlinker: This is neat!

djc:
I made an editing pass, split this in multiple commits, and merged it into main just now. Final commit is here: bc08ee6dfb3c8cab7150983ae6523fdc6187bc1c. Thanks for working on this!

src/progress_bar.rs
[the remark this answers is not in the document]
nlinker: Let's recap the requirements, the collection should have the following operation:
- add the new element to the collection (`insert(e)`)
- remove some element from the collection (`remove(e)`)
- get and remove _any_ existing number from the collection (`if let Some(e) = pop()`)

For me `HashSet` fulfils all the requirements with constant time for the some redundant memory consumption.
`BTreeSet` brings logarithmic time for all the operations and more overhead compared to `HashSet`, but given the bound fulfills the requirements as well.
`VecDeque` brings sequential time to the `remove`  operation, and doesn't have `remove(e)`, we have to use `retain(|x| x != e)` as the workaround.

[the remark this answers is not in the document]
djc: I just changed it to `Vec` after reviewing the requirements some more.

nlinker:
> I made an editing pass, split this in multiple commits, and merged it into main just now. Final commit is here: [bc08ee6](https://github.com/mitsuhiko/indicatif/commit/bc08ee6dfb3c8cab7150983ae6523fdc6187bc1c). Thanks for working on this!

Cool, thanks, Dirkjan! I tried to use `VecDeque` for `free_set`, it works too.
Also I'd make a changes into `multi-tree` example to show the new behavior.

PR #259, opened 2021-03-17
Unify Estimate behavior

mibac138:
Up until now Estimate would be in a fully initialized state after
calling `new`, however after calling `reset` it would get fully
initialized only on the next call to `record_step`. This is an
unexpected disparity that this commit unifies. Now, when calling `reset`
 you must also provide a `start_value`, so that the Estimate can not be
 in a semi-initialized state anymore.


djc:
Thanks!

PR #261, opened 2021-03-19
Fix deadlock when using `ProgressBar::abandon`

mibac138:
This is basically the same as #187 except that the problem wasn't in `impl Drop` but in `fn abandon` this time

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -795,6 +795,16 @@ mod tests {
     }
 
     #[test]
+    fn test_abandon_deadlock() {
+        let mpb = MultiProgress::new();
+        let pb = mpb.add(ProgressBar::new(1));
+        pb.set_draw_delta(2);
+        pb.abandon();
+        drop(pb);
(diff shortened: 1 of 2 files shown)


src/progress.rs
```diff
@@ -1070,6 +1072,17 @@ fn test_draw_delta_deadlock() {
     mpb.join().unwrap();
 }
 
+#[test]
+fn test_abanon_deadlock() {
```
djc, line 1076: Typo.

src/progress.rs
```diff
@@ -432,6 +432,7 @@ impl ProgressState {
     /// Finishes the progress bar and leaves the current message and progress.
     pub fn abandon(&mut self) {
         self.update_and_draw(|state| {
+            state.draw_next = state.pos;
```
djc, line 435: This is not very obvious. Should we just always draw in state DoneVisible? Or give the closure a return type that will force it to draw?

djc:
Thanks!

PR #262, opened 2021-03-29
Update rayon version

Titaniumtown:


djc:
Is this actually necessary? If not there is no reason to do this, since 1.0 and 1.5 are compatible Cargo will automatically pull in rayon 1.5 whenever you build indicator.

Titaniumtown:
I don't see any harm in it :shrug:

djc:
We usually declare the minimum compatible version that's known to work. Bumping unnecessarily can make life harder for downstream users that rely on an older version for whatever reason.

Titaniumtown:
Got it!

PR #272, opened 2021-04-12
Make state methods publicly available

christophe-rannou:
A few methods of the progress bar underlying state can be useful to external operator, adding wrapper method to expose them.


djc:
Thanks!

PR #274, opened 2021-04-12
Fix download examples templates

mibac138:
Related https://github.com/mitsuhiko/indicatif/issues/263#issuecomment-817138712


djc:
Thanks!

PR #275, opened 2021-04-16
Unwrap ProgressFinish

mibac138:
This changes default behavior - now a progress bar without an explicitly set on_finish will call `finish`, when previously it would call `finish_and_clear` ([`src/style.rs:103`](https://github.com/mitsuhiko/indicatif/compare/main...mibac138:on-finish?expand=1#diff-5452dc5fb15e73154c0ecc33e87a14379a97f817fe0cdb0fb9d17df712feb8fbR103)). Also, `on_finish` won't be reset every time `finish_with_style` is called.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -45,7 +45,7 @@ pub struct ProgressStyle {
     tick_strings: Vec<Box<str>>,
     progress_chars: Vec<Box<str>>,
     template: Box<str>,
-    pub(super) on_finish: Option<ProgressFinish>,
+    on_finish: ProgressFinish,
     // how unicode-big each char in progress_chars is
     char_width: usize,
 }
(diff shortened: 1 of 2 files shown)


djc:
Thanks for the PR -- can you explain why you think these changes represent an improvement?

And also, would you mind going over your other open PRs to see if they are still relevant/useful?

mibac138:
I'm working on #192, probably will update today, and #132 will be obsolete once I finish another PR, no eta on it though. #256 I have currently postponed and will do some refactoring before returning to it.
* I agree about changing the default finish behavior but didn't want to change `ProgressFinish::default` without feedback. Changed it now to keep the same behavior.
* I find it surprising that `on_finish` is reset after `finish_with_style`. I think this is the expected behavior from an end user perspective.

src/style.rs
```diff
@@ -15,14 +15,14 @@ use unicode_segmentation::UnicodeSegmentation;
 pub enum ProgressFinish {
     /// Finishes the progress bar and leaves the current message.
     /// Same behavior as calling [`ProgressBar::finish()`].
-    Default,
+    Finish,
```
djc, line 18: I think it would be clearer if we call this `AndLeave`, what do you think?

mibac138: Sounds good, it's more consistent with the rest, changed it

PR #276, opened 2021-04-16
MultiProgress::remove should verify the ProgressBar's remote

mibac138:


djc:
Why? What problem is this intended to solve?

mibac138:
Mostly a sanity check to prevent a situation of this kind from happening:
```rust
let mp1 = MultiProgress::new();
let mp2 = MultiProgress::new();
let pb = mp1.add(ProgressBar::new());
mp2.remove(pb);
```
Which could mess up the output. While unlikely to happen I think it's helpful to document internal invariants and it just might save someone a headache if for some obscure reason they'd be using two multi progresses.

djc:
Okay, please add a single-line comment with something like `// Check that this progress bar is owned by the current MultiProgress`.

PR #277, opened 2021-04-16
Shrink ProgressDrawTarget size

mibac138:


src/state.rs
```diff
@@ -468,9 +471,8 @@ impl ProgressDrawTarget {
             } => {
                 if draw_state.finished
                     || draw_state.force_draw
-                    || rate.is_none()
-                    || last_draw.is_none()
-                    || last_draw.unwrap().elapsed() > rate.unwrap()
+                    || rate == Duration::from_secs(0)
```
djc, line 466: I honestly don't like this use of a sentinel value. The benefit of saving 8 bytes doesn't really seem worth it to me.

djc (commented):
I like the `last_state` -> `last_line_count` change, that seems like a nice simplification. On the other hand, I'm less sure about removing the `Option` wrappers in favor of using sentinel values if the main benefit is just a few bytes of memory usage.

djc:
Okay, I will merge this.

PR #278, opened 2021-04-19
Add ProgressDrawState::split_off_orphan_lines

mibac138:
This simplifies logic in `MultiProgress::join_impl` and makes it easier to follow


src/progress_bar.rs
```diff
@@ -691,20 +691,8 @@ impl MultiProgress {
...
-            } else {
-                draw_state.lines
-            };
-
-            let draw_state = ProgressDrawState {
-                lines,
-                orphan_lines: 0,
-                ..draw_state
-            };
+            let orpan = draw_state.split_off_orphan_lines();
```
djc: By default, I'm not a fan of adding single-use functions or methods. I find that the added indirection can often make it harder to understand what's going on. What is the point of this change?

PR #279, opened 2021-04-24
Have cargo example more closely match cargo

LovecraftianHorror:
Hello thanks for the great crate! I loved all the examples and just wanted to improve on the `cargo` example a bit. Just a quick overview of the changes:

- Bumped up the time to "build" each crate so it's easier to follow what's happening
- Added a `Compiling ...` line for each crate
    - This did involve storing a dummy version as well
- Added a `Finished ...` line to better match `cargo`

### Old example

old_cargo_example

### New example

new_cargo_example


djc:
Thanks!

PR #281, opened 2021-04-30
Make "unicode-width" dependency actually optional

### LINKED ISSUE #280: indicatif depends on unicode-width unconditionally, even though the feature is optional
The `unicode-width` dependency is supposed to be pulled in only when the `improved_unicode` feature is enabled:

https://github.com/mitsuhiko/indicatif/blob/73a0be3e7c9b0706856fd463ee8c478be58c1556/Cargo.toml#L29

However, it is always present in the dependency tree regardless because of `console` crate:

```
cargo tree | grep --before-context=8 unicode-width
├── indicatif v0.16.0
│   ├── console v0.14.1
│   │   ├── lazy_static v1.4.0
│   │   ├── libc v0.2.86
│   │   ├── regex v1.4.6
│   │   │   └── regex-syntax v0.6.23
│   │   ├── terminal_size v0.1.16
│   │   │   └── libc v0.2.86
│   │   └── unicode-width v0.1.8
```

It appears that `indicatif` also needs to add `default-features=false` annotation to the `console` dependency.

Shnatsel:
Fixes #280

This also disables the `ansi-parsing` feature in `console`, I don't know if that breaks anything or not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = ["screenshots/*"]
 regex = { version = "1.3.1", default-features = false, features = ["std"] }
 lazy_static = "1.0"
 number_prefix = "0.4"
-console = ">=0.9.1, <1.0.0"
+console = { version = ">=0.9.1, <1.0.0", default-features = false }
 unicode-segmentation = { version = "1.6.0", optional = true }
 unicode-width = { version = "0.1.7", optional = true }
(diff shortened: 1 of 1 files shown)


djc:
I checked, and I don't think indicatif uses any of the code paths that rely on ansi-parsing in the console crate.

Thanks for your many contributions to security in Rust's ecosystem!

PR #283, opened 2021-05-14
Update expand_template regex to recognize from_dotted_str() 256 colors

MoSal:
Related to https://github.com/mitsuhiko/console/pull/98.


djc:
Thanks!

PR #284, opened 2021-05-17
Remove MultiProgress::join() based on #231

### LINKED ISSUE #125: MultiProgress should support tokio (async)

aj-bagwell:
This change moves drawing of MultiProgress from the join() thread to the child ProgressBar updates.  fixes #125, #33 and lay groud work for work on #205

**It is backwards-incompatible.**

It is based on the work of @marienz  in #231 more discussion and detail on that pull request.

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -1,14 +1,13 @@
 use std::borrow::Cow;
 use std::fmt;
 use std::io;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::mpsc::{channel, Receiver, Sender};
-use std::sync::{Arc, Mutex, RwLock, Weak};
+use std::sync::RwLock;
+use std::sync::{Arc, Mutex, Weak};
 use std::thread;
 use std::time::{Duration, Instant};
 
 use crate::state::{
-    MultiObject, MultiProgressState, ProgressDrawState, ProgressDrawTarget, ProgressDrawTargetKind,
+    MultiProgressState, ProgressDrawState, ProgressDrawTarget, ProgressDrawTargetKind,
     ProgressState, Status,
 };
 use crate::style::ProgressStyle;
@@ -507,20 +506,10 @@ impl ProgressBar {
 }
 
 /// Manages multiple progress bars from different threads.
+#[derive(Debug)]
 pub struct MultiProgress {
     state: Arc<RwLock<MultiProgressState>>,
-    joining: AtomicBool,
-    tx: Sender<(usize, ProgressDrawState)>,
-    rx: Receiver<(usize, ProgressDrawState)>,
 }
-
-impl fmt::Debug for MultiProgress {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("MultiProgress").finish()
-    }
-}
-
-unsafe impl Sync for MultiProgress {}
 
 impl Default for MultiProgress {
     fn default() -> MultiProgress {
@@ -540,18 +529,14 @@ impl MultiProgress {
 
     /// Creates a new multi progress object with the given draw target.
     pub fn with_draw_target(draw_target: ProgressDrawTarget) -> MultiProgress {
-        let (tx, rx) = channel();
         MultiProgress {
             state: Arc::new(RwLock::new(MultiProgressState {
-                objects: Vec::new(),
+                draw_states: Vec::new(),
                 free_set: Vec::new(),
                 ordering: vec![],
                 draw_target,
                 move_cursor: false,
             })),
-            joining: AtomicBool::new(false),
-            tx,
-            rx,
         }
     }
 
@@ -592,20 +577,15 @@ impl MultiProgress {
     }
 
     fn push(&self, pos: Option<usize>, pb: ProgressBar) -> ProgressBar {
-        let new = MultiObject {
-            done: false,
-            draw_state: None,
-        };
-
         let mut state = self.state.write().unwrap();
         let idx = match state.free_set.pop() {
             Some(idx) => {
-                state.objects[idx] = Some(new);
+                state.draw_states[idx] = None;
                 idx
             }
             None => {
-                state.objects.push(Some(new));
-                state.objects.len() - 1
+                state.draw_states.push(None);
+                state.draw_states.len() - 1
             }
         };
 
@@ -614,11 +594,15 @@ impl MultiProgress {
             _ => state.ordering.push(idx),
         }
 
+        assert!(
+            state.len() == state.ordering.len(),
+            "Draw state is inconsistent"
+        );
+
         pb.set_draw_target(ProgressDrawTarget {
             kind: ProgressDrawTargetKind::Remote {
                 state: self.state.clone(),
                 idx,
-                chan: Mutex::new(self.tx.clone()),
             },
         });
         pb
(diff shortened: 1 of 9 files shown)


src/progress_bar.rs
```diff
@@ -592,20 +577,15 @@ impl MultiProgress {
...
-        let new = MultiObject {
-            done: false,
-            draw_state: None,
-        };
-
         let mut state = self.state.write().unwrap();
         let idx = match state.free_set.pop() {
             Some(idx) => {
-                state.objects[idx] = Some(new);
+                state.draw_states[idx] = None;
```
djc, line 583: What exactly is the contract here for `Some` vs `None`? Other than being done, a draw state can now also be `None` if it hasn't been drawn yet? Why hold off on initializing the draw state here?

src/progress_bar.rs
```diff
@@ -643,131 +622,16 @@ impl MultiProgress {
...
+    pub fn clear(&self) -> io::Result<()> {
+        let mut state = self.state.write().unwrap();
+        let move_cursor = state.move_cursor;
+        state.draw_target.apply_draw_state(ProgressDrawState {
+            lines: vec![],
+            orphan_lines: 0,
+            finished: true,
+            force_draw: true,
+            move_cursor,
+        })?;
```
djc, line 634: It would be nice to extract this method from `join_impl()` in a separate commit before this one, so it becomes clearer how this changed.

src/state.rs
```diff
@@ -271,32 +271,75 @@ pub(crate) struct MultiProgressState {
 }
 
 impl MultiProgressState {
-    fn width(&self) -> usize {
```
djc, line 274: Please keep `width()` here to minimize changes in this large commit.

src/state.rs
```diff
@@ -503,37 +505,44 @@ impl ProgressDrawTarget {
...
+                    *last_line_count = draw_state.lines.len() - draw_state.orphan_lines;
+                    *leaky_bucket += 1.0;
+                }
+                let ticks = last_draw.elapsed().as_secs_f64()/rate.as_secs_f64();
+                *leaky_bucket -= ticks;
+                if *leaky_bucket < 0.0 {
+                    *leaky_bucket = 0.0;
+                }
+                *last_draw = Instant::now();
+                Ok(())
```
djc, line 534: Can we keep this outside the match, as before? I prefer that stylistically, and it makes this change easier to review.

djc (commented):
Thank you for working on this! In order to make sure I can review this closely, I'm asking for some changes to the commit history presented in this PR. I usually use git's interactive rebase to do that, but if you're unfamiliar with that stuff I can also help split it up more. The goal is to have a few commits that each independently could build and pass the tests.

src/state.rs
aj-bagwell: We can, and I am happy to do that if it is what you want, I moved it inside as it was the only branch that the code applied to and it needs all the fields of `ProgressDrawTargetKind::Term` so it felt cleaner to move it inside that pass them all around in a tuple.

It would now look like:
`let (term, last_line_count, rate, last_draw, leaky_bucket) = match self.kind {`

If you confirm that you prefer it that way I will do that.

aj-bagwell: I can split it into two commits if that makes it nicer

djc: Maybe it would be an improvement to wrap all those fields in a separate type instead (in a separate commit)?

djc: It looks like you made the opposite change from that I intended. I would like still like to keep the core drawing routine out of the match statement.

src/state.rs
```diff
@@ -555,13 +564,16 @@ impl ProgressDrawTarget {
...
 }
+
+const MAX_GROUP_SIZE: f64 = 32.0;
 #[derive(Debug)]
 pub(crate) enum ProgressDrawTargetKind {
     Term {
         term: Term,
         last_line_count: usize,
-        rate: Duration,
+        refresh_rate: Option<u64>,
```
djc, line 574: It's not clear to me why this is necessary. Can you make this change a separate commit?

src/progress_bar.rs
aj-bagwell: I am not sure what to initialize it to. I can't `draw()` the progressbar yet as that ends up calling back into the `MultiProgress::draw()` which needs the a lock on the `self.state` and assumes that push logic is really done.

I could give it some bogus empty draw state with no lines, but option flt semanticaly cleaner

src/state.rs
aj-bagwell: I added this after the inital pull request as I hit a divide by zero error when disabling the refresh rate.
as I had `ticks = elapsed.as_sec / rate` where `rate` is ms between ticks

So instead of doing 1/rate in the setter I decided to store it as it was passed in as Hz with `None` as no buffering.

I looked a spliting it into two commits but this commit replaces the `last_draw.elapsed() > rate` that which relies on the `let rate = refresh_rate.map(|x| Duration::from_millis(1000 / x))` so they don't really make sense separately.

src/state.rs
aj-bagwell: Done!

src/state.rs
```diff
@@ -257,9 +256,10 @@ impl Drop for ProgressState {
     }
 }
 
+#[derive(Debug)]
 pub(crate) struct MultiProgressState {
     /// The collection of states corresponding to progress bars
```
djc, line 261: Can you add some discussion here about how `None` is used both at the start and after it's finished?

src/progress_bar.rs
```diff
@@ -948,8 +803,9 @@ mod tests {
 
         let state = mp.state.read().unwrap();
         // the removed place for p1 is reused
-        assert_eq!(state.objects.len(), 2);
-        assert_eq!(state.objects.iter().filter(|obj| obj.is_some()).count(), 1);
+        assert_eq!(state.draw_states.len(), 2);
+        assert_eq!(state.free_set.len(), 1);
+        assert_eq!(state.draw_states.len() - state.free_set.len(), 1);
```
djc, line 808: Let's add a `len()` method to `MultiProgressState` to encapsulate this logic?

src/state.rs
djc: And why can't we keep it `Duration` rather than `u64`?

aj-bagwell: because it is now a frequency measured in Hz (1/s) rather than time meausred in µs

src/state.rs
aj-bagwell: Sorry I had pulled it out this way before I saw your comment, I'll do it the way you requested now.

aj-bagwell: I have moved the leaky bucket logic to it's own struct, and put the rest of the code back at the end.

boc-tothefuture:
I don't know enough about indicatif to know why it causes the defect, but the current PR does not respect when a progress bar is hidden.

```
        debug!("Bar Hidden?: {}", hidden_bar.is_hidden());
        let bar_after_add = mp.add(hidden_bar);
        debug!("Hidden after add?: {}", bar_after_add.is_hidden());
        bar_after_add.set_draw_target(ProgressDrawTarget::hidden());
        debug!("Hidden after draw target?: {}", bar_after_add.is_hidden());
```
Corresponding output:

```
Bar Hidden?: true
Hidden after add?: false
Hidden after draw target?: true
```

mibac138:
@boc-tothefuture are you sure this PR actually changed this behavior? See the [docs](https://docs.rs/indicatif/0.16.0/indicatif/struct.ProgressBar.html#method.set_draw_target) for `set_draw_target` which note this case explicitly in the current version.

boc-tothefuture:
> @boc-tothefuture are you sure this PR actually changed this behavior? See the [docs](https://docs.rs/indicatif/0.16.0/indicatif/struct.ProgressBar.html#method.set_draw_target) for `set_draw_target` which note this case explicitly in the current version.

You are correct, the same behavior exists in the currently released version.  I did not think I was experiencing that until I switched.  My apologies.

src/progress_bar.rs
aj-bagwell: Done, I added some asserts that in `push` and `remove` to check that the two lengths match, mostly to have a non test use for the new `len` method to keep the dead code check happy, but also as I think it is a good sanity check.

src/state.rs
aj-bagwell: I have changed this to Option<LeakyBucket> now to keep all the rate limiting logic in it's own new struct base on your suggestion above.

src/state.rs
```diff
@@ -257,12 +256,14 @@ impl Drop for ProgressState {
...
-    pub(crate) objects: Vec<Option<MultiObject>>,
-    /// Set of `None` elements in the `objects` vector
+    /// the state is None for bars that have not yet been drawn or have been removed
+    pub(crate) draw_states: Vec<Option<ProgressDrawState>>,
+    /// Set of removed bars, should have coreseponding `None` elements in the `draw_states` vector
```
djc, line 264: Typo: `coreseponding`.

src/progress_bar.rs
djc: Nice, thanks.

src/state.rs
```diff
@@ -570,13 +577,46 @@ impl ProgressDrawTarget {
...
+const MAX_GROUP_SIZE: f64 = 32.0;
+
+#[derive(Debug)]
+pub(crate) struct LeakyBucket {
+    leak_rate: f64,
+    last_update: Instant,
+    bucket: f64,
+}
+
+/// Rate limit but allow ocational bursts above desired rate
```
djc, line 590: Typo: `ocational`.

djc:
Except for the typos, this is looking pretty good! Thanks for working through my comments.

src/state.rs
aj-bagwell: Fixed

src/state.rs
aj-bagwell: Fixed

aj-bagwell:
Typos fixed also fixed the failing CI

djc:
I just pushed another test that calls `MultiProgress.join()`, would you mind rebasing on main and pushing that once more? I can merge it after that.

aj-bagwell:
Rebased on the the latest main (ignore the moment of stupidity where I rebase it on master instead).

Thank for all your help an patience with this.

Issue #285, opened 2021-05-17
unreachable!() statement is reachable


matthiasbeyer:
This line:

https://github.com/mitsuhiko/indicatif/blob/6a49439748cae10785cc0ec16d8e79af5e4dd47f/src/state.rs#L456

is reachable. I just hit it with the following backtrace:

```
thread 'main' panicked at 'internal error: entered unreachable code', /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:456:47
stack backtrace:                                                        
   0: rust_begin_unwind                     
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5    
   1: core::panicking::panic_fmt                                                                                                                              
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:50:5
   3: indicatif::state::ProgressDrawTarget::width
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:456:47
   4: indicatif::state::MultiProgressState::width
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:275:9
   5: indicatif::state::ProgressDrawTarget::width
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:455:65
   6: indicatif::state::ProgressState::width
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:84:9
   7: indicatif::style::ProgressStyle::format_state
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/style.rs:299:35
   8: indicatif::state::ProgressState::draw
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:236:17
   9: indicatif::state::ProgressState::update_and_draw
             at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/indicatif-0.16.0/src/state.rs:124:13
[...]

PR #286, opened 2021-05-17
Fix: Code is actually reachable

matthiasbeyer:
This fixes an reachable `unreachable!()` statement.
This might not be the approprite fix, although it resulted in no panic
in the code I used to test it, so it seems appropriate as a first idea
for a proper fix.

---

If you think this is a good first idea, please release it as a patchlevel update (`v0.16.1`)!

Related to #285 (but there might be a better fix)

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -453,7 +453,7 @@ impl ProgressDrawTarget {
         match self.kind {
             ProgressDrawTargetKind::Term { ref term, .. } => term.size().1 as usize,
             ProgressDrawTargetKind::Remote { ref state, .. } => state.read().unwrap().width(),
-            ProgressDrawTargetKind::Hidden => unreachable!(),
+            ProgressDrawTargetKind::Hidden => 0,
         }
     }
 


mibac138:
Turns out `is_hidden` must return false for `ProgressDrawTarget::Remote` because otherwise progress bars won't notify the `MultiProgress` about their state changes thus deadlocking the program. However, a progress bars notifies a `MultiProgress` by (among others) sending it's drawn state and to draw this state `width` is required (for `wide_bar` and `wide_message`). `MultiProgress` just forwards `fn width` to `draw_state.width` which actually might be `Hidden` and actually reaching the 'unreachable' branch. I believe this is the simplest solution to this issue as otherwise we'd need to add a new mechanism that skips drawing when a progress bar's draw target is remote and that remote's target is hidden, which seems relatively complex compared to this one line fix.

MCVE that deadlocks:
```rust
use indicatif::{MultiProgress, ProgressDrawTarget, ProgressBar};

fn main() {
    let mpb = MultiProgress::with_draw_target(ProgressDrawTarget::hidden());
    let pb = mpb.add(ProgressBar::new(123));
    pb.finish();
    mpb.join().unwrap();
}
```

djc:
I just published 0.16.1 which contains this fix (along with a few other ones).

PR #290, opened 2021-05-21
ProgressFolder shouldn't finish progress bars on completion

### LINKED ISSUE #289: The disappearing progress bar
I have some issues when trying to use indicatif with rayon and I end up doing my own painting. Here is the minimum reproducible thing.

```rust
// indicatif = { version = "0.16.1", features = ["rayon"] }
// rayon = "1.5.1"
use indicatif::ParallelProgressIterator;
use rayon::prelude::*;
use std::thread;
use std::time::Duration;

fn main() {
    let v: Vec<_> = (0..20).collect();
    let n = v
        .par_iter()
        .progress()
        .map(|x| {
            x % 2 == 0
        })
        .count();
    println!("{}", n);
}                                                        
```

output-opt

mibac138:
ProgressFolders represent only a part of the whole iterator and
so shouldn't finish progress bars as they don't have knowledge
of the completion of the rest of folders and otherwise could
prematurely finish a progress bar. Progress bar will still finish
when the iterator is dropped.

Fixes #289

### DIFF
--- a/src/rayon.rs
+++ b/src/rayon.rs
@@ -180,9 +180,6 @@ impl<T, C: Folder<T>> Folder<T> for ProgressFolder<C> {
     }
 
     fn complete(self) -> C::Result {
-        if !self.progress.is_finished() {
-            self.progress.finish_using_style();
-        }
         self.base.complete()
     }
 


djc:
Thanks!

PR #293, opened 2021-06-12
Fixed precision issue at high progress rates

### LINKED ISSUE #10: time_per_step() should return f64, not Duration.

kayru:
Fixes #10. 

`Estimate::time_per_step()` returns a `Duration`, which has a limited precision of 1ns. Unfortunately, this makes it impossible to handle high progress rates (>1 billion items per second).

The proposed solution is to add `seconds_per_step()` and deprecate time_per_step() in a future release. This PR intentionally leaves the old APIs in place to preserve backwards compatibility.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -5,7 +5,7 @@ use std::thread;
 use std::time::{Duration, Instant};
 
 use crate::style::{ProgressFinish, ProgressStyle};
-use crate::utils::{duration_to_secs, secs_to_duration, Estimate};
+use crate::utils::{secs_to_duration, Estimate};
 use console::Term;
 
 /// The state of a progress bar at a moment in time.
@@ -83,9 +83,15 @@ impl ProgressState {
         self.draw_target.width()
     }
 
-    /// Return the current average time per step
+    /// Return the current average time per step with nanosecond precision
+    #[allow(dead_code)]
     pub fn avg_time_per_step(&self) -> Duration {
-        self.est.time_per_step()
+        secs_to_duration(self.est.seconds_per_step())
+    }
+
+    /// Return the current average time in seconds per step with sub-nanosecond precision
+    pub fn avg_seconds_per_step(&self) -> f64 {
+        self.est.seconds_per_step()
     }
 
     /// The expected ETA
@@ -93,7 +99,7 @@ impl ProgressState {
         if self.len == !0 || self.is_finished() {
             return Duration::new(0, 0);
         }
-        let t = duration_to_secs(self.avg_time_per_step());
+        let t = self.avg_seconds_per_step();
         // add 0.75 to leave 0.25 sec of 0s for the user
         secs_to_duration(t * self.len.saturating_sub(self.pos) as f64 + 0.75)
     }
@@ -108,11 +114,11 @@ impl ProgressState {
 
     /// The number of steps per second
     pub fn per_sec(&self) -> u64 {
-        let avg_time = self.avg_time_per_step().as_nanos();
-        if avg_time == 0 {
+        let avg_time = self.avg_seconds_per_step();
+        if avg_time == 0.0 {
             0
         } else {
-            (1_000_000_000 / avg_time) as u64
+            (1.0 / avg_time) as u64
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/utils.rs
```diff
@@ -75,12 +75,18 @@ impl Estimate {
     }
 
     pub fn record_step(&mut self, value: u64) {
+        self.record_step_with_elapsed_time(value, self.start_time.elapsed())
+    }
+
+    /// This helper function takes an explicit elapsed duration so that it could also be used in
+    /// test code, which must be deterministic.
+    fn record_step_with_elapsed_time(&mut self, value: u64, elapsed: Duration) {
```
djc, line 83: Is there some way we can test this that does not depend on just extracting a single-caller method here?

src/state.rs
```diff
@@ -84,16 +84,23 @@ impl ProgressState {
     }
 
     /// Return the current average time per step
+    #[allow(dead_code)] // Keep for API compatibility. Should deprecate.
```
djc, line 87: I think I'd like to just keep this but also add some explanation to the docs here about the limitations of `Duration` resolution.

src/state.rs
```diff
@@ -84,16 +84,23 @@ impl ProgressState {
...
     pub fn avg_time_per_step(&self) -> Duration {
-        self.est.time_per_step()
+        secs_to_duration(self.est.seconds_per_step())
+    }
+
+    /// Return the current average time in seconds per step.
```
djc, line 92: Style nit: please strip the period from the first comment line (it's a sentence fragment, not a sentence), and add an empty line between the two lines.

djc (commented):
Thank you for working on this!

src/state.rs
kayru: What might be a reason to keep Duration around? It might be a source of bugs for users of the crate, similar to the bugs this PR fixes.

src/utils.rs
kayru: An alternative option that I have considered was to just add an current time Instant parameter to `record_step()` (instead of calling .elapsed() internally), however I have discarded that approach as it's changing the public API even more. I agree that single-caller helper is not ideal. It's a compromise between deterministic testability and public API churn.

djc: Actually, `Estimate` is not part of the public API at all, so I think we can just change whatever we like here.

src/state.rs
djc: I think that `Duration` as a type is easier to work with than a floating point type: it has a bunch of facilities that cover common use cases for dealing with durations, and it's more obvious to understand when you have it in your code. There are definitely use cases where the 1ns resolution is not an issue, so those users might like to get the more obvious type.

src/utils.rs
kayru: Aha, ok. `pub`-s gave the impression that it is. I'll see what `record_step()` idea looks like.

kayru: Looks a bit neater. Also improved test_time_per_step() itself a little bit.

src/state.rs
kayru: I'm definitely not going to argue about this much further and it's not a hill I'm willing to die on, but it _seems_ that average time per step can very easily get <1ns when dealing with bytes. Surfacing this to the user of the library allows them to introduce a bug that won't be caught at compile time and may not even happen until their application runs on a sufficiently fast machine (i.e. reading files from SATA vs NVMe SSD or using 1 vs 10 GbE). I'll be happy with just the minimal fix that allows built-in estimates and draw rate throttling functionality to work in my application :)

src/utils.rs
djc: Let's change `record_step` to take a `now: Instant` argument, instead.

djc:
I've gone ahead and just merged this, will iterate a bit on it.

kayru:
Yay, thank you! ❤️

PR #295, opened 2021-06-17
Implement vertical alignment for MultiProgress

nlinker:
Add an attribute to `MultiProgress` that allows us to make top (default) and bottom alignment, shown on the video:

https://user-images.githubusercontent.com/940363/122414493-011c8100-cfa9-11eb-81fc-a2a04dfd0cf0.mov


src/progress_bar.rs
```diff
@@ -489,6 +490,34 @@ impl ProgressBar {
...
+    Top,
+    Bottom,
+}
+
+impl Default for MultiProgressAlignment {
+    fn default() -> Self {
+        Self::Top
+    }
+}
+
```
djc, line 520: These definitions should probably not live in this module, let's move them into `draw_state`, below the `ProgressDrawState` definition.

src/progress_bar.rs
```diff
@@ -539,6 +569,11 @@ impl MultiProgress {
         self.state.write().unwrap().move_cursor = move_cursor;
     }
 
+    /// Set alignment flag:
```
djc, line 572: Let's ditch the colon here.

src/progress_bar.rs
```diff
@@ -609,12 +644,14 @@ impl MultiProgress {
...
         let mut state = self.state.write().unwrap();
         let move_cursor = state.move_cursor;
+        let alignment = state.alignment;
         state.draw_target.apply_draw_state(ProgressDrawState {
             lines: vec![],
             orphan_lines: 0,
             finished: true,
             force_draw: true,
             move_cursor,
+            alignment,
```
djc, line 654: Could we just assign `state.alignment` directly?

src/progress_bar.rs
```diff
@@ -489,6 +490,34 @@ impl ProgressBar {
     }
 }
 
+/// The alignment value for `MultiProgress`, the alignment controls how the multi progress
```
djc, line 493: Thanks for the great docs here! They should start with a single line that has a short one-line explanation, then an empty line, then the longer explanation.

src/draw_target.rs
```diff
@@ -368,6 +383,8 @@ pub(crate) struct ProgressDrawState {
     pub force_draw: bool,
     /// True if we should move the cursor up when possible instead of clearing lines.
     pub move_cursor: bool,
+    /// Controls how the multi progress is aligned if some its progress bars get removed, default is `Top`
```
djc, line 386: Probably just remove `its` here.

djc: Why is this duplicated between `ProgressDrawState` and `MultiProgressState`? Shouldn't one of them be enough?

examples/multi-alignment.rs
```diff
@@ -0,0 +1,299 @@
+use std::time::Duration;
```
djc, line 1: Instead of adding yet another example, can we change one of the existing ones to switch alignment based on a command-line flag? I think there are already too many examples and we probably shouldn't add any at this point (and this seems like a fairly niche feature).

djc (commented):
Thanks, I can see how this might be useful!

src/draw_target.rs
nlinker: I see no easy way to pass the flag from `MultiProgressState` to `ProgressDrawTarget::apply_draw_state` function. So I used the similar pattern, that already exists for `move_cursor` flag.

Will try to find the better way.

src/progress_bar.rs
nlinker: Unfortunately, this wouldn't compile: _error[E0502]: cannot borrow `state` as immutable because it is also borrowed as mutable_
```rust
        state.draw_target.apply_draw_state(ProgressDrawState {
            lines: vec![],
            orphan_lines: 0,
            finished: true,
            force_draw: true,
            move_cursor: state.move_cursor,
            alignment: state.alignment,
        })?;
```

examples/multi-alignment.rs
nlinker: Thanks for review! I have extended `multi-tree-ext` example instead.

src/progress_bar.rs
djc: Ah, that's what I was afraid of.

src/draw_target.rs
djc: I'm fine with leaving it for now.

djc:
Thanks!

PR #296, opened 2021-06-17
Fix rounding of time by HumanDuration

### LINKED ISSUE #294: Weird rounding by HumanDuration
Hi, first of all thank you for this very nice library.
However, a duration of 1 hour 59 min and 59 seconds is rounded down to "1 h" (and similarly for other numbers), which seems a bit surprising to me. I would expect the switch from "2 h" to "1 h" to happen at 1 hour and 30 minutes.
I think there is only a couple of lines to change in this block https://github.com/mitsuhiko/indicatif/blob/main/src/format.rs#L44 to change this behaviour. Would you mind if I sent a pull request for this?

Armavica:
Fixes #294. The behaviour before this pull request was the following (taking the example of the hour but almost every other unit would be equivalent)
```
3 hours (until 180 minutes are remaining)
2 hours (until 120 minutes are remaining)
1 hour (until 60 minutes are remaining)
59 minutes (until 59*60 seconds are remaining
58 minutes (etc.)
```
We want to replace this truncation with a proper rounding. There are several ways to go about that.

1. Perform the computations in float or in integers. I chose to use floats because of the easier rounding.
2. Decide what to display between a unit and the next. Intuitively, one could want to see:
```
3 hours (until 2 hours and 30 minutes are remaining)
2 hours (until 90 minutes are remaining)
1 hour (until 59 minutes and 30 seconds are remaining)
59 minutes (until 58 minutes and 30 seconds are remaining)
58 minutes (etc.)
```
However, the indication "1 unit" would then be displayed for more than half a unit, which is a large gap relatively to the unit. To mitigate this issue, we never display "1 unit" (except for seconds). What we actually see is then the following:
```
3 hours (until 2 hours and 30 minutes are remaining)
2 hours (until 89 minutes and 30 seconds minutes are remaining)
89 minutes (until 88 minutes and 30 seconds are remaining)
88 minutes (until 87 minutes and 30 seconds are remaining)
87 minutes (etc.)
```
3. Decide what to do below 1 second. With this rounding rather than truncating algorithm, the hack of adding 0.75 seconds to the eta does not seem useful anymore, so it was cleaned up.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -3,6 +3,13 @@ use std::time::Duration;
 
 use number_prefix::NumberPrefix;
 
+const SECOND: Duration = Duration::from_secs(1);
+const MINUTE: Duration = Duration::from_secs(60);
+const HOUR: Duration = Duration::from_secs(60 * 60);
+const DAY: Duration = Duration::from_secs(24 * 60 * 60);
+const WEEK: Duration = Duration::from_secs(7 * 24 * 60 * 60);
+const YEAR: Duration = Duration::from_secs(365 * 24 * 60 * 60);
+
 /// Wraps an std duration for human basic formatting.
 #[derive(Debug)]
 pub struct FormattedDuration(pub Duration);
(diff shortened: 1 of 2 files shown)


Armavica:
Please let me know if this behaviour seems acceptable, or if I should find a way to keep displaying `1 hour` until `59 minutes` takes over.

djc:
Alternative approach: between 2 hours and 2 minutes, should we stick to displaying minutes? Could generalize this to a rule that we skip to the next-smaller unit once we get below a value of two.

Armavica:
I implemented your suggestion. The behaviour is now the following:
```
3 hours (until 2 hours and 30 minutes are remaining)
2 hours (until 2 hours are remaining)
120 minutes (until 119 minutes and 30 seconds are remaining)
119 minutes (until 118 minutes and 30 seconds are remaining)
118 minutes (etc.)
```

Armavica:
By the way, I wonder if this hack needs to be kept with this new rounding.
https://github.com/mitsuhiko/indicatif/blob/b0079878a33a0b63023081fb0c69c84487f3c992/src/state.rs#L111-L112

src/format.rs
```diff
@@ -43,23 +43,16 @@ impl fmt::Display for FormattedDuration {
...
 impl fmt::Display for HumanDuration {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let t = self.0.as_secs();
+        let t = self.0.as_millis();
         let alt = f.alternate();
         macro_rules! try_unit {
             ($secs:expr, $sg:expr, $pl:expr, $s:expr) => {
-                let cnt = t / $secs;
-                if cnt == 1 {
+                let cnt = t / (500 * $secs);
```
djc, line 50: This is looking really good, thanks for the extensive tests!

Just one more nit: can you add a few lines of comments here that explain the strategy here? In particular, why we are using ms instead of seconds and are operating in half-seconds here.

src/format.rs
```diff
@@ -70,7 +63,13 @@ impl fmt::Display for HumanDuration {
         try_unit!(60 * 60, "hour", "hours", "h");
         try_unit!(60, "minute", "minutes", "m");
         try_unit!(1, "second", "seconds", "s");
-        write!(f, "0{}", if alt { "s" } else { " seconds" })
+        if t >= 1500 {
```
djc, line 66:
```suggestion
        if t >= 1500 { // 1500ms
```

src/format.rs
```diff
@@ -70,7 +63,13 @@ impl fmt::Display for HumanDuration {
...
+            write!(f, "2{}", if alt { "s" } else { " seconds" })
+        } else if t >= 500 {
```
djc, line 68:
```suggestion
        } else if t >= 500 { // 500ms
```

src/format.rs
Armavica: Thank you for your comment, but I managed to simplify this logic and I don't need milliseconds nor half units here any more.

src/format.rs
Armavica: I found a less ambiguous way to make these comparisons, with more types :)

Armavica:
Thank you for your review! But I think I found an even better way thanks to your comments. I was initially bothered by the fact that "1 hour" would be displayed from 1.5 to 0.5 hour, because that is too much of a gap without any information. However, I felt that switching to the next smaller unit as soon as we reach 2 units was falling a bit too much on the inverse excess. Indeed, one would see at this moment a switch between 1 significant digit (e.g. "2 hours") and 3 significant digits ("120m"). It is supposed to be a rough estimate, and probably shouldn't ever need 3 significant digits. So when I arrive at 1.5 unit, I don't display the single unit and just skip to the next smallest unit.
To sum up, it would be:
```
3 hours (until 2 hours and 30 minutes are remaining)
2 hours (until 90 minutes are remaining)
90 minutes (until 89 minutes and 30 seconds are remaining)
89 minutes (until 88 minutes and 30 seconds are remaining)
88 minutes (etc.)
```
I finally think that the rounding feels right with this algorithm.

Armavica:
There, I think that I got everything that I wanted. Sorry for the noise. I updated the first post.

src/format.rs
```diff
@@ -41,36 +41,63 @@ impl fmt::Display for FormattedDuration {
     }
 }
 
+// `HumanDuration` should be as intuitively understandable as possible.
+// So we want to round, not truncate (otherwise 1 hour and 59 minutes
```
djc, line 45: Dangling parenthesis; if the parenthesis was intended to be closed near the end of the sentence, perhaps use a semi-colon or some other way instead of the long parenthetical?

src/format.rs
```diff
@@ -41,36 +41,63 @@ impl fmt::Display for FormattedDuration {
...
+// would display an ETA of "1 hour" which underestimates the time remaining
+// by a factor 2.
+//
+// To make the precision more uniform, we avoid displaying "1 unit"
+// (except for seconds), because it would be displayed for an important
+// time relatively to the unit itself.  Instead, when we arrive around
```
djc, line 51: I find "for an important time" a bit hard to understand. Can you phrase that more simply? Also, please use single spaces after periods.

src/format.rs
```diff
@@ -41,36 +41,63 @@ impl fmt::Display for FormattedDuration {
...
+        // FIXME when `Duration::SECOND` is stable
+        let second = Duration::from_secs(1);
+        let minute = 60 * second;
+        let hour = 60 * minute;
+        let day = 24 * hour;
+        let week = 7 * day;
+        let year = 365 * day;
+        // FIXME when `div_duration_f64` is stable
+        let t = self.0.as_secs_f64();
+        if self.0 + week / 2 >= year + year / 2 {
```
djc, line 71: There's a lot of repetition here. I'm pretty sure this could be written as a loop given the right const input tuples...

src/format.rs
```diff
@@ -100,3 +127,242 @@ impl fmt::Display for BinaryBytes {
...
+    use super::*;
+
+    const NANO: Duration = Duration::from_nanos(1);
+    const MILLI: Duration = Duration::from_millis(1);
+    const SECOND: Duration = Duration::from_secs(1);
+    const MINUTE: Duration = Duration::from_secs(60);
+    const HOUR: Duration = Duration::from_secs(60 * 60);
+    const DAY: Duration = Duration::from_secs(24 * 60 * 60);
+    const WEEK: Duration = Duration::from_secs(7 * 24 * 60 * 60);
+    const YEAR: Duration = Duration::from_secs(365 * 24 * 60 * 60);
```
djc, line 142: Why not pull these out of `tests` and use them in the `fmt::Display` impl as well, rather than redefining them every time?

src/format.rs
```diff
@@ -100,3 +127,242 @@ impl fmt::Display for BinaryBytes {
...
+
+    #[test]
+    fn human_duration_less_than_one_second() {
```
djc, line 145: Let's test `Debug` and `Display` once per function and pick one to run all the other tests. Then, we can also inline the value definition, which makes everything a lot more compact.

Armavica:
Thank you for your detailed review! I applied all of your comments and pushed a new version.

djc:
Thank you! I have more nits to pick but I'll just tweak some things myself.

PR #297, opened 2021-06-20
once_cell

burrbull:


djc:
Why is this an improvement?

burrbull:
> Why is this an improvement?

`once_cell` is on the way to be standard Rust lazy type.
https://github.com/rust-lang/rust/issues/74465

djc:
I subscribe to that issue; it has been "on the way" for a year now, with no particularly concrete end in sight. Just that it's headed for std doesn't seem like a compelling reason to switch.

djc:
In one larger workspace I work on, lazy_static was in the Cargo.lock 19 times whereas lazy_static was there only 4 times. On net, this therefore seems likely to increase people's dependencies. Until it's actually part of the standard library (of the minimum supported Rust version), I think it's better to stick with lazy_static.

djc:
(Still open to more compelling rationales for changing it, but haven't seen one here so far.)

PR #301, opened 2021-07-06
Add ability to customize template keys

redzic:
This PR makes the `pos` and `len` fields of `ProgressState` to have `pub` visibility. I changed the `per_sec` method of `ProgressState` to return an `f64`, and `{per_sec}` in the template now rounds to an integer via the `format!` macro (rather than truncating by default), and is explicitly truncated in other places of the code. With this patch, you can now use `.key(...)` method from `ProgressStyle` to add a custom formatting function and template key. This PR also adds another example to show the usage of this feature.

I also changed the `position` method to just be 2 separate methods for returning the position and length, because the usage beforehand where `position()` was called before matching the template keys and then you would just reuse those for `{pos}` and `{len}` now seems like an anti-pattern with this hash map based approach.  

Also I had to remove `#[derive(Debug)]` for `ProgressStyle` because I was getting errors of "`Debug` impl is not general enough" or something like that, and it didn't seem to be necessary anyway.

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -187,7 +187,9 @@ impl ProgressBar {
     pub fn set_draw_rate(&self, n: u64) {
         let mut state = self.state.lock().unwrap();
         state.draw_rate = n;
-        state.draw_next = state.pos.saturating_add(state.per_sec() / n);
+        state.draw_next = state
+            .pos
+            .saturating_add((state.per_sec() / n as f64) as u64);
     }
 
     /// Manually ticks the spinner or progress bar
@@ -477,7 +479,7 @@ impl ProgressBar {
     }
 
     /// Returns the current rate of progress
-    pub fn per_sec(&self) -> u64 {
+    pub fn per_sec(&self) -> f64 {
         self.state.lock().unwrap().per_sec()
     }
 
--- a/src/state.rs
+++ b/src/state.rs
@@ -8,10 +8,10 @@ use crate::draw_target::{ProgressDrawState, ProgressDrawTarget};
 use crate::style::{ProgressFinish, ProgressStyle};
 
 /// The state of a progress bar at a moment in time.
-pub(crate) struct ProgressState {
+pub struct ProgressState {
     pub(crate) style: ProgressStyle,
-    pub(crate) pos: u64,
-    pub(crate) len: u64,
+    pub pos: u64,
+    pub len: u64,
     pub(crate) tick: u64,
     pub(crate) started: Instant,
     pub(crate) draw_target: ProgressDrawTarget,
@@ -82,11 +82,6 @@ impl ProgressState {
         pct.max(0.0).min(1.0)
     }
 
-    /// Returns the position of the status bar as `(pos, len)` tuple.
-    pub fn position(&self) -> (u64, u64) {
-        (self.pos, self.len)
-    }
-
     /// Returns the current message of the progress bar.
     pub fn message(&self) -> &str {
         &self.message
@@ -120,12 +115,12 @@ impl ProgressState {
     }
 
     /// The number of steps per second
-    pub fn per_sec(&self) -> u64 {
-        let avg_time = self.est.seconds_per_step();
-        if avg_time == 0.0 {
-            0
+    pub fn per_sec(&self) -> f64 {
+        let per_sec = 1.0 / self.est.seconds_per_step();
+        if per_sec.is_nan() {
+            0.0
         } else {
-            (1.0 / avg_time) as u64
+            per_sec
         }
     }
 
@@ -157,7 +152,7 @@ impl ProgressState {
         }
         if new_pos >= self.draw_next {
             self.draw_next = new_pos.saturating_add(if self.draw_rate != 0 {
-                self.per_sec() / self.draw_rate
+                (self.per_sec() / self.draw_rate as f64) as u64
             } else {
                 self.draw_delta
             });
--- a/src/style.rs
+++ b/src/style.rs
@@ -4,12 +4,24 @@ use crate::format::{BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, Hu
 use crate::state::ProgressState;
 use crate::utils::{expand_template, pad_str};
 use std::borrow::Cow;
+use std::collections::HashMap;
 
 #[cfg(feature = "improved_unicode")]
 use unicode_segmentation::UnicodeSegmentation;
 
+pub type Format = fn(&ProgressState) -> String;
+
+#[derive(Clone)]
+pub(crate) struct FormatMap(HashMap<&'static str, Format>);
+
+impl Default for FormatMap {
+    fn default() -> Self {
+        FormatMap(HashMap::new())
+    }
+}
+
 /// Controls the rendering style of progress bars
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct ProgressStyle {
     tick_strings: Vec<Box<str>>,
     progress_chars: Vec<Box<str>>,
@@ -17,6 +29,7 @@ pub struct ProgressStyle {
     on_finish: ProgressFinish,
     // how unicode-big each char in progress_chars is
     char_width: usize,
+    format_map: FormatMap,
 }
 
 #[cfg(feature = "improved_unicode")]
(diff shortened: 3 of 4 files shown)


examples/custom-template.rs
```diff
@@ -0,0 +1,26 @@
+use std::thread;
```
djc, line 1: Please use the new feature in an existing example, and please do it in a separate commit.

src/progress_bar.rs
```diff
@@ -187,7 +187,7 @@ impl ProgressBar {
     pub fn set_draw_rate(&self, n: u64) {
         let mut state = self.state.lock().unwrap();
         state.draw_rate = n;
-        state.draw_next = state.pos.saturating_add(state.per_sec() / n);
+        state.draw_next = state.pos.saturating_add(state.per_sec() as u64 / n);
```
djc, line 190: I think we would get better data if the conversion to `u64` wraps the division, right?

src/state.rs
```diff
@@ -82,9 +82,14 @@ impl ProgressState {
         pct.max(0.0).min(1.0)
     }
 
-    /// Returns the position of the status bar as `(pos, len)` tuple.
-    pub fn position(&self) -> (u64, u64) {
-        (self.pos, self.len)
+    /// Returns the position of the status bar.
```
djc, line 85: If we're going to make the fields public we should just get rid of getter(s).

src/state.rs
```diff
@@ -120,12 +125,12 @@ impl ProgressState {
     }
 
     /// The number of steps per second
-    pub fn per_sec(&self) -> u64 {
-        let avg_time = self.est.seconds_per_step();
-        if avg_time == 0.0 {
-            0
+    pub fn per_sec(&self) -> f64 {
+        let per_sec = 1.0 / self.est.seconds_per_step();
```
djc, line 129: Why did you change the structure like this? I think I prefer comparing to zero rather than `is_nan()`.

src/state.rs
```diff
@@ -157,7 +162,7 @@ impl ProgressState {
         }
         if new_pos >= self.draw_next {
             self.draw_next = new_pos.saturating_add(if self.draw_rate != 0 {
-                self.per_sec() / self.draw_rate
+                self.per_sec() as u64 / self.draw_rate
```
djc, line 165: Cast after division?

src/style.rs
```diff
@@ -4,19 +4,81 @@ use crate::format::{BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, Hu
 use crate::state::ProgressState;
 use crate::utils::{expand_template, pad_str};
 use std::borrow::Cow;
+use std::collections::HashMap;
 
 #[cfg(feature = "improved_unicode")]
 use unicode_segmentation::UnicodeSegmentation;
 
+pub type Format = fn(&ProgressState) -> String;
```
djc, line 12: I don't think the `FormatMap` wrapper is useful here, since it's private to the crate anyway. Let's just rename `FormatMapInner` to `FormatMap`.

src/style.rs
```diff
@@ -129,6 +193,12 @@ impl ProgressStyle {
         self
     }
 
+    /// Adds a custom key parser that references a `&ProgressState` to the template.
+    pub fn key(mut self, key: &'static str, f: Format) -> ProgressStyle {
```
djc, line 197: Let's call this `with_key()` (and please remove the period from the docstring).

src/style.rs
```diff
@@ -4,19 +4,81 @@ use crate::format::{BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, Hu
...
 #[cfg(feature = "improved_unicode")]
 use unicode_segmentation::UnicodeSegmentation;
 
+pub type Format = fn(&ProgressState) -> String;
+type FormatMapInner = HashMap<&'static str, Format>;
+
+#[derive(Clone)]
+pub(crate) struct FormatMap(FormatMapInner);
+
+impl Default for FormatMap {
```
djc, line 18: I think we should leave the built-in formatters outside of the format map, the map should be used only to augment/replace the default set of keys. In `expand_template()`, let's cache whether the map is empty and only do lookups if it isn't.

djc (commented):
Looking pretty good, I have a bunch of smallish suggestions.

src/state.rs
redzic: Two reasons: even when comparing the denominator to zero, I sometimes get `NaN` instead of 0 as the result anyway. Probably due to using `==` on floating point values. So, semantically, I think that comparing to `NaN` is more accurate.

Other reason is that the codegen seems a bit better when comparing to `NaN`. The compiler only generates branchless code when comparing to NaN, and there are fewer instructions overall. https://godbolt.org/z/f4fhPs7ec

djc: Sounds good!

examples/custom-template.rs
redzic: Thanks, deleted the example for now. Will modify an existing example in another commit later.

src/progress_bar.rs
redzic: Fixed. I cast both values to `f64` now and then cast to `u64` at the end.

src/state.rs
redzic: Removed these methods.

src/state.rs
redzic: Thanks, fixed.

src/style.rs
redzic: Done.

src/style.rs
redzic: Removed `FormatMapInner`.

src/style.rs
redzic: Alright, that makes sense. However, I'm not really sure how you would do this in `expand_template` as it is right now. Should `expand_template` also take a reference to the hash map of custom keys as well? Also, how would it know if the key was not found without changing the `format_state` function? Currently, it just falls back to returning an empty string and just using that in the template. Perhaps this could be changed to return an `Option<String>`?

To me, it seems easier to just do all of that inside of the closure in `format_state`. It doesn't really seem necessary to cache whether the map is empty, as it seems very cheap to check for that by just calling `.is_empty()`: https://godbolt.org/z/M5fEPjWE4

djc: Ah, sorry, I guess I mean sort of like it is now, in the closure passed to `expand_template()`.

redzic: Ok, will do. One implication of that approach is that the default keys cannot be overridden, because they are always matched first. Perhaps this should be mentioned in the documentation?

redzic: The fact that default template keys are not overridable is now mentioned in the docstring of `with_key`.

redzic:
@djc should be ready to review

src/progress_bar.rs
```diff
@@ -478,7 +480,7 @@ impl ProgressBar {
 
     /// Returns the current rate of progress
     pub fn per_sec(&self) -> u64 {
-        self.state.lock().unwrap().per_sec()
+        self.state.lock().unwrap().per_sec() as u64
```
djc, line 483: How do you feel about just making this also return `f64`?

src/style.rs
```diff
@@ -261,13 +283,24 @@ impl ProgressStyle {
...
-                "bytes_per_sec" => format!("{}/s", HumanBytes(state.per_sec())),
-                "binary_bytes_per_sec" => format!("{}/s", BinaryBytes(state.per_sec())),
+                "bytes_per_sec" => format!("{}/s", HumanBytes(state.per_sec() as u64)),
+                "binary_bytes_per_sec" => format!("{}/s", BinaryBytes(state.per_sec() as u64)),
                 "eta_precise" => format!("{}", FormattedDuration(state.eta())),
                 "eta" => format!("{:#}", HumanDuration(state.eta())),
                 "duration_precise" => format!("{}", FormattedDuration(state.duration())),
                 "duration" => format!("{:#}", HumanDuration(state.duration())),
-                _ => "".into(),
+                key => {
```
djc, line 292: Put together like this, the separate `format_map.0.is_empty()` case looks silly. Let's assume that `.get()` will be reasonably quick in the case of a zero-length map.

I propose that we do `if let Some(_) = self.format.map.0.get(var)` before the `match var.key`, so that we can override the default keys.

redzic: Seems reasonable. Hash table lookups, especially on small strings, should be quite fast.

src/progress_bar.rs
redzic: `f64` is probably more consistent, so I think it'd be better.

src/style.rs
redzic: This change has been reverted, as the default keys are now overridable.

PR #306, opened 2021-08-09
Adding ProgressIterator::progress_with_style() and taking care of minor typos and clippy lints

### LINKED ISSUE #305: Style ProgressIterator without overwriting length
Hi, 

I'm trying to use a `ProgressIterator` by following the example given in the [documentation](https://docs.rs/indicatif/0.16.2/indicatif/#iterators): 

```
use indicatif::ProgressIterator;

for _ in (0..1000).progress() {
    // ...
}
```

I would like to style my progress bar, however. A bit further down in the documentation, the following snippet is given: 

```
use indicatif::{ProgressBar, ParallelProgressIterator};
use rayon::iter::{ParallelIterator, IntoParallelRefIterator};

// Use `ProgressBar::with_style` to change the view
let pb = ProgressBar::new();
let v: Vec<_> = (0..100000).collect();
let v2: Vec<_> = v.par_iter().progress_with(pb).map(|i| i + 1).collect();
assert_eq!(v2[0], 1);
```

Unlike the example, however, I cannot call `ProgressBar::new()` without specifying a length. Is this limited to these parallel iterators? I only include `indicatif::{ProgressBar, ProgressIterator, ProgressStyle}`. 

So, basically, I just want to provide my `ProgressIterator` with a `ProgressStyle`, but it appears that I need to do that in the form of a new `ProgressBar`. That would be fine, if I could do that without overwriting the length that is automatically determined by the `ProgressIterator`. Is that possible?

LeCyberDucky:
I've added `ProgressIterator::progress_with_style()`, as discussed in https://github.com/mitsuhiko/indicatif/issues/305. Further, I have corrected a small typo and taken care of two clippy warnings. 

Resolves #305

### DIFF
--- a/examples/multi-tree-ext.rs
+++ b/examples/multi-tree-ext.rs
@@ -224,7 +224,7 @@ pub fn main() {
                     let pb = mp2.insert(item.index, item.progress_bar.clone());
                     pb.set_prefix("  ".repeat(item.indent));
                     pb.set_message(&item.key);
-                    items.insert(item.index, &item);
+                    items.insert(item.index, item);
                 }
                 Elem::RemoveItem(Index(index)) => {
                     let item = items.remove(*index);
--- a/examples/multi-tree.rs
+++ b/examples/multi-tree.rs
@@ -115,7 +115,7 @@ fn main() {
                     let elem = &ELEMENTS[el_idx];
                     let pb = mp2.insert(elem.index + 1, elem.progress_bar.clone());
                     pb.set_message(format!("{}  {}", "  ".repeat(elem.indent), elem.key));
-                    tree.lock().unwrap().insert(elem.index, &elem);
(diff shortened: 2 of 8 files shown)


djc (approved):

djc:
Looks great, thanks!

LeCyberDucky:
Cool! I forgot one thing, though: With this in place, perhaps the snippet in the documentation that led me here in the first place could/should also be updated, since it still shows `ProgressBar::new()` without a length argument. I'm not sure where that snippet is located in the repositor, though.

djc:
It's here: https://github.com/mitsuhiko/indicatif/blob/main/src/lib.rs#L78.

djc:
Would you mind fixing the final clippy lint issue, too? Thanks!

djc:
I squashed your commits a little bit and rebased them onto main. Thanks!

PR #307, opened 2021-09-11
Add an example demonstrating a custom spinner and Tokio

silvioprog:


silvioprog:
Example running on Ubuntu:

Peek 2021-09-11 02-13

silvioprog:
Any change to apply this example? 😅

djc:
3 days is a little soon to ping again IMO -- I'm just a volunteer maintaining this.

As for the change itself, I think indicatif already has a lot of examples, and in particular it already has a tokio example and a spinner example. So adding this as yet another example would not seem to be an improvement to me -- it just clutters up the examples directory even more.

djc:
Going to close this for now.

PR #308, opened 2021-09-11
Added AsyncRead and AsyncWrite support for ProgressBar

x0f5c3:
Added support for tokio Read, Write and Seek traits to allow for using the ProgressBar with tokio async io traits by wrapping them in ProgressBarIter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ console = { version = ">=0.9.1, <1.0.0", default-features = false }
 unicode-segmentation = { version = "1.6.0", optional = true }
 unicode-width = { version = "0.1.7", optional = true }
 rayon = { version = "1.0", optional = true }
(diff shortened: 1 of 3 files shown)


djc:
Thanks, this seems reasonable! I'm not sure the tests are exercising anything particularly useful: `ProgressBarIter` itself does constrain the type its inner `T`, so the test doesn't really help in this case. Maybe expand the test a little to check that the `ProgressBarIter<T>` is `AsyncRead` if `T` is `AsyncRead` and similar for `AsyncWrite`?

x0f5c3:
I think the CI job should pass now since I moved the async related uses under `#[cfg(feature = "tokio")]`

djc:
Just a little formatting left.

x0f5c3:
Sorry for that, forgot to run cargo fmt, should be fixed now

djc:
Thanks!

PR #309, opened 2021-09-11
Pass through stream_position in ProgressBarIter

rlee287:
This avoids sending a stream_position update and also preserves optimizations that the underlying object may have made for stream_position.

For example, when wrapping a `BufReader` object, adding this impl allows the use of `.stream_position()` in functions that take a generic `Seek` object without causing the performance drop associated with the default impl of `stream_position` calling `seek(SeekFrom::Current(0))` and throwing away the buffer each time.


djc:
This makes sense, thanks!

PR #314, opened 2021-09-19
Add sleep in iterator example to see bars

bidoubiwa:
I've added the sleep to be able to see the bars when I run the example. They were very quickly done without a `sleep`.


djc:
Makes sense, thanks!

PR #315, opened 2021-10-02
Add AsyncBufRead implementation for ProgressBarIter

x0f5c3:
Added AsyncBufRead implementation for ProgressBarIter<W: tokio::io::AsyncBufRead + tokio::io::AsyncRead + Unpin>

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,11 +1,11 @@
 use crate::progress_bar::ProgressBar;
 use std::convert::TryFrom;
-use std::io::{self, IoSliceMut, Error, SeekFrom};
+use std::io::{self, IoSliceMut};
 use std::iter::FusedIterator;
-use std::pin::Pin;
-use std::task::{Context, Poll};
 #[cfg(feature = "tokio")]
-use tokio::io::ReadBuf;
+use tokio::io::{ReadBuf, SeekFrom};
+#[cfg(feature = "tokio")]
+use std::{pin::Pin, task::{Context, Poll}};
 
 /// Wraps an iterator to display its progress.
 pub trait ProgressIterator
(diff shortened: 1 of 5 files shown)


src/iter.rs
```diff
@@ -199,6 +201,26 @@ impl<W: tokio::io::AsyncSeek + Unpin> tokio::io::AsyncSeek for ProgressBarIter<W
...
+    fn poll_fill_buf(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<&[u8]>> {
+        let pb = self.progress.clone();
+        Pin::new(self.get_mut().it.borrow_mut())
+            .poll_fill_buf(cx)
+            .map(|x| {
+                x.map(|buf| {
+                    pb.inc(buf.len() as u64);
+                    buf
+                })
+            })
```
djc, line 217: We can spell it like this:

```suggestion
        let this = self.get_mut();
        let result = Pin::new(&mut this.it).poll_fill_buf(cx);
        if let Poll::Ready(Ok(buf)) = &result {
            this.progress.inc(buf.len() as u64);
        }
        result
```

This is a fair bit more compact, easier to follow (IMO) and avoids the `ProgressBar` clone. What do you think?

src/iter.rs
```diff
@@ -199,6 +201,26 @@ impl<W: tokio::io::AsyncSeek + Unpin> tokio::io::AsyncSeek for ProgressBarIter<W
         Pin::new(&mut self.it).poll_complete(cx)
     }
 }
+#[cfg(feature = "tokio")]
```
djc, line 204: Please keep an empty line around between the implementations.

src/iter.rs
```diff
@@ -1,4 +1,6 @@
 use crate::progress_bar::ProgressBar;
+#[cfg(feature = "tokio")]
```
djc, line 2: If you accept the other change, this won't be necessary anymore.

djc (commented):
Thanks, this makes sense! Suggestion for a different way to write it:

src/iter.rs
x0f5c3: Removed

src/iter.rs
x0f5c3: I agree, it was pretty workaroundish in my version, just to satisfy the borrow checker

PR #316, opened 2021-10-18
Enable `ansi-parsing` feature for `console`

### LINKED ISSUE #300: If using colors and bold, wide_bar doesn't visually fill up the entire terminal width
Basically, it seems like the color and boldness are included in the width calculation, even though they shouldn't be, so that `wide_bar` can actually visually use the full width of the terminal when using color.

Here is a modified example of finebars to show the discrepancy between the colored and colorless versions in the width:

```rust
use std::thread;
use std::time::Duration;

use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
use rand::{thread_rng, Rng};

fn main() {
    let styles = [
        ("Rough bar:", "█  ", "red"),
        ("Fine bar: ", "█▉▊▋▌▍▎▏  ", "yellow"),
        ("Vertical: ", "█▇▆▅▄▃▂▁  ", "green"),
        ("Fade in:  ", "█▓▒░  ", "blue"),
        ("Blocky:   ", "█▛▌▖  ", "magenta"),
    ];

    {
        let m = MultiProgress::new();

        let handles: Vec<_> = styles
            .iter()
            .map(|s| {
                let pb = m.add(ProgressBar::new(512));
                pb.set_style(
                    ProgressStyle::default_bar()
                        .template("{prefix}▕{wide_bar}▏{msg}")
                        .progress_chars(s.1),
                );
                pb.set_prefix(s.0);
                let wait = Duration::from_millis(thread_rng().gen_range(0..3));
                thread::spawn(move || {
                    for i in 0..512 {
                        pb.inc(1);
                        pb.set_message(format!("{:3}%", 100 * i / 512));
                        thread::sleep(wait);
                    }
                    pb.finish_with_message("100%");
                })
            })
            .collect();

        for h in handles {
            let _ = h.join();
        }
    }

    {
        let m = MultiProgress::new();

        let handles: Vec<_> = styles
            .iter()
            .map(|s| {
                let pb = m.add(ProgressBar::new(512));
                pb.set_style(
                    ProgressStyle::default_bar()
                        .template(&format!("{{prefix:.bold}}▕{{wide_bar:.{}}}▏{{msg}}", s.2))
                        .progress_chars(s.1),
                );
                pb.set_prefix(s.0);
                let wait = Duration::from_millis(thread_rng().gen_range(0..3));
                thread::spawn(move || {
                    for i in 0..512 {
                        pb.inc(1);
                        pb.set_message(format!("{:3}%", 100 * i / 512));
                        thread::sleep(wait);
                    }
                    pb.finish_with_message("100%");
                })
            })
            .collect();

        for h in handles {
            let _ = h.join();
        }
    }
}
```

redzic:
Fixes #300.

As it turns out, `console`'s `strip_ansi_codes` function is a no-op unless the `ansi-parsing` feature is enabled (https://docs.rs/console/0.15.0/src/console/utils.rs.html#10-16).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = ["screenshots/*"]
 regex = { version = "1.3.1", default-features = false, features = ["std"] }
 lazy_static = "1.0"
 number_prefix = "0.4"
-console = { version = "0.15.0", default-features = false }
+console = { version = "0.15.0", default-features = false, features = ["ansi-parsing"] }
 unicode-segmentation = { version = "1.6.0", optional = true }
 unicode-width = { version = "0.1.7", optional = true }
 rayon = { version = "1.0", optional = true }


djc:
Would it be feasible to add a test for this?

PR #324, opened 2021-11-16
Use once_cell instead of lazy_static

sigmaSd:
The motivation is that console already uses once_cell, so this should in theory decrease the total number of dependencies.

Note: The formatting has changed, do you happen to use some rustfmt flags? if so it would be better if you have a `.rustfmt.toml` file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ exclude = ["screenshots/*"]
 
 [dependencies]
 regex = { version = "1.3.1", default-features = false, features = ["std"] }
-lazy_static = "1.0"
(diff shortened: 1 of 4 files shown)


djc:
The formatting changed because rustfmt usually doesn't work inside calls to macros, and you moved those constants out of macro context. So that's a good thing. Makes sense to me to make this change, thanks!

sigmaSd:
I see, thanks as well

djc:
(Note that #319 moves lazy_static to be for examples only, so if I figure out the regression it introduced we'll be able to do without once_cell for the library itself.)

PR #325, opened 2021-11-17
ProgressBar: Add with_elapsed for overridding elapsed time

zhaofengli:
This PR adds `ProgressBar::with_elapsed()` which provides a way to override the elapsed time of a progress bar.

This is useful for scenarios where you dynamically create additional ProgressBars for "sub-tasks" and want to preserve the elapsed time in the output. Since ProgressBar exposes the elapsed time as a Duration, `with_elapsed` is more ergonomic than `with_start_time`.

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -77,6 +77,12 @@ impl ProgressBar {
         self
     }
 
+    /// A convenience builder-like function for a progress bar with a given elapsed time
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks, this makes sense! (Especially thanks for the pre-emptive note about the `with_start_time()` alternative.)

PR #327, opened 2021-11-27
[Request for comment; not ready for merge]: add insert_after / insert_before to MultiProgress

chris-laplante:
This greatly simplifies the code. It also clarifies the API very nicely:
the old MultiProgress::insert method is not practically useful since the
indices do not necessarilly correspond to visual order, except for
specific usage patterns.

The ability to insert after/before particular progress bars is much more
useful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ regex = { version = "1.3.1", default-features = false, features = ["std"] }
 once_cell = "1.8.0"
 number_prefix = "0.4"
 console = { version = "0.15.0", default-features = false, features = ["ansi-parsing"] }
+generational_token_list = "0.1.1"
 unicode-segmentation = { version = "1.6.0", optional = true }
 unicode-width = { version = "0.1.7", optional = true }
 rayon = { version = "1.0", optional = true }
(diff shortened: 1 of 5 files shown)


chris-laplante:
This pull request leverages a new crate I developed (https://github.com/Agilent/generational_token_list). 

Still to do here:

- [ ] Update examples
- [ ] Documentation

djc:
I'm on the fence about using your crate. Basically, this seems to exchange the use of indices (with panics at run-time if we get it wrong) for something more efficient that relies on unsafe code. Since our use of this functionality seems fairly limited, I'm not sure it makes sense to use something like your crate over just using indexes. We could wrap the current code in a separate type to make it easier to verify.

Even if your crate was very well tested (with comprehensive test coverage including Miri) and audited by someone conversant in unsafe, I'm not sure I'd be up for doing this here.

mostthingsweb:
The only use of `unsafe` is in `GenerationalTokenList::iter_mut`, which this crate isn't using. I'd be happy to gate the `unsafe` code behind an "unsafe" feature (and in fact I'll probably do that anyway).

djc:
That would be great! I do agree the `insert_after()`/`insert_before()` design makes a lot more sense, but taken on a new/immature extra dependency to do it feels like not quite worth it.

chris-laplante:
See #331 instead

chris-laplante:
@djc Do you think there's any chance I could reconstitute this changeset (i.e. using `[generational_token_list](https://github.com/Agilent/generational_token_list)`). It would alleviate #419 and also make my life with #426 easier.

djc:
Let's discuss this in a new issue? Can you layout the core problem and why the generational list is your preferred way to solve it?

PR #329, opened 2021-12-08
Document interaction between `wide_bar` and `wide_msg`

sigmaSd:
Fix https://github.com/mitsuhiko/indicatif/issues/328


djc:
Thanks!

PR #330, opened 2021-12-10
Add categories keyword to Cargo.toml

jqnatividad:
indicatif is already a very popular crate, but this helps with discoverability


djc:
Good call, thanks!

PR #331, opened 2021-12-14
add insert_after / insert_before to MultiProgress

chris-laplante:
Also introduce internal InsertLocation enum to clarify MultiProgress:push


chris-laplante:
Here's a rough cut of what I propose :)

src/progress_bar.rs
```diff
@@ -633,10 +641,28 @@ impl MultiProgress {
...
+    /// Inserts a progress bar after an existing one.
+    ///
+    /// The progress bar added will have the draw target changed to a
+    /// remote draw target that is intercepted by the multi progress
+    /// object overriding custom `ProgressDrawTarget` settings.
+    pub fn insert_after(&self, after: &ProgressBar, pb: ProgressBar) -> ProgressBar {
+        self.push(InsertLocation::After(after), pb)
+    }
+
+    fn push(&self, pos: InsertLocation, pb: ProgressBar) -> ProgressBar {
```
djc, line 665: Let's not call the `InsertLocation` `pos` here? That seems confusing.

src/progress_bar.rs
```diff
@@ -558,6 +558,14 @@ impl Default for MultiProgress {
     }
 }
 
+enum InsertLocation<'a> {
```
djc, line 561: Style nit: please move this definition down below the `impl WeakProgressBar`.

examples/multi-insert-relative.rs
```diff
@@ -0,0 +1,75 @@
+use indicatif::{MultiProgress, ProgressBar, ProgressFinish, ProgressStyle};
```
djc, line 1: One of my pet peeves is that indicatif already has too many examples. Do you think you could modify one of the existing examples to use `insert_before()` and/or `insert_after()`?

djc (commented):
Thanks, this looks good! Some comments below.

src/progress_bar.rs
chris-laplante: Done

src/progress_bar.rs
chris-laplante: Done

examples/multi-insert-relative.rs
chris-laplante: Sure, I can modify the "multi.rs" example and will do so now. As I alluded in the previous version of this pull request, the other multi-* examples are too complicated to change to use `insert_before`/`insert_after`.

chris-laplante: Done

djc:
Thanks!

chris-laplante:
Thank you too :)

PR #333, opened 2021-12-16
Implement ProgressBar::suspend

ishitatsuyuki:
This function hides the progress bar temporarily, execute `f`, then redraws the progress bar, which is useful for external code that writes to the standard output.

The idea is originally raised in #92.

It is planned to eventually create a global function like `ProgressBar::instance` that allows calling `suspend` from e.g. a log handler, but this currently doesn't generalize to `MultiProgressBar` which could be an obstacle.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -152,7 +152,7 @@ impl ProgressDrawTarget {
                 ref mut last_line_count,
                 leaky_bucket: Some(ref mut leaky_bucket),
             } => {
-                if draw_state.finished || draw_state.force_draw || leaky_bucket.try_add_work() {
+                if draw_state.force_draw || leaky_bucket.try_add_work() {
                     (term, last_line_count)
                 } else {
                     // rate limited
@@ -205,8 +205,7 @@ impl ProgressDrawTarget {
                         ProgressDrawState {
                             lines: vec![],
                             orphan_lines: 0,
-                            finished: true,
-                            force_draw: false,
+                            force_draw: true,
                             move_cursor: false,
                             alignment: Default::default(),
                         },
@@ -262,7 +261,7 @@ impl MultiProgressState {
     }
 
     pub(crate) fn draw(&mut self, idx: usize, draw_state: ProgressDrawState) -> io::Result<()> {
-        let force_draw = draw_state.finished || draw_state.force_draw;
+        let force_draw = draw_state.force_draw;
         let mut orphan_lines = vec![];
 
         // Split orphan lines out of the draw state, if any
@@ -301,18 +300,12 @@ impl MultiProgressState {
             }
         }
 
-        // !any(!done) is also true when iter() is empty, contrary to all(done)
-        let finished = !self
-            .draw_states
-            .iter()
-            .any(|x| !x.as_ref().map(|s| s.finished).unwrap_or(false));
         self.draw_target.apply_draw_state(ProgressDrawState {
             lines,
             orphan_lines: orphan_lines_count,
             force_draw: force_draw || orphan_lines_count > 0,
             move_cursor: self.move_cursor,
             alignment: self.alignment,
-            finished,
         })
     }
 
@@ -376,8 +369,6 @@ pub(crate) struct ProgressDrawState {
     pub lines: Vec<String>,
     /// The number of lines that shouldn't be reaped by the next tick.
     pub orphan_lines: usize,
-    /// True if the bar no longer needs drawing.
-    pub finished: bool,
     /// True if drawing should be forced.
     pub force_draw: bool,
     /// True if we should move the cursor up when possible instead of clearing lines.
@@ -387,12 +378,11 @@ pub(crate) struct ProgressDrawState {
 }
 
 impl ProgressDrawState {
-    pub(crate) fn new(lines: Vec<String>, finished: bool) -> Self {
+    pub(crate) fn new(lines: Vec<String>, force_draw: bool) -> Self {
         Self {
             lines,
             orphan_lines: 0,
-            finished,
-            force_draw: false,
+            force_draw,
             move_cursor: false,
             alignment: Default::default(),
         }
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -138,7 +138,7 @@ impl ProgressBar {
                 }
                 ms = state.steady_tick;
 
-                state.draw().ok();
+                state.draw(false).ok();
             } else {
                 break;
             }
@@ -248,7 +248,6 @@ impl ProgressBar {
         let draw_state = ProgressDrawState {
             lines,
             orphan_lines,
-            finished: state.is_finished(),
             force_draw: true,
             move_cursor: false,
             alignment: Default::default(),
@@ -728,7 +727,6 @@ impl MultiProgress {
         state.draw_target.apply_draw_state(ProgressDrawState {
             lines: vec![],
             orphan_lines: 0,
-            finished: true,
             force_draw: true,
             move_cursor,
             alignment,
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -231,11 +231,14 @@ impl ProgressBar {
 
     /// Print a log line above the progress bar
     ///
+    /// If the progress bar is hidden (e.g. when standard output is not a terminal), `println()`
+    /// will not do anything. If you want to write to the standard output in such cases as well, use
+    /// [`suspend`] instead.
+    ///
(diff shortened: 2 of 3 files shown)


djc (commented):
LGTM! How hard would it be to add some kind of test for this?

ishitatsuyuki:
Actually I'm thinking about an alternative way to implement this: instead of taking a closure, we return a `SuspendHandle` that when dropped, restores the progress bar and unlocks the mutex.

The reason is that the `tracing` crate's API exposes the writer like this:
```
pub trait MakeWriter<'a> {
    type Writer: Write;
    fn make_writer(&'a self) -> Self::Writer;
    (... optional methods)
}
```

If we used the closure-style API then we would need to wrap the `Write` trait and call `suspend()` for each of the `Write` function, which could be inefficient depending on how things are implemented. The `Drop` based approach would be easier, only requiring a wrapper struct to be returned as the `Writer`.

Or maybe we could expose both. BTW I found that it might be useful for us to pass on the return value from the closure so I'll add it in an upcoming commit.

As for automated testing, I don't have a good idea how it could be implemented yet. Any suggestions?

djc:
I think using a drop guard makes sense for this. Unless there's a good reason I'd prefer not to have both.

ishitatsuyuki:
I realized the closure one is better after all, as the drop guard will have a lifetime specifier, making various things harder to do. Especially when dealing with a static `Mutex<WeakProgressBar>`, it's not possible to obtain a `&'static mut ProgressBar` so the mutex guard cannot really be held.

Let's go with the closure one. It's kinda ugly to rely on the implementation detail, but with the current tracing-subscriber implementation it isn't less efficient than the drop guard implementation since the write is performed with a single `write_all` call.

<details>
<summary>Code I'm using in my prototype, for reference</summary>
```rust
struct SuspendProgressBarWriter<T: Write>(T);

impl<T: Write> SuspendProgressBarWriter<T> {
    fn suspend<F: FnOnce(&mut Self) -> R, R>(&mut self, f: F) -> R {
        let handle = PROGRESS_BAR.lock().unwrap().upgrade();
        if let Some(mut p) = handle {
            p.suspend(|| f(self))
        } else {
            f(self)
        }
    }
}

static PROGRESS_BAR: Lazy<Mutex<WeakProgressBar>> = Lazy::new(|| Mutex::new(WeakProgressBar::default()));

impl<T: Write> Write for SuspendProgressBarWriter<T> {
    #[inline]
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        self.suspend(|this| this.0.write(buf))
    }

    #[inline]
    fn write_vectored(&mut self, bufs: &[io::IoSlice<'_>]) -> io::Result<usize> {
        self.suspend(|this| this.0.write_vectored(bufs))
    }

    #[inline]
    fn flush(&mut self) -> io::Result<()> {
        self.suspend(|this| this.0.flush())
    }

    #[inline]
    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
        self.suspend(|this| this.0.write_all(buf))
    }

    #[inline]
    fn write_fmt(&mut self, fmt: std::fmt::Arguments<'_>) -> io::Result<()> {
        self.suspend(|this| this.0.write_fmt(fmt))
    }
}
```
</details>

ishitatsuyuki:
Patch updated:
- suspend takes `&self` instead of `&mut self` (I forgot that ProgressBar is Arc/Mutex wrapped)
- Fixed broken doc intra-links
- Add in a change to WeakProgressBar to make tracking progress bar globally easier

djc:
Thanks!

hasezoey:
@ishitatsuyuki could you update [your message](https://github.com/console-rs/indicatif/pull/333#issuecomment-995899268) to be properly formatted and/or is there another example (/documentation) of working with this function for the `log` crate and `indicatif`?

ishitatsuyuki:
I fixed the formatting. For env_logger, I think using [`env_logger::fmt::Target`](https://docs.rs/env_logger/latest/env_logger/fmt/enum.Target.html) with a `SuspendProgressBarWriter(std::io::stdout())` will work.

hasezoey:
> I fixed the formatting. For env_logger, I think using [`env_logger::fmt::Target`](https://docs.rs/env_logger/latest/env_logger/fmt/enum.Target.html) with a `SuspendProgressBarWriter(std::io::stdout())` will work.

thanks for the response and the prototype, though `env_logger` does not support outputting to something else than `stderr` and `stdout`, see https://github.com/env-logger-rs/env_logger/issues/208

for now i use a similar logger / crate: `simplelog`

---

for anyone that wants to use a "improved" version of the prototype from https://github.com/console-rs/indicatif/pull/333#issuecomment-995899268 :

```rs
use std::io::Write;

use indicatif::{
	ProgressBar,
	WeakProgressBar,
};
use parking_lot::RwLock;

// Utils file, may contain various small helper functions for main binary

#[derive(Debug)]
pub struct ProgressHandler<T: Write + Send>(T);

lazy_static::lazy_static! {
	/// Stores the reference to the current Progressbar, if there even is one
	/// uses "parking_lot::RwLock", because it is faster than the standart libraries and to allow many readers (read-only access to the variable)
	static ref CURRENT_PROGRESS_BAR: RwLock<Option<WeakProgressBar>> = RwLock::new(None);
}

impl<T: Write + Send> ProgressHandler<T> {
	/// Helper function to execute everything before running "inner_function"
	fn handle<F: FnOnce(&mut Self) -> R, R>(&mut self, inner_function: F) -> R {
		let handle = get_progress_bar();

		return match handle {
			Some(pb) => pb.suspend(|| return inner_function(self)),
			None => inner_function(self),
		};
	}

	/// Create a new instance of "Self"
	pub fn new(pipe: T) -> Self {
		return Self(pipe);
	}
}

impl<T: Write + Send> Write for ProgressHandler<T> {
	#[inline]
	fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
		return self.handle(|this| this.0.write(buf));
	}

	#[inline]
	fn flush(&mut self) -> std::io::Result<()> {
		return self.handle(|this| this.0.flush());
	}
}

/// Sets the reference to the current ProgressBar for the writer
pub fn set_progress_bar(pb: Option<WeakProgressBar>) {
	*CURRENT_PROGRESS_BAR.write() = pb;
}

/// Returns the Progressbar, if it exists, otherwise "None"
fn get_progress_bar() -> Option<ProgressBar> {
	return CURRENT_PROGRESS_BAR.read().as_ref()?.upgrade();
}
```

and then the `simplelog` integration:

```rs
use simplelog::*;
WriteLogger::init(
	LevelFilter::Debug,
	ConfigBuilder::new().set_thread_mode(ThreadLogMode::Both).build(),
	ProgressHandler::new(std::io::stdout()),
)
.expect("Failed to create Logger");
```

ishitatsuyuki:
In that case, it's also possible to use [`Builder::format`](https://docs.rs/env_logger/latest/env_logger/struct.Builder.html#method.format) which essentially allows control over the whole `writeln!` call. The drawback of this is that you need to can't use env_logger's default formatting and need to write your own.

wbogocki:
The workaround is problematic because if the PB is either finished or hidden then it also discards all log messages. This is impossible to get around because `.is_hidden()`, `.is_finished()` and `.println()` each lock and unlock the mutex individually so there's a race condition if we tried to do the checks like this:

```rust
if pb.is_hidden() || pb.is_finished() { // Race condition on the ||
    println!(...); // Race condition because the if check is not necessarily valid anymore
} else {
   pb.println(...);
}
```

`.suspend()` is likewise problematic because the PB can't be accessed inside the suspend closure because all of the methods try to lock the mutex which would've already been locked inside of the closure and result in a deadlock.

Perhaps it'd be better to make a `.println_always()` that ignores whether the PB is hidden or finished?

Personally, I'm tempted to either patch the crate locally or get the private mutex with some trickery, lock it, do my things, unlock it.

This is just my two cents, once 0.17 is released and `suspend()` works it's good enough for my use case :+1:

ishitatsuyuki:
> .suspend() is likewise problematic because the PB can't be accessed inside the suspend closure because all of the methods try to lock the mutex which would've already been locked inside of the closure and result in a deadlock.

Is this a problem for you? I thought that you won't need to access the progress bar inside `suspend`. It's supposed that you call `println!()` or whatever write operation to stdout.

wbogocki:
No, it's not a problem for me, you're good. I was unsure whether `println!()` and `pb.println()` would have the same effect, I'm guessing they would so I'll be good. I'm only slightly annoyed with how closed the internals are so you can't patch over them when needed.

PR #338, opened 2021-12-29
Don't add an extra line after the progress bar

sigmaSd:
Fix https://github.com/mitsuhiko/indicatif/issues/334

The math seems to checkout, but I can't be 100 % confidant

- I prefer not to change console because this current behavior could be relied on by crates
- I'm using write_str which is doc_hidden is that ok?


sigmaSd:
- I did not test on windows

djc:
Seems okay, thanks!

Issue #339, opened 2021-12-30
Style position and length with commas


dabreegster:
This is a fantastic crate, thanks so much for building and maintaining it!

I'm currently including `{pos}/{len}` in my template string, but for huge data, something like "33857009/42672348" isn't terribly readable to me. I'd like to style it as "33,857,009/42,672,348". I'm proposing:

1) Adding a `HumanCount` structure along the lines of `HumanBytes`
2) Adding `human_pos` and `human_len` as template keys, in https://github.com/mitsuhiko/indicatif/blob/0626585b006e9ba11df065c0d0d3df275d202257/src/style.rs#L270

Very open to better names than these. To actually do the formatting could be a bit tricky, since there are different styles based on locale: https://docs.microsoft.com/en-us/globalization/locale/number-formatting. 

Option 1) Just hardcode the use of US-style commas.
2) Use [num_format](https://docs.rs/num-format/latest/num_format/) and allow the user to explicitly override a `Locale`, possibly as a new `set_locale` method on `ProgressBar`
3) Use `num_format` and enable the `with-system-locale` feature
4) Use `num_format`, let people specify the locale in the string (like `{len:en}`), and have some kind of parsing to get to the `Locale` enum

Any preferences here? I'm happy to send a PR; the hard part is making these decisions. Thanks!

djc:
`HumanCount`, `human_len` and `human_pos sound good to me. Let punt on the localization question and stick to US-style formatting for now. PR welcome!

PR #340, opened 2021-12-30
Add HumanCount and template keys to print position and length with co…

dabreegster:
…mmas. #339

Motivation and trade-offs described in the issue. This was quite straightforward. Already using in https://github.com/dabreegster/rampfs/commit/7332180b1b5abea2e3741a20ae2684cbcc08a1d4, demo:
Screenshot from 2021-12-30 17-26-30

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -30,6 +30,10 @@ pub struct DecimalBytes(pub u64);
 #[derive(Debug)]
 pub struct BinaryBytes(pub u64);
 
+/// Formats counts for human readability using commas
+#[derive(Debug)]
+pub struct HumanCount(pub u64);
+
 impl fmt::Display for FormattedDuration {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut t = self.0.as_secs();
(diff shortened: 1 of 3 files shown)


src/format.rs
```diff
@@ -125,6 +129,23 @@ impl fmt::Display for BinaryBytes {
     }
 }
 
+impl fmt::Display for HumanCount {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use fmt::Write;
+
+        let num = self.0.to_string();
```
dabreegster, line 136: There's almost certainly a more efficient implementation. Let me know if I should work on this or benchmark it

djc: Hmm, not sure there's an obviously more efficient implementation.

I do think the backwards counting loop variable is a little confusing, how about changing the loop to just use `for (i, c) in num.chars().enumerate()` and rename your current `i` to `len` (immutable), then maybe add a binding for `let pos = len - i;` or something?

djc (commented):
This mostly looks good to me, one style nit and I'll be happy to merge it!

src/format.rs
dabreegster: Agreed, I think the new version is a bit simpler. Another idea is to iterate in reverse, but then that requires reversing the output string

PR #345, opened 2022-01-02
Add a space between numbers and their units

firasuke:
I think it's better to have some sort of separator (e.g. space) between numbers and their units.


firasuke:
@djc thanks for merging this, I got a question though, shouldn't:
```Rust
NumberPrefix::Standalone(number) => write!(f, "{:.0}B", number),
```
also be changed to:
```Rust
NumberPrefix::Standalone(number) => write!(f, "{:.0} B", number),
```
I saw no change though when changing `NumberPrefix::Standalone(number)`, perhaps because it's only changing the number and not the prefix? If so, why is there a `B` in there?

Thanks again!

firasuke:
Also it seems that some tests need to be modified after this changed.

djc:
Sorry, I clearly wasn't paying enough attention when I merged this. I reverted it in 20c569379a39a856324c535e419ea523da0bd902, can you do another take to fix the CI issues?

Issue #348, opened 2022-01-02
Why is there an additional space in the end?


firasuke:
It seems that regardless of the template of the progress bar, there's always an extra space " " in the end.

Is this behavior intended? If so, can it be changed?

Thanks!

djc:
I don't know what the original rationale is exactly, but it seems to be necessary: #342.

sigmaSd:
Im not sure if there is an alternative, without that space issue like the ones pointed above arises, maybe we can only add a space when the bar size is the terminal width size or just hide the cursor or just keep this behaviour

firasuke:
I think it will be better to make both of them optional, but it shouldn't default to add an extra space in the end, and maybe provide an option to hide the cursor as well.

sigmaSd:
Actually I think 342 is not solved with the space, if the user tries to use println after a bar that is less then the terminal width it will exibit the original issue behaviour

So the space hack is invalid its needs to be rwmoved

This cycles back to the original issue we made the progressbar not have a new line at the end, I think its simply the tradeoff without a newline the user will have 342 behaviour and he will simply need to be aware of it so he can use an extra println

Personaly I think the tradeoff is worth it, the extra new line always looked weird, and the fix from the user side if neededis simple

sigmaSd:
Also we can make the default situation better by making the default bar fill the whole terminal width instead of width - 1

sigmaSd:
So it should look something like this:
```patch
diff --git a/src/draw_target.rs b/src/draw_target.rs
index 1bb9c56..2803a0d 100644
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -396,9 +396,6 @@ impl ProgressDrawState {
             } else {
                 // Don't append a '\n' if this is the last line
                 term.write_str(line)?;
-                // Also append a ' ' to keep the original chars count
-                // This is important if the line was meant to fill the entire width
-                term.write_str(" ")?;
             }
         }
         Ok(())
diff --git a/src/style.rs b/src/style.rs
index f446e14..df047ee 100644
--- a/src/style.rs
+++ b/src/style.rs
@@ -390,7 +390,7 @@ impl<'a> WideElement<'a> {
         state: &ProgressState,
         buf: &mut String,
     ) -> String {
-        let left = state.width().saturating_sub(measure_text_width(&cur));
+        let left = state.width().saturating_sub(measure_text_width(&cur)) + 1;
         match self {
[...]

Issue #342, opened 2021-12-30
A newline character should be printed after completion


redzic:
This code used to print the progress bar, and then "Something" on the next line:

```rust
use std::thread;
use std::time::Duration;

use indicatif::ProgressBar;

fn main() {
    let pb = ProgressBar::new(1024);
    for _ in 0..1024 {
        pb.inc(1);
        thread::sleep(Duration::from_millis(5));
    }
    pb.finish_with_message("done");
    println!("Something");
}

```

However after #338 it prints the progress bar and the first character of what should be the next line ("S") on the same line, and then "omething" on the next line, like so:

```
██████████████████ 1024/1024S
omething
```

djc:
Thanks for testing! @sigmaSd, would you be able to look into this?

Separately, I wonder how we could improve our testing story to catch this kind of thing.

sigmaSd:
Yes , I think there is an io flush missing, hopefully I can look at it tomorrow

sigmaSd:
I read the code again, it does look to work as intended since I removed the new line at the end, but I guess the progress bar line still contains a place for one more character, maybe it will work if add an extra character something like `write_str(line + ' ')`

sigmaSd:
For testing, what about adding a custom draw target, something like:

```patch
diff --git a/src/draw_target.rs b/src/draw_target.rs
index 1bb9c56..d35030d 100644
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -136,11 +136,14 @@ impl ProgressDrawTarget {
             ProgressDrawTargetKind::Term { ref term, .. } => term.size().1 as usize,
             ProgressDrawTargetKind::Remote { ref state, .. } => state.read().unwrap().width(),
             ProgressDrawTargetKind::Hidden => 0,
+            #[cfg(test)]
+            ProgressDrawTargetKind::TestTerm { ref term, .. } => term.width(),
         }
     }
 
     /// Apply the given draw state (draws it).
     pub(crate) fn apply_draw_state(&mut self, draw_state: ProgressDrawState) -> io::Result<()> {
+        #[cfg(not(test))]
         let (term, last_line_count) = match self.kind {
             ProgressDrawTargetKind::Term {
                 ref term,
@@ -169,11 +172,22 @@ impl ProgressDrawTarget {
             // Hidden, finished, or no need to refresh yet
             _ => return Ok(()),
         };
+        #[cfg(test)]
+        let (term, last_line_count) = match self.kind {
[...]

PR #350, opened 2022-01-10
Remove the extra space, and make the wide element fill the whole term…

sigmaSd:
…inal width

fix https://github.com/console-rs/indicatif/issues/348
fix https://github.com/console-rs/indicatif/issues/342 again

Recap of the situation:
- https://github.com/console-rs/indicatif/pull/338 removed the extra line that indicatif always adds
- But now issue like  #342 happens
- Initially I thought just adding an extra space is a good idea to fill the width
- But that looks weird, so the new solution is to increase the wide elements size to fill the terminal width

If no wide elements is used, any printing after will happen on the same line as the progress bar, this might be not what users expects but I think its a desirable change, and reverting to the old behavior is a simple as adding println!() after the progress bar

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -396,9 +396,6 @@ impl ProgressDrawState {
             } else {
                 // Don't append a '\n' if this is the last line
                 term.write_str(line)?;
-                // Also append a ' ' to keep the original chars count
-                // This is important if the line was meant to fill the entire width
-                term.write_str(" ")?;
             }
         }
         Ok(())
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -396,6 +396,10 @@ impl ProgressDrawState {
             } else {
                 // Don't append a '\n' if this is the last line
                 term.write_str(line)?;
+                // Keep the cursor on the right terminal side
+                // So that next user writes/prints will happen on the next line
+                let line_width = console::measure_text_width(line);
+                term.move_cursor_right(usize::from(term.size().1) - line_width)?;
             }
         }
         Ok(())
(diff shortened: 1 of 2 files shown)


djc:
What happens if there is no wide element?

sigmaSd:
if there is no wide element,
- the bar wont fill the terminal width
- there is no ending new line
 
The next prints will be on the same line

djc:
> if there is no wide element,
> 
>     * the bar wont fill the terminal width
>     * there is no ending new line
> 
> The next prints will be on the same line 

So that doesn't sound like what we want. Maybe in absence of a wide element maybe we should have a sort of virtual wide element that fills the line up with spaces?

sigmaSd:
I personaly prefer to leave that to the user

What If I actually want to print on the same line, cool that just works

What if I want the old behaivour and print on the next line, simple add println!() and then  print what I actually want


Other ideas I thought about is println!() on Progressbar drop but that gets not easy when considering multi progressbar, thats why I adovocate the above for its simplicity

sigmaSd:
I guess the virtual space is also possible but it means that the cursor will always be locked to the right edge of the terminal, which might also look weird

djc:
Isn't that what it already looks like if we have a wide element today (or after this PR)?

djc:
The question is how this compares to the situation pre-#338. It seems to me that pre-#338, the cursor always ended up in the first column of the line after the progress bar. IMO changing that to have the cursor consistenly in the last column of the progress bar line seems the best approach. I don't see much virtue in having the cursor be directly after the bar if the bar is non-wide.

sigmaSd:
That's reasonable, I added a commit with your suggestion.

djc:
Thanks!

PR #356, opened 2022-01-26
Actually keep the cursor on the right terminal side

sigmaSd:
Test case:
```rust

fn main() {
    let pb = ProgressBar::new(500).with_style(ProgressStyle::default_bar().template("hello {bar}"));
    for _ in 0..1024 {
        pb.inc(1);
    }
    println!("Something");
}
```

https://github.com/console-rs/indicatif/pull/350/commits/1b13d6f695e8f8954d2c5203a24cb9854f4f896a didn't actually work as you can see with the test case, apparently moving the cursor to the right will never reach the last column, writing spaces instead actually works


djc:
Thanks for fixing it!

PR #361, opened 2022-02-02
Draw progress bars into draw states

djc:
This clarifies the internal structure and avoids reallocating some buffers.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -140,56 +140,31 @@ impl ProgressDrawTarget {
     }
 
     /// Apply the given draw state (draws it).
-    pub(crate) fn apply_draw_state(&mut self, draw_state: ProgressDrawState) -> io::Result<()> {
-        let (term, last_line_count) = match self.kind {
+    pub(crate) fn apply_draw_state(&mut self, mut draw_state: ProgressDrawState) -> io::Result<()> {
+        match self.kind {
             ProgressDrawTargetKind::Term {
                 ref term,
                 ref mut last_line_count,
                 leaky_bucket: None,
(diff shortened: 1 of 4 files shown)


src/draw_target.rs
```diff
@@ -388,7 +363,24 @@ impl ProgressDrawState {
         }
     }
 
-    pub fn draw_to_term(&self, term: &Term) -> io::Result<()> {
+    fn draw_to_term(&mut self, term: &Term, last_line_count: &mut usize) -> io::Result<()> {
```
djc, line 463: @chris-laplante I think this PR moves all the `Term` interaction into a single function which could easily take an `&impl TermLike`, making your trait PR easier (and avoiding boxing/trait objects). What do you think?

chris-laplante: It's a lot to digest but I think it looks good!

chris-laplante: The only thing I might suggest (which is not specific to this PR, but the extent of the changes has reminded me of it) is that I think some of these files are starting to grow pretty big and it's getting hard for me to remember where things are. For example, I feel like maybe we should move the `MultiProgress` out of `progress_bar.rs` into its own `multi_progress.rs`? And perhaps `MultiProgressState` should move out of `state.rs` into `multi_progress.rs`? Again, not specific to this PR so feel free to disregard :)

djc: That's fair, I'll see if I can improve on that. Right now the inherent impl for `MultiProgressState` is mostly concerned with drawing so that particular example of moving `MultiProgressState` might not work well, but I'll see if things can be improved.

Issue #358, opened 2022-02-02
0.17.0-rc.2 panic on win10 git-bash 


ubearline:
### demo
``` rust
extern crate indicatif;

use indicatif::{ProgressBar, ProgressStyle, MultiProgress};
use std::thread::sleep;
use std::time::Duration;

fn main() {
    let m = MultiProgress::new();
    let pb = m.insert(0, ProgressBar::new(128));
    let filename= "some_file";
    pb.set_style(
        ProgressStyle::default_bar()
            .template(
                ("{spinner:.green} [{elapsed_precise}] [{bar:50.green/cyan}]{percent:>3}% ["
                    .to_string()
                    + filename
                    + "] [{pos}/{len}]{eta} {msg}")
                    .as_str(),
            )
            .with_key("eta", |state| {
                let eta = state.eta().as_secs_f64();
                if eta > 0.0 {
                    format!(" [{:.1}s]", state.eta().as_secs_f64())
                } else {
                    "".to_string()
                }
            })
            .progress_chars(">>-"),
    );


    for i in 0..128{
        pb.set_position(i);
        sleep(Duration::from_millis(10))
    }
    pb.finish_with_message("done");
}

```
### win10 powershell normal
![图片](https://user-images.githubusercontent.com/32406183/152120579-beaed874-a231-4ef5-925d-30921508923f.png)

### win10 cmd normal
![图片](https://user-images.githubusercontent.com/32406183/152121140-d4725ee0-f728-404f-a749-5b7e0909d5cc.png)

### win10 git-bush panic (It was working in the last version 0.17.0-beta.1)
![图片](https://user-images.githubusercontent.com/32406183/152120627-2563e263-0cdd-4ed7-85fc-b53a9c4060eb.png)

### panic log
``` 
thread 'main' panicked at 'attempt to subtract with overflow', C:\Users\Dami\.cargo\registry\src\mirrors.sjtug.sjtu.edu.cn-7a04d2510079875b\indicatif-0.17.0-rc.2\src\draw_target.rs:402:44
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', C:\Users\Dami\.cargo\registry\src\mirrors.sjtug.sjtu.edu.cn-7a04d2510079875b\indicatif-0.17.0-rc.2\src\draw_target.rs:165:22
stack backtrace:
[...]

PR #364, opened 2022-02-05
Fix panic when calculating padding spaces

sigmaSd:
The user progress template can exceed the terminal width, so we use a
`saturating_sub` when calculating padding spaces to clamp it at 0 in
this case.

fix https://github.com/console-rs/indicatif/issues/358


djc (approved):
Thanks!

PR #371, opened 2022-02-10
Pass force_draw into drawable()

### LINKED ISSUE #368: regression: `finish_and_clear` is broken in rc.2 and current main branch
This code removes the bar in `0.16.2` but leaves bar visible in both `0.17.0-rc.2` and current main branch.

```
use indicatif::ProgressBar;
fn main() {
    let progress = ProgressBar::new(10);
    progress.inc(1);
    progress.finish_and_clear();
}
```

djc:
This is a regression from my work on drawing into draw states. Specifically, before 57e6ee282d5dd71f7846bd28e3f0369fd3f6d452 `force_draw` was passed in as part of the `ProgressDrawState`, but after that commit it was no longer handled correctly. Fixes #368.

@chris-laplante I think this is pretty much what you were suggesting, right? If you want to give it a quick look that would be great.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -156,7 +156,7 @@ impl ProgressDrawTarget {
     }
 
     /// Apply the given draw state (draws it).
-    pub(crate) fn drawable(&mut self) -> Option<Drawable<'_>> {
+    pub(crate) fn drawable(&mut self, force_draw: bool) -> Option<Drawable<'_>> {
         match &mut self.kind {
             ProgressDrawTargetKind::Term {
                 term,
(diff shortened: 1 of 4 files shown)


chris-laplante:
> This is a regression from my work on drawing into draw states. Specifically, before [57e6ee2](https://github.com/console-rs/indicatif/commit/57e6ee282d5dd71f7846bd28e3f0369fd3f6d452) `force_draw` was passed in as part of the `ProgressDrawState`, but after that commit it was no longer handled correctly. Fixes #368.
> 
> @chris-laplante I think this is pretty much what you were suggesting, right? If you want to give it a quick look that would be great.

Exactly what I was thinking! Thanks for taking care of it :)

djc:
Thanks for the quick review!

chris-laplante:
@djc: Question for you. I'm going to develop a (eventually comprehensive) test suite for indicatif based on the in-memory work I did. They will be a fair bit of code (though I plan on writing a macro or two to reduce boilerplate), so I'm thinking they should go in the top-level tests/ directory. What do you think?

djc:
For any tests that don't need access to private APIs, putting them in an integration test (that is, in the top-level tests directory) makes sense to me! Thanks for your efforts on improving test coverage.

Out of curiosity, what's your motivation/use case for indicatif?

djc:
Nice!

PR #378, opened 2022-02-16
InMemoryTerm: fix for write_line and add unit tests

chris-laplante:


src/in_memory.rs
```diff
@@ -83,12 +83,18 @@ impl TermLike for InMemoryTerm {
 
     fn write_line(&self, s: &str) -> std::io::Result<()> {
         let mut state = self.state.lock().unwrap();
+
+        // Don't try to handle writing lines with additional newlines embedded in them - it's not
+        // worth the extra code for something that indicatif doesn't even do. May revisit in future.
+        assert!(
```
djc, line 93: Maybe this should be a `debug_assert!()`?

chris-laplante: Good point, I'll change it

src/in_memory.rs
```diff
@@ -46,6 +46,10 @@ impl InMemoryTerm {
         rows.reverse();
         rows.join("\n")
     }
+
+    pub fn cursor_position(&self) -> (u16, u16) {
```
djc, line 50: I don't think this needs to be pub?

PR #379, opened 2022-02-17
Add render test suite

chris-laplante:


Cargo.toml
```diff
@@ -30,3 +30,8 @@ tokio = { version = "1", features = ["time", "rt"] }
 default = ["unicode-width", "console/unicode-width"]
 improved_unicode = ["unicode-segmentation", "unicode-width", "console/unicode-width"]
 in_memory = ["vt100"]
+
+[[test]]
+name = "render"
+path = "tests/render.rs"
+required-features = ["in_memory"]
```
djc, line 37: Instead of this, I'd prefer putting a `#![cfg(feature = "in_memory")]` at the top of `render.rs`.

djc:
Yeah... I think we should rip out the draw limiting in the `BarState` and instead rely on the `LeakyBucket` stuff in the draw target.

chris-laplante:
> Yeah... I think we should rip out the draw limiting in the `BarState` and instead rely on the `LeakyBucket` stuff in the draw target.

Feel free to repurpose this PR if you want to add commits to do that. It might make sense given to develop those fixes in tandem with the render test suite. I can help flesh out the test suite (as originally intended) but I'm not sure I can help much with `LeakyBucket` though.

djc:
Do you want to test? https://github.com/console-rs/indicatif/pull/380

djc:
I added back the test case in 3930cfc324439bb9dc8977f5380e263827b79544 after I reverted some of my draw limiting changes and merged the move into an integration test as 0c153b8bfadb28fd45ec76b284609d66e7ae8e24. Thanks!

PR #380, opened 2022-02-17
Remove draw limiting from the ProgressBar state

djc:
We now have rate limiting in the draw target, which is takes care
of most of the costs of drawing. Limiting earlier is not as
helpful in terms of performance, and reduces accuracy.

### DIFF
--- a/examples/fastbar.rs
+++ b/examples/fastbar.rs
@@ -2,11 +2,8 @@ use std::time::Instant;
 
 use indicatif::{HumanDuration, ProgressBar};
 
-fn many_units_of_easy_work(n: u64, label: &str, draw_delta: Option<u64>) {
+fn many_units_of_easy_work(n: u64, label: &str) {
     let pb = ProgressBar::new(n);
-    if let Some(v) = draw_delta {
-        pb.set_draw_delta(v);
-    }
 
     let mut sum = 0;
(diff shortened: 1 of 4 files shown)


chris-laplante:
> What about if we just always update?

I feel like there is still a hole in the design. `LeakyBucket` should (IMHO) only come into play when `pos` and/or `tick` change. Any other change to `ProgressStyle` (such as changing the message, prefix, style, e.g.) or hiding/showing the progress bar should not be rate limited since they are important visual changes. 

Granted, I have not encountered this issue, but I think they exist in the code.

chris-laplante:
However in the meantime, I think this change is good

djc:
> I feel like there is still a hole in the design. `LeakyBucket` should (IMHO) only come into play when `pos` and/or `tick` change. Any other change to `ProgressStyle` (such as changing the message, prefix, style, e.g.) or hiding/showing the progress bar should not be subject to rate limiting since they are important visual changes.

I think this is already how it works today. The condition for returning a `Some(Drawable::Term { .. })` for `ProgressDrawTargetKind::Term` is `force_draw || has_capacity` (where `has_capacity` is effectively the outcome of the `LeakyBucket` limiter -- and `force_draw` is what most things like changing the message/prefix/style are already using.

PR #381, opened 2022-02-17
Do not use boxcar windowing when displaying the final total of the speed

Yatekii:
This is especially relevant when speed varies a lot. Then confusing things like this https://github.com/probe-rs/probe-rs/issues/801 can happen. This is unfortunately only barely, if at all, visible in the examples as speed, whilst varying a little, is steady.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -240,11 +240,15 @@ impl ProgressState {
 
     /// The number of steps per second
     pub(crate) fn per_sec(&self) -> f64 {
-        let per_sec = 1.0 / self.est.seconds_per_step();
-        if per_sec.is_nan() {
-            0.0
+        if matches!(&self.status, Status::InProgress) {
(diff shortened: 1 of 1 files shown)


src/style.rs
```diff
@@ -313,18 +313,51 @@ impl ProgressStyle {
...
-                                .unwrap(),
-                            "binary_bytes_per_sec" => buf
-                                .write_fmt(format_args!(
-                                    "{}/s",
-                                    BinaryBytes(state.per_sec() as u64)
-                                ))
-                                .unwrap(),
+                            "per_sec" => {
+                                if matches!(&state.status, Status::InProgress) {
+                                    buf.write_fmt(format_args!("{:.4}/s", state.per_sec()))
```
djc, line 318: I'm open to this change, but this way seems unnecessarily repetitive. It looks like we can put the logic here into a separate `ProgressState` method?

djc (approved):

PR #395, opened 2022-03-14
ProgressBar: ensure ProgressBar::with_style doesn't overwrite msg/prefix

chris-laplante:


src/progress_bar.rs
```diff
@@ -55,8 +55,16 @@ impl ProgressBar {
...
 
     /// A convenience builder-like function for a progress bar with a given style
-    pub fn with_style(self, style: ProgressStyle) -> ProgressBar {
-        self.state().style = style;
+    pub fn with_style(self, mut style: ProgressStyle) -> ProgressBar {
+        {
+            let mut state = self.state();
+            // Retain current message and prefix - they are actually stored as part of the style,
+            // however the user probably doesn't know this.
+            style.message = state.style.message.clone();
```
djc, line 63: We can probably do this without cloning. Maybe `mem::swap()` and then overwrite?

Style nit: I'd prefer to just call `drop(state);` instead of having the nested scope.

djc (commented):
Thanks!

I've been meaning to ask, would you be interested in joining as a collaborator? I get the sense you're pretty busy, but your contributions have been great and it'd be good to have someone who can review my changes.

chris-laplante:
> Thanks!
> 
> I've been meaning to ask, would you be interested in joining as a collaborator? I get the sense you're pretty busy, but your contributions have been great and it'd be good to have someone who can review my changes.

Yes I'd love to be a contributor :). You are right though, I am definitely pretty busy lol. But I'll do my best to help review changes.

src/progress_bar.rs
chris-laplante: Will do!

chris-laplante:
> > Yes I'd love to be a contributor :). You are right though, I am definitely pretty busy lol. But I'll do my best to help review changes.
> 
> Great! You should have an invite.

Got it :). I will refrain from any direct pushes for the foreseeable future. I'd prefer to have a second set of eyes on any changes I make so I'll make more PRs.

djc (approved):

PR #396, opened 2022-03-16
Add ProgressBar.style().

andrewchambers:


src/progress_bar.rs
```diff
@@ -54,6 +54,11 @@ impl ProgressBar {
         }
     }
 
+    /// Get the a clone of current progress bar style.
+    pub fn current_style(self) -> ProgressStyle {
```
djc, line 58: Please change the name to `style()`, and return a reference. Then, the caller can clone it.

djc (commented):
Hey, thanks for submitting a PR. What's your use case here?

andrewchambers:
The use case is a function that takes a progress bar, wishes to perform an operation, then restore the progress bar to its old style.

src/progress_bar.rs
andrewchambers: Is this possible when it is behind a mutex?

andrewchambers:
Sorry I must have been a bit tired, fixed the sloppy comment.

djc (approved):

src/progress_bar.rs
djc: That's a fair point, it is not!

PR #402, opened 2022-03-17
Account for alignment when truncating

### LINKED ISSUE #400: Possible to truncate beginning of wide_msg?
I'm wanting to have wide_msg truncate the beginning of the string rather than the end of the string.

So basically if we use this sentence as an example, truncating to 40 characters:

```
an example, truncating to 40 characters:
```

As opposed to the existing behavior:

```
So basically if we use this sentence as 
```

I tried to find a way to determine what the current width of wide_msg is so I could manually do this, but I didn't see any obvious public exposure of it.

djc:
Fixes #400.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -319,6 +319,12 @@ impl ProgressStyle {
                         }
                     };
 
+                    if buf == "\x00" {
+                        // Don't expand for wide elements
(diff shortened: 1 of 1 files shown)


chris-laplante (approved):
LGTM!

PR #403, opened 2022-03-17
Fix `Estimator` panic

arxanas:
This is happening in some of my CI runs, like here: https://github.com/arxanas/git-branchless/runs/5581044721?check_suite_focus=true

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -336,7 +336,7 @@ impl Estimator {
     }
 
     fn record(&mut self, new: u64, now: Instant) {
-        let delta = new - self.prev.0;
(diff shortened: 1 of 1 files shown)


arxanas:
I've also been seeing panics in `AtomicPosition::allow`, but I also wasn't able to reproduce them.

djc:
I think #404 might fix that.

arxanas:
I filed https://github.com/console-rs/indicatif/issues/405 to track the `AtomicPosition::allow` panic. I think it's not addressed in #404; it's probably overflowing because of the `as u8` somehow, which is still present there.

PR #406, opened 2022-03-18
Fix addition overflow in `AtomicPosition`

### LINKED ISSUE #405: Panic in `AtomicPosition::allow`
This is the panic, from [this CI run](https://github.com/arxanas/git-branchless/runs/5596335762?check_suite_focus=true):

```
The application panicked (crashed).
Message:  attempt to add with overflow
Location: C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\indicatif-0.17.0-rc.9\src\state.rs:420
```

It corresponds to this line:

https://github.com/console-rs/indicatif/blob/ea1f7336be4320afca7dc4ad9353e64053e01343/src/state.rs#L420

Note that this is an addition which is overflowing, unlike the subtractions which are addressed in #403 and #404.

arxanas:
Closes #405.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -351,7 +351,7 @@ impl RateLimiter {
 
         // We add `new` to `capacity`, subtract one for returning `true` from here,
         // then make sure it does not exceed a maximum of `MAX_BURST`, then store it.
-        self.capacity = Ord::min(MAX_BURST, self.capacity + new as u8 - 1);
+        self.capacity = Ord::min(MAX_BURST as u128, (self.capacity as u128) + new - 1) as u8;
         // Store `prev` for the next iteration after subtracting the `remainder`.
         self.prev = now - Duration::from_nanos(remainder as u64);
         true
--- a/src/state.rs
+++ b/src/state.rs
@@ -418,6 +418,15 @@ pub(crate) struct AtomicPosition {
 }
 
 impl AtomicPosition {
+    pub(crate) fn new() -> Self {
+        Self {
+            pos: AtomicU64::new(0),
+            capacity: AtomicU8::new(MAX_BURST),
+            prev: AtomicU64::new(0),
+            start: Instant::now(),
+        }
+    }
+
     pub(crate) fn allow(&self, now: Instant) -> bool {
         if now < self.start {
             return false;
@@ -465,17 +474,6 @@ impl AtomicPosition {
 
     pub(crate) fn set(&self, pos: u64) {
         self.pos.store(pos, Ordering::Release);
-    }
-}
-
-impl Default for AtomicPosition {
-    fn default() -> Self {
-        Self {
-            pos: AtomicU64::new(0),
-            capacity: AtomicU8::new(MAX_BURST),
-            prev: AtomicU64::new(0),
-            start: Instant::now(),
-        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/state.rs
```diff
@@ -426,7 +426,7 @@ impl AtomicPosition {
         let (new, remainder) = ((diff / INTERVAL), (diff % INTERVAL));
         // We add `new` to `capacity`, subtract one for returning `true` from here,
         // then make sure it does not exceed a maximum of `MAX_BURST`.
-        capacity = Ord::min(MAX_BURST, capacity + new as u8 - 1);
+        capacity = Ord::min(MAX_BURST, u64::from(capacity).saturating_add(new) as u8 - 1);
```
djc, line 429: I think we do the saturating add in the `u8` domain, not in the `u64` domain.

arxanas: @djc is there a saturating constructor for `u8` that sets the value to `u8::MAX` if it wouldn't fit? I don't think we want to write `new as u8` here, since if `new` is particularly large, the most-significant bits will be cut off, and you could instead end up with a `new` value which is very small. That would cause the `capacity` to be assigned a lower than expected value. That being said, I don't understand the performance implications of this code (there don't seem to be any benches?), so I don't know which kinds of operations are expensive.

djc: I've used the fastbar example as a bench. I don't think integer operations will make a substantial impact, though, so we can just do what's necessary here, something like `Ord::min(capacity, u8::MAX as u128) as u8` or something?

djc: Also as long as you're fixing things here, please also audit the very similar logic in the `draw_target::RateLimiter`.

arxanas: The code seems to end up ugly no matter what configuration I try because all of the numeric types are different. The latest version upcasts, saturates, and then downcasts. In particular, I'm trying to avoid the use of `as`, since it could lead to weird bugs due to the fact that it truncates rather than saturates.

src/draw_target.rs
```diff
@@ -350,7 +351,15 @@ impl RateLimiter {
 
         // We add `new` to `capacity`, subtract one for returning `true` from here,
         // then make sure it does not exceed a maximum of `MAX_BURST`, then store it.
-        self.capacity = Ord::min(MAX_BURST, self.capacity + new as u8 - 1);
+        self.capacity = Ord::min(
```
djc, line 354: This is indeed pretty ugly, and IMO we can simplify this. Avoiding `as` conversions makes sense for fallible conversions but not for infallible ones. We can also define `MAX_BURST` as `u128` if that makes life easier. I don't like the use of `clamp()` with 0 as the floor since we're operating in the unsigned domain. That would get us this:

```rust
// Invariant: `new` must be > 1, otherwise we would have returned above 
self.capacity = Ord::min(self.capacity as u128 + new - 1, MAX_BURST) as u8;
```

arxanas:
Sorry for the delay. Resolved by doing arithmetic in `u128`.

FWIW, I would advise avoiding the use of `as` altogether, for the reasons at https://gist.github.com/SimonSapin/550ada9ebedcd21c41cf4e95891e03d7#not-yet-deprecating-the-as-keyword

djc (approved):

djc:
That discussion is pretty old at this point, though. Here's a more recent one:

https://internals.rust-lang.org/t/lets-deprecate-as-for-lossy-numeric-casts/16283/45

Note that this talks about deprecating `as` for lossy casts, which is not what we're doing here.

arxanas:
@djc Thanks for merging. Please let me know when there's another release/pre-release so that I can publish a version of my package which depends on `indicatif` to crates.io.

djc:
I've published rc10.

Issue #397, opened 2022-03-16
Strange behavior of spinners and enable_steady_tick?


chris-laplante:
Try running this code:

```
use std::thread;
use std::time::Duration;

use indicatif::{ProgressBar, ProgressFinish};

fn main() {
    let mut spinner: Option<ProgressBar> = None;

    for _ in 0..100 {
        let new_spinner = ProgressBar::new_spinner()
            .with_finish(ProgressFinish::AndLeave)
            .with_message("doing stuff");
        new_spinner.enable_steady_tick(Duration::from_millis(10));
        thread::sleep(Duration::from_millis(500));
        spinner.replace(new_spinner).map(|t| t.finish());
    }
}
```

Two things happen that I can't explain. 

1) The first spinner is not actually finished (red arrow)
2) Occasionally, some of the other spinners also don't finish. (blue arrow)

(1) happens even if you change the loop to only go from 0..2.

![image](https://user-images.githubusercontent.com/40474653/158625277-9ce98f47-c92a-456a-9749-ce2c235f6af8.png)

I question whether there is some kind of race condition somewhere in Ticker.

djc:
Hmm, I think your example code is not doing what you expect it to do. If you add some debug stuff like this:

```
diff --git a/src/state.rs b/src/state.rs
index 7e8a30f..265af6f 100644
--- a/src/state.rs
+++ b/src/state.rs
@@ -32,6 +32,7 @@ impl BarState {
     /// in the [`ProgressStyle`].
     pub(crate) fn finish_using_style(&mut self, now: Instant, finish: ProgressFinish) {
         self.state.status = Status::DoneVisible;
+        println!("\n\nfinish {}\n\n", self.style.message);
         match finish {
             ProgressFinish::AndLeave => self.state.pos.set(self.state.len),
             ProgressFinish::WithMessage(msg) => {
@@ -297,6 +298,7 @@ impl Ticker {
         thread::sleep(self.interval);
         while let Some(arc) = self.weak.upgrade() {
             let mut state = arc.lock().unwrap();
+            println!("ticker locked: {}", state.style.message);
             let interval = match state.ticker {
                 Some((interval, _)) if !state.state.is_finished() => interval,
                 _ => return,
```

And change the message like `.with_message(format!("doing stuff {}", i))` you can see that after the first iteration of the loop there are always two spinners alive at the same time, so I think they could be writing to the terminal in more or less arbitrary order.

[...]

PR #407, opened 2022-03-19
Fix Ticker issue #397

chris-laplante:
I think there are two parts to #397:

1. In addition to dropping the `MutexGuard<BarState>`, we also need to drop the `Arc<Mutex<BarState>>`, otherwise the reference count will never drop to 0 while the thread sleeps. The main thread will exit, but the destructor (on the `Ticker` thread's stack) will not run.
2. Even with (1) in place (which seems to fix the issue), I think to be safe we also need to `join` the `JoinHandle`. Trouble is I am not yet sure where to put the `join`. It can't go in the `Drop` impl for `BarState` since then the thread would probably try to join itself. Perhaps we need to move the ticker state into an `Arc<Mutex<Option<TickerState>>>` in `ProgressBar` or something. Then we could rewrite `Ticker::run` to use, e.g. [`CondVar::wait_timeout`](https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html#method.wait_timeout) instead of `thread::sleep(self.interval)`.

So more experimenting is necessary.

### DIFF
--- a/src/in_memory.rs
+++ b/src/in_memory.rs
@@ -41,6 +41,7 @@ impl InMemoryTerm {
             .into_iter()
             .rev()
             .skip_while(|line| line.is_empty())
+            .map(|line| line.trim_end().to_string())
             .collect();
 
         // Un-reverse the rows and join them up with newlines
--- a/src/state.rs
+++ b/src/state.rs
@@ -309,6 +309,7 @@ impl Ticker {
             self.interval = interval;
             state.draw(false, Instant::now()).ok();
             drop(state); // Don't forget to drop the lock before sleeping
+            drop(arc); // Also need to drop Arc otherwise BarState won't be dropped
             thread::sleep(self.interval);
         }
     }
--- a/tests/render.rs
+++ b/tests/render.rs
@@ -4,6 +4,7 @@ use indicatif::{
     InMemoryTerm, MultiProgress, ProgressBar, ProgressDrawTarget, ProgressFinish, ProgressStyle,
     TermLike,
 };
+use std::time::Duration;
 
 #[test]
 fn basic_progress_bar() {
@@ -240,5 +241,30 @@ And so is this
 
 Another line printed"#
             .trim()
+    );
+}
+
+#[test]
+fn ticker_drop() {
+    let in_mem = InMemoryTerm::new(10, 80);
+    let mp =
+        MultiProgress::with_draw_target(ProgressDrawTarget::term_like(Box::new(in_mem.clone())));
+
+    let mut spinner: Option<ProgressBar> = None;
+
+    for i in 0..5 {
+        let new_spinner = mp.add(
+            ProgressBar::new_spinner()
+                .with_finish(ProgressFinish::AndLeave)
+                .with_message(format!("doing stuff {}", i)),
+        );
+        new_spinner.enable_steady_tick(Duration::from_millis(50));
+        spinner.replace(new_spinner);
+    }
+
+    drop(spinner);
+    assert_eq!(
+        in_mem.contents(),
+        "  doing stuff 0\n  doing stuff 1\n  doing stuff 2\n  doing stuff 3\n  doing stuff 4"
     );
 }


djc:
Why do we need to join the thread to be safe? I would argue that, once the `Arc` has been dropped, the `Weak` can no longer be upgraded and the ticker thread will just stop.

Dropping the `Arc` is a good call and I think it's worth merging that.

chris-laplante:
> Why do we need to join the thread to be safe? I would argue that, once the `Arc` has been dropped, the `Weak` can no longer be upgraded and the ticker thread will just stop.

I think there's a race condition if `main()` returns/panics while the `Ticker` loop is holding onto the `Arc` (i.e. before it drops it and then sleeps).

djc:
> I think there's a race condition if `main()` returns/panics while the `Ticker` loop is holding onto the `Arc` (i.e. before it drops it and then sleeps).

Have you seen this happen? I'm still confused, can you spell out the order of events for me and why something would be bad?

chris-laplante:
> Have you seen this happen? I'm still confused, can you spell out the order of events for me and why something would be bad?

I have not seen it happen, no. The order of events would be:

Granted, it's a small window of opportunity. In practice I'd expect 99% of the time that the `Ticker` thread is sleeping with the `Arc` dropped. But I also don't see any reason why we _shouldn't_ join the thread.

chris-laplante:
Also, to be clear - I certainly may be mistaken :). I'm not super experienced with multi-threaded Rust yet (or Rust in general for that matter). So please take this with a grain of salt.

djc:
I'm also not super experienced with joining threads -- the `Arc<Mutex<Option<TickerState>>>` sounds like a lot of complexity for what seems like an edge case (in other words, complexity might be a reason not to join the thread?). Maybe add a `Drop` impl on `ProgressBar` that will block on locking the `Mutex` and sets the `BarState::ticker` to `None`?

djc:
(Also I still think we should merge this without blocking on that other stuff.)

chris-laplante:
> I'm also not super experienced with joining threads -- the `Arc<Mutex<Option<TickerState>>>` sounds like a lot of complexity for what seems like an edge case (in other words, complexity might be a reason not to join the thread?). Maybe add a `Drop` impl on `ProgressBar` that will block on locking the `Mutex` and sets the `BarState::ticker` to `None`?

I'm also fine with merging this for now. And yeah the complexity would kind of suck :/. Since `ProgressBar` can be cloned, I don't think we can add a drop implementation on it directly.

I'd also be fine with holding off on making any additional changes until (if) the hypothetical issue is uncovered in the wild.

djc (approved):

djc:
I suppose we could also add yet another layer of indirection, `ProgressBarInner`, which holds a `Mutex<BarState>`, with `ProgressBar` holding the `ProgressBarInner`? That way we could add a `Drop` on `ProgressBarInner`...

djc:
If `ProgressBarInner` holds the `Arc` we're back at the current state, where `ProgressBar` holds the `Arc`. I'm not entirely sure we actually have a problem if `ProgressBar` can be cloned? We can still have a `Drop` impl for it and just make that impl idempotent? Like, it will do work depending on the `Arc`'s strong refcount or something?

chris-laplante:
> If ProgressBarInner holds the Arc we're back at the current state, where ProgressBar holds the Arc.

The difference would be that `Ticker` would never hold an `Arc` to `ProgressBarInner` so it cannot prevent is destruction.

> We can still have a Drop impl for it and just make that impl idempotent? Like, it will do work depending on the Arc's strong refcount or something?

Maybe just this?

```
impl Drop for ProgressBar {
    fn drop(&mut self) {
        let _ = self.state();
    }
}
```

chris-laplante:
I think checking the strong refcount in `drop` is off the table, because of [this note](https://doc.rust-lang.org/std/sync/struct.Arc.html#safety-3):

> Another thread can change the strong count at any time, including potentially between calling this method and acting on the result.

chris-laplante:
I've been able to reproduce the race condition by using `Barrier`s to force it to happen. Please give this fork a try: https://github.com/chris-laplante/indicatif/tree/cpl/fun_with_barriers  and try running: `cargo run --example=long-spinner`

By default I have `Barrier` <span>#</span>1 and <span>#</span>2 uncommented. If you comment out barrier <span>#</span>2 and uncomment <span>#</span>3, you see correct behavior (since executing proceeds after the `Arc` is dropped). You can also comment out all the `Barrier`s and you should get correct behavior (unless of course, the race happens organically :wink:).

djc:
Awesome work making it testable! Maybe file it as a separate issue? I'm going to try to not think about it too much more today until after work. 😄

chris-laplante:
Thanks :). And sounds good, and will do! 

I'm going to investigate the possibility of adding some automated tests for `Ticker`, using similar techniques and perhaps mocking with https://docs.rs/mockall/latest/mockall/ or something. For example, it would be cool to be able to single-step the `Ticker` `n` times and confirm that the output is what we would expect.

But that will also have to wait until I'm done with some actual work today haha.

PR #408, opened 2022-03-20
Fix add with overflow panic

rgreinho:
While working on adding a new example to showcase the use of indicatif
with async multi-progress bars, I faced a random panic bug:

`thread 'main' panicked at 'attempt to add with overflow',
       src/state.rs:424:40`

As a result this patch provides a new example and a fix for a bug it
exposed.

PR #410, opened 2022-03-20
Update tokio example

rgreinho:
Updates the tokio with an example where a main progress bar monitors child
progress bars started asynchronously.

This examples also remove the blocking call in the async function, an uses the
`#[tokio::main]` macro instead of initializing the runtime manually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ exclude = ["screenshots/*"]
 
 [dependencies]
 console = { version = "0.15", default-features = false, features = ["ansi-parsing"] }
+futures = "0.3.21"
 number_prefix = "0.4"
 rayon = { version = "1.1", optional = true }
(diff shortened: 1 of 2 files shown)


rgreinho:
This PR is a follow up of #408.

djc:
I'm sorry, I should maybe have looked at this code more closely before suggesting you merge this into an existing example. I don't think this change makes sense as it doesn't seem to cover particularly new ground -- to me, the integration of a `MultiProgress` bar with async/await code seems fairly straightforward.

(This PR also makes other undesirable changes, like pulling a futures dependency into the library dependencies even though it's only used for an example -- that should at most be a dev-dependency, and it looks like we only need futures-util here.)

I appreciate the effort to contribute here, but I think this will not help -- people are more likely to need a single `ProgressBar` example in async code, and if they want to see how `MultiProgress` works they can look at other examples that show how that works.

PR #415, opened 2022-03-21
Store length as an Option

### LINKED ISSUE #409: Spinners wrong pos/human_pos/per_sec after .finish()
Spinners display a very large (incorrect) number when `.finish()`ed if `{pos}`/`{human_pos}`/`{per_sec}` are included in the style template.

From the source code, I see that when calling `.finish()` the `pos` is set to the `len` of the `ProgressState`. When creating a `new_spinner` though, this `len` is set to `!0`, which I have never seen in Rust, but calling `.finish()` displays a very large number that looks like an unsigned underflow or an uninitialized integer.

An easy way to get around this is to use `.abandon()` which doesn't override `pos` with `len`.

djc:
Previously we used a sentinel value of u64::MAX, which made it
harder to see what was going on or handle this case right.

Fixes #409.

### DIFF
--- a/examples/multi-tree-ext.rs
+++ b/examples/multi-tree-ext.rs
@@ -193,7 +193,7 @@ pub fn main() {
         ELEMENTS
             .iter()
             .map(|e| match e {
-                Elem::AddItem(item) => item.progress_bar.length(),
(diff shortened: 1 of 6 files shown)


chris-laplante (approved):
Looks good here

PR #417, opened 2022-03-22
rework Ticker (fix #416)

### LINKED ISSUE #416: Ticker thread might block final draw on progress bar
The order of events would be:

From https://github.com/console-rs/indicatif/pull/407#issuecomment-1074392506:

> Please give this fork a try: https://github.com/chris-laplante/indicatif/tree/cpl/fun_with_barriers and try running: cargo run --example=long-spinner
> 
> By default I have Barrier #1 and #2 uncommented. If you comment out barrier #2 and uncomment #3, you see correct behavior (since executing proceeds after the Arc is dropped). You can also comment out all the Barriers and you should get correct behavior (unless of course, the race happens organically 😉).

chris-laplante:
Here's a first draft of how I think `Ticker` should work. 

So far the only testing I've done is to confirm that the long-spinner example works OK. I have not yet tested the modified version of it (which uses the `Barrier`s).

Fixes #416.

TODO:
- [ ] Write tests

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -6,9 +6,9 @@ use std::{fmt, mem};
 use std::{io, thread};
 
 #[cfg(test)]
-use std::sync::atomic::AtomicBool;
+use once_cell::sync::Lazy;
 #[cfg(test)]
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::draw_target::ProgressDrawTarget;
 use crate::state::{AtomicPosition, BarState, ProgressFinish, Reset};
@@ -143,7 +143,7 @@ impl ProgressBar {
             return;
         }
 
-        self.stop_and_replace_ticker(Some(Ticker::new(interval, &self.state)));
+        self.stop_and_replace_ticker(Some(interval));
     }
 
     /// Undoes [`ProgressBar::enable_steady_tick()`]
@@ -151,13 +151,13 @@ impl ProgressBar {
         self.stop_and_replace_ticker(None);
     }
 
-    fn stop_and_replace_ticker(&self, new_ticker: Option<Ticker>) {
+    fn stop_and_replace_ticker(&self, interval: Option<Duration>) {
         let mut ticker_state = self.ticker.lock().unwrap();
         if let Some(ticker) = ticker_state.take() {
             ticker.stop();
         }
 
-        *ticker_state = new_ticker;
+        *ticker_state = interval.map(|interval| Ticker::new(interval, &self.state));
     }
 
     /// Manually ticks the spinner or progress bar
@@ -537,8 +537,7 @@ impl WeakProgressBar {
 }
 
 pub(crate) struct Ticker {
-    stop_received: Arc<Mutex<bool>>,
-    cond_var: Arc<Condvar>,
+    stopping: Arc<(Mutex<bool>, Condvar)>,
     join_handle: Option<thread::JoinHandle<()>>,
 }
 
@@ -553,79 +552,75 @@ impl Drop for Ticker {
 static TICKER_RUNNING: AtomicBool = AtomicBool::new(false);
 
 impl Ticker {
-    pub(crate) fn new(interval: Duration, bar_state: &Arc<Mutex<BarState>>) -> Ticker {
+    pub(crate) fn new(interval: Duration, bar_state: &Arc<Mutex<BarState>>) -> Self {
         debug_assert!(!interval.is_zero());
 
+        // A `Mutex<bool>` is used as a flag to indicate whether the ticker was requested to stop.
+        // The `Condvar` is used a notification mechanism: when the ticker is dropped, we notify
+        // the thread and interrupt the ticker wait.
         #[allow(clippy::mutex_atomic)]
-        let cancel_pending = Arc::new(Mutex::new(false));
-        let cond_var = Arc::new(Condvar::new());
-
+        let stopping = Arc::new((Mutex::new(false), Condvar::new()));
         let control = TickerControl {
-            stop_received: cancel_pending.clone(),
-            cond_var: cond_var.clone(),
-            weak_state: Arc::downgrade(bar_state),
+            stopping: stopping.clone(),
+            state: Arc::downgrade(bar_state),
         };
 
-        let join_handle = thread::spawn(move || {
-            #[cfg(test)]
-            TICKER_RUNNING.store(true, Ordering::SeqCst);
-
-            while let Some(arc) = control.weak_state.upgrade() {
-                let mut state = arc.lock().unwrap();
-                if state.state.is_finished() {
-                    break;
-                }
-
-                if state.state.tick != 0 {
-                    state.state.tick = state.state.tick.saturating_add(1);
-                }
-
-                state.draw(false, Instant::now()).ok();
-
-                drop(state); // Don't forget to drop the lock before sleeping
-                drop(arc); // Also need to drop Arc otherwise BarState won't be dropped
-
-                // Wait for `interval` but return early if we are notified to stop
-                let result = control
-                    .cond_var
-                    .wait_timeout_while(
-                        control.stop_received.lock().unwrap(),
-                        interval,
-                        |stopped| !*stopped,
-                    )
-                    .unwrap();
-
-                // If the wait didn't time out, it means we were notified to stop
-                if !result.1.timed_out() {
-                    break;
-                }
-            }
-
-            #[cfg(test)]
-            TICKER_RUNNING.store(false, Ordering::SeqCst);
-        });
-
-        Ticker {
-            stop_received: cancel_pending,
-            cond_var,
+        let join_handle = thread::spawn(move || control.run(interval));
+        Self {
+            stopping,
             join_handle: Some(join_handle),
         }
     }
 
     pub(crate) fn stop(&self) {
-        *self.stop_received.lock().unwrap() = true;
-        self.cond_var.notify_one();
+        *self.stopping.0.lock().unwrap() = true;
+        self.stopping.1.notify_one();
     }
 }
 
 struct TickerControl {
-    stop_received: Arc<Mutex<bool>>,
-    cond_var: Arc<Condvar>,
-    weak_state: Weak<Mutex<BarState>>,
+    stopping: Arc<(Mutex<bool>, Condvar)>,
+    state: Weak<Mutex<BarState>>,
 }
 
-#[cfg(test)]
-use once_cell::sync::Lazy;
+impl TickerControl {
+    fn run(&self, interval: Duration) {
+        #[cfg(test)]
+        TICKER_RUNNING.store(true, Ordering::SeqCst);
+
+        while let Some(arc) = self.state.upgrade() {
+            let mut state = arc.lock().unwrap();
+            if state.state.is_finished() {
+                break;
+            }
+
+            if state.state.tick != 0 {
+                state.state.tick = state.state.tick.saturating_add(1);
+            }
+
+            state.draw(false, Instant::now()).ok();
+
+            drop(state); // Don't forget to drop the lock before sleeping
+            drop(arc); // Also need to drop Arc otherwise BarState won't be dropped
+
+            // Wait for `interval` but return early if we are notified to stop
+            let (_, result) = self
+                .stopping.1
+                .wait_timeout_while(self.stopping.0.lock().unwrap(), interval, |stopped| {
+                    !*stopped
+                })
+                .unwrap();
+
+            // If the wait didn't time out, it means we were notified to stop
+            if !result.timed_out() {
+                break;
+            }
+        }
+
+        #[cfg(test)]
+        TICKER_RUNNING.store(false, Ordering::SeqCst);
+    }
+}
 
 // Tests using the global TICKER_RUNNING flag need to be serialized
 #[cfg(test)]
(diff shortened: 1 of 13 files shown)


djc:
If you're moving code around that should be in a separate commit from any other changes you make, otherwise reviewing this becomes very hard. I'm not convinced that the `Ticker` is large enough to warrant a separate module, so I think we should probably hold off on that refactoring until the end. If you start with adding the test I will probably try to relentlessly simplify while still passing the test, as I think this looks like way too much stuff...

chris-laplante:
> If you're moving code around that should be in a separate commit from any other changes you make, otherwise reviewing this becomes very hard. I'm not convinced that the `Ticker` is large enough to warrant a separate module, so I think we should probably hold off on that refactoring until the end. If you start with adding the test I will probably try to relentlessly simplify while still passing the test, as I think this looks like way too much stuff...

Understood - I've moved the code back.

chris-laplante:
Also just to add some context for the changes - instead of adding a `ProgressBarInner`, I realized it was easier to just pull the `Ticker` up a level from `BarState` into `ProgressBar`.

djc:
I feel like this is over-indexing on "we must join the thread" instead of solving the original problem of "we should make sure to do run a finish draw even if the ticker happens to hold the lock while the parent progress bar is finishing", resulting in a lot of stuff -- doubling the amount of code relating to the ticker and pulling in the additional `CondVar` type. I think the other idea of having a `Drop` impl block on the `Ticker` being done would be simpler and just as effective?

chris-laplante:
> I feel like this is over-indexing on "we must join the thread" instead of solving the original problem of "we should make sure to do run a finish draw even if the ticker happens to hold the lock while the parent progress bar is finishing", resulting in a lot of stuff -- doubling the amount of code relating to the ticker and pulling in the additional `CondVar` type. I think the other idea of having a `Drop` impl block on the `Ticker` being done would be simpler and just as effective?

IMHO a condition variable is the classic way to approach a problem like this (at least in C++), but if you can think of a different way I'd be happy to consider.

djc:
Yeah, so the way I have in mind is that we move the `Drop` impl from `BarState` to another newtype that wraps the `Mutex<BarState>`, and finishes the bar only once it can lock the `Mutex`.

chris-laplante:
> Yeah, so the way I have in mind is that we move the `Drop` impl from `BarState` to another newtype that wraps the `Mutex<BarState>`, and finishes the bar only once it can lock the `Mutex`.

Sure, as long as we can guarantee that the `drop` doesn't happen inside the `Ticker` thread, otherwise (unless we `join` the thread somewhere) the `drop` might not happen.

chris-laplante:
I can't think of a way to write a test for this without ensuring the thread is joined. 

@djc were you going to try to implement your proposal, or did you want me to take a stab at it?

djc:
As for the test, can we include the barriers in a test somehow?

chris-laplante:
> As for the test, can we include the barriers in a test somehow? Or build a render test out of the long spinner example?

Sure, I'll give it a shot!

chris-laplante:
I don't think there is a good way to write a test for this :/. The `Barrier` approach is helpful to force the race to manifest, but I don't think there is any way to use them to assert that a race _doesn't_ exist. At the very least, adding `Barrier`s only in `#cfg[test]` will add synchronization that then isn't present in the release build.

I think the best way forward would be for you to give your approach a shot. The "test" may be for me to try to insert `Barrier`s into it to break it like I did with the current `Ticker` code.

chris-laplante:
Actually, I may have just thought of a way... Will open a new PR with the test if I can get it to work.

chris-laplante:
It's been a while but I'm back at this... I can't figure out a way to test this particular implementation without using a nightly feature ([JoinHandle::is_finished](https://doc.rust-lang.org/nightly/std/thread/struct.JoinHandle.html#method.is_finished)), but I am still pretty certain that this is the correct approach.

djc:
I've been working on a bunch of other stuff (notably bringing chrono back to life), but am still thinking about this. I still feel like the requirement to join threads is to strong. As I understand it, the goal is that a non-ticker thread has the final draw, and it feels like we should be able to achieve that without having to join the thread. In fact, I think another goal here is that the non-ticker thread's `drop` of the `ProgressBar` should not wait for the ticker thread to be done with sleeping, since the sleep interval can be arbitrarily long.

chris-laplante:
> In fact, I think another goal here is that the non-ticker thread's `drop` of the `ProgressBar` should not wait for the ticker thread to be done with sleeping, since the sleep interval can be arbitrarily long.

FWIW this `CondVar` based approach does accomplish this, since the wait will be interrupted :)

djc:
I pushed a work-in-progress approach that includes the barrier-based test to the `ticker-termination-test`, it involves adding another `Mutex`. I think that could work, but it's not quite there yet -- the issue is in figuring out exactly how the progress bar state mutex and the ticker mutex should overlap.

Can you articulate what a `CondVar` solves here that a `Mutex` cannot?

chris-laplante:
> Can you articulate what a `CondVar` solves here that a `Mutex` cannot?

The `CondVar` lets us say "wait until timeout has elapsed OR we are signaled to stop". Condition variable is its own unique synchronization primitive so it's hard to compare to a plain mutex.

chris-laplante:
Also, I'm not convinced that joining the thread is too strong of a guarantee though I concede it may be. But at the end of the day, I'd argue it is much easier to reason about than if you try to build something with a bunch of mutexes. And it leads to simpler code.

djc:
(Please ping me explicitly when you think this is ready for review.)

chris-laplante:
@djc sorry, yes I think this is ready. I've been testing it locally and no problems so far (in fact I was running this code when I discovered #424)

chris-laplante:
@djc - any further thoughts on this? I'd love to get it merged in soon. It has been working for me so far in all my testing.

src/progress_bar.rs
```diff
@@ -506,10 +531,106 @@ impl WeakProgressBar {
...
+}
+
+struct TickerControl {
+    stop_received: Arc<Mutex<bool>>,
+    cond_var: Arc<Condvar>,
+    weak_state: Weak<Mutex<BarState>>,
+}
+
+#[cfg(test)]
+use once_cell::sync::Lazy;
```
djc, line 628: Let's move the import up with the other imports?

src/progress_bar.rs
```diff
@@ -506,10 +531,106 @@ impl WeakProgressBar {
...
+    }
+
+    pub(crate) fn stop(&self) {
+        *self.stop_received.lock().unwrap() = true;
+        self.cond_var.notify_one();
+    }
+}
+
+struct TickerControl {
+    stop_received: Arc<Mutex<bool>>,
```
djc, line 622: Naming nits: I think I prefer `stopping` over `stop_received` and `state` over `weak_state` (the "weak" part is obvious from the type). `cond_var` is completely redundant with the type, can you think of some name that describes the role?

src/progress_bar.rs
```diff
@@ -506,10 +531,106 @@ impl WeakProgressBar {
...
+        let cancel_pending = Arc::new(Mutex::new(false));
+        let cond_var = Arc::new(Condvar::new());
+
+        let control = TickerControl {
+            stop_received: cancel_pending.clone(),
+            cond_var: cond_var.clone(),
+            weak_state: Arc::downgrade(bar_state),
+        };
+
+        let join_handle = thread::spawn(move || {
```
djc, line 569: Is there any particular reason this needs to be a closure? I dislike the level of nesting we get into here and it'd be nice to make it explicit what state is getting passed into the closure.

src/progress_bar.rs
```diff
@@ -506,10 +531,106 @@ impl WeakProgressBar {
...
+        self.stop();
+        self.join_handle.take().map(|handle| handle.join());
     }
 }
 
+#[cfg(test)]
+static TICKER_RUNNING: AtomicBool = AtomicBool::new(false);
+
+impl Ticker {
+    pub(crate) fn new(interval: Duration, bar_state: &Arc<Mutex<BarState>>) -> Ticker {
```
djc, line 556: Style nit: please use `Self` instead of `Ticker` here and in the return value.

src/progress_bar.rs
```diff
@@ -132,19 +139,35 @@ impl ProgressBar {
...
+        self.stop_and_replace_ticker(Some(Ticker::new(interval, &self.state)));
     }
 
     /// Undoes [`ProgressBar::enable_steady_tick()`]
     pub fn disable_steady_tick(&self) {
-        self.state().ticker = None;
+        self.stop_and_replace_ticker(None);
+    }
+
+    fn stop_and_replace_ticker(&self, new_ticker: Option<Ticker>) {
```
djc, line 154: I think this should take an `interval: Option<Duration>` instead of `Option<Ticker>`, which simplifies the caller. I would also consider defining a `TickerState` which contains the `Mutex<Option<Ticker>>` and see if that cleans things up.

djc (commented):
Here's an initial round of feedback. I expect I'll have some more things after this stuff has been addressed.

chris-laplante:
Suggestions applied. Sorry, I didn't do it by squashing since it was too annoying (because the structs moved).

src/progress_bar.rs
```diff
@@ -1,12 +1,17 @@
 use std::borrow::Cow;
-use std::io;
-use std::sync::MutexGuard;
 use std::sync::{Arc, Mutex, Weak};
+use std::sync::{Condvar, MutexGuard};
```
djc, line 3: Please merge the import statements for `std::sync`.

src/progress_bar.rs
```diff
@@ -1,12 +1,17 @@
...
 use std::time::{Duration, Instant};
 use std::{fmt, mem};
+use std::{io, thread};
```
djc, line 6: Please merge the import statements for `fmt, mem` and `io, thread`.

src/progress_bar.rs
```diff
@@ -1,12 +1,17 @@
...
+
+#[cfg(test)]
+use once_cell::sync::Lazy;
+#[cfg(test)]
+use std::sync::atomic::{AtomicBool, Ordering};
```
djc, line 11: Please move the `std::sync::atomic` imports up to the `std` block.

src/progress_bar.rs
```diff
@@ -506,10 +531,103 @@ impl WeakProgressBar {
...
         let state = self.state.upgrade()?;
         let pos = self.pos.upgrade()?;
-        Some(ProgressBar { state, pos })
+        let ticker = self.ticker.upgrade()?;
+        Some(ProgressBar { state, pos, ticker })
+    }
+}
+
+pub(crate) struct Ticker {
+    stopping_flag: Arc<Mutex<bool>>,
```
djc, line 540: So what value does the `_flag` suffix add?

src/progress_bar.rs
```diff
@@ -506,10 +531,103 @@ impl WeakProgressBar {
...
+                state.state.tick = state.state.tick.saturating_add(1);
+            }
+
+            state.draw(false, Instant::now()).ok();
+
+            drop(state); // Don't forget to drop the lock before sleeping
+            drop(arc); // Also need to drop Arc otherwise BarState won't be dropped
+
+            // Wait for `interval` but return early if we are notified to stop
+            let result = self
```
djc, line 609: How about we only bind the timeout result here, since we don't need the `MutexGuard` anyway? So something like `let (_, result) = `?

djc (commented):
Bunch of nits, but I think this is pretty close!

src/progress_bar.rs
djc: Yeah, maybe not, let's leave it for now.

src/progress_bar.rs
chris-laplante: Before I added the `_flag` suffix, I had `stopping` and `stopping_set` for the `Mutex` and `CondVar`, respectively. I thought `stopping_set` sounded ambiguous - i.e. does it have semantics of `HashSet` or something? So I added `_flag` to clarify. 

I guess if we merge the `Mutex<bool>` and `Condvar` into a single `Arc` we won't have to worry about naming though :)

src/progress_bar.rs
chris-laplante: Is it OK if I set up `rustfmt.toml` so this kind of stuff is handled automatically on `cargo fmt`?

djc: Sure, sounds great!

djc (approved):

djc:
Nice, thanks for seeing this through!

chris-laplante:
> Nice, thanks for seeing this through!

Of course, thanks for your guidance and feedback :)

stuhood:
Hey folks!

A bisect between `0.16.2` and `0.17.0` showed that this PR broke spinner updates for our use case. The `0.16.2`->`0.17.0` update diff looks like this: https://github.com/pantsbuild/pants/compare/main...stuhood:stuhood/indicatif-0.17

I don't see anything in the `0.17.0` release notes about changes to expectations around ticking: should we be manually calling `tick`, or is it required to `enable_steady_tick` now? One slightly unusual thing about our usecase is that we only use `bar.set_message`, and no other progress reporting methods.

djc:
@stuhood sorry about that, and thanks for doing the bisection! Can you file a new issue to track this problem with a minimal reproduction of your issue?

chris-laplante:
Sorry about the breakage :(

> One slightly unusual thing about our usecase is that we only use `bar.set_message`, and no other progress reporting methods.

Yes, I think that's what's doing it. A short-term fix should be to add a call to `tick` after you `set_message`.

Issue #391, opened 2022-03-01
Feature request: with_key taking full closures


kC0pter:
Adding keys with the with_key method currently only takes function pointers or non-capturing closures that are coercible to function pointers.
It would be a nice addition to change that to full closures to be able to capture locals to add to the printing.

djc:
The problem here is that `ProgressStyle` is `Clone` (which seems like a useful property), and it doesn't look like the keys could be `Clone` while also using dynamic dispatch.

1Dragoon:
I think the main thing we're looking for in with_keys closures is that we want to be able to store stateful information that the with_key method can then have read-only access to. So what if we could provide an Option<Box<dyn SomeTrait>> that ProgressState holds on to for us, maybe even ticks it for us, and our with_keys function pointer can then read from it?

Issue #394, opened 2022-03-12
Option to smooth ETA and *_per_second


aawsome:
First, thanks a lot for this wonderful crate. It is awesome to use!

I have a minor issue, I have a workload with highly changing operation rates (in bytes per seconds) which is kind of desired behavior as sometimes caching is used and sometimes not. Now, displaying ETA and bytes_per_second works but is of course very oscillating.
Is it possible to add an option which allows to smooth these value, e.g. that a longer time period is used for the calculation?

djc:
They are already smoothed to some extent, but not enough for your workload, I guess? On current main, you can add your own formatters, this would allow you to estimate based on total elapsed time / current position. Would that work for you?

I've been thinking about including formatters for that in the default set, likely called `linear_eta` (or maybe `uniform_eta`).

aawsome:
Thanks a lot for your fast answer! This is a very good idea, I'm now using

```rust
            .with_key("my_eta", |s| 
                 match (s.pos(), s.len()){
                    (0, _) => "-".to_string(),
                    (pos,len) => format!("{:#}", HumanDuration(Duration::from_secs(s.elapsed().as_secs() * (len-pos)/pos))),
                })

```
I'd appreciate to have some that in the default set of formatters. If you like, I can try to prepare a PR for this.

There is still some "flickering" for the rate *_per_seconds in my use case, but this can be very much tolerated as it correctly reflects the situation.

djc:
Can't you use a similar solution to make your own take on `_per_seconds`?

aawsome:
> Can't you use a similar solution to make your own take on `_per_seconds`?

Yes, of course that would work. But as this has more a "local" character (gives the current rate) I decided to simply use the default formatter.

For the ETA, this is more a of a global character and therefore I used the solution above.
What would you consider a good name? "linear_eta"? Or "global_eta"

If you like, I can submit a PR to add two default formatters "*_eta" and "*_eta_precise".

1Dragoon:
I was thinking the same thing just now, but also applicable to other measurements rather than just ETA. For example, bytes_per_sec keeps bouncing all over the place - it doesn't seem like it is a 1 second interval because I end up watching it bounce rapidly from the KiB scale to the GiB scale.

[...]

PR #420, opened 2022-03-25
Stateful Formatters

1Dragoon:
Very much a janky WIP, and some of the objects I introduced could use better naming though I avoided doing so in the interest of retaining compatibility with existing code. Haven't tested this against older rustc, but as of my current rustc version (latest stable) there are no lints from either regular check or clippy. But let me know if this seems like an ok direction and I'll revise it further.

~~Somewhen I'll modify that example to show what a more bumpy file transfer looks like, and the advantage of averaging it by the method I used.~~ done

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,7 +225,7 @@ mod progress_bar;
 #[cfg(feature = "rayon")]
 mod rayon;
 mod state;
-mod style;
+pub mod style;
 mod term_like;
 
 pub use crate::draw_target::ProgressDrawTarget;
--- a/src/state.rs
+++ b/src/state.rs
@@ -75,6 +75,11 @@ impl BarState {
         if let Reset::All = mode {
             self.state.pos.reset(now);
             self.state.status = Status::InProgress;
+
+            for (_, tracker) in self.style.format_map.iter_mut() {
+                tracker.reset(&self.state, now);
+            }
+
             let _ = self.draw(false, now);
         }
     }
@@ -117,6 +122,10 @@ impl BarState {
         let pos = self.state.pos.pos.load(Ordering::Relaxed);
         self.state.est.record(pos, now);
         let _ = self.draw(false, now);
+
+        for (_, tracker) in self.style.format_map.iter_mut() {
+            tracker.tick(&self.state, now);
+        }
     }
 
     pub(crate) fn println(&mut self, now: Instant, msg: &str) {
--- a/src/style.rs
+++ b/src/style.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::fmt::{self, Write};
 use std::mem;
+use std::time::Instant;
 
 use console::{measure_text_width, Style};
 #[cfg(feature = "unicode-segmentation")]
@@ -11,7 +12,6 @@ use crate::format::{
 };
 use crate::state::{ProgressState, TabExpandedString, DEFAULT_TAB_WIDTH};
 
-/// Controls the rendering style of progress bars
 #[derive(Clone)]
 pub struct ProgressStyle {
     tick_strings: Vec<Box<str>>,
@@ -19,8 +19,8 @@ pub struct ProgressStyle {
     template: Template,
     // how unicode-big each char in progress_chars is
     char_width: usize,
-    format_map: HashMap<&'static str, fn(&ProgressState) -> String>,
     tab_width: usize,
+    pub(crate) format_map: HashMap<&'static str, Box<dyn ProgressTracker>>,
 }
 
 #[cfg(feature = "unicode-segmentation")]
@@ -136,9 +136,13 @@ impl ProgressStyle {
         self
     }
 
-    /// Adds a custom key that references a `&ProgressState` to the template
-    pub fn with_key(mut self, key: &'static str, f: fn(&ProgressState) -> String) -> ProgressStyle {
-        self.format_map.insert(key, f);
+    /// Adds a custom key that owns a [`ProgressTracker`] to the template
+    pub fn with_key<S: ProgressTracker + 'static>(
+        mut self,
+        key: &'static str,
+        f: S,
+    ) -> ProgressStyle {
+        self.format_map.insert(key, Box::new(f));
         self
     }
 
@@ -150,7 +154,7 @@ impl ProgressStyle {
         Ok(self)
     }
 
-    pub(crate) fn current_tick_str(&self, state: &ProgressState) -> &str {
+    fn current_tick_str(&self, state: &ProgressState) -> &str {
         match state.is_finished() {
             true => self.get_final_tick_str(),
             false => self.get_tick_str(state.tick),
@@ -237,8 +241,8 @@ impl ProgressStyle {
                     alt_style,
                 } => {
                     buf.clear();
-                    if let Some(formatter) = self.format_map.get(key.as_str()) {
-                        buf.push_str(&formatter(state).replace('\t', &" ".repeat(self.tab_width)));
+                    if let Some(tracker) = self.format_map.get(key.as_str()) {
+                        tracker.write(state, &mut TabRewriter(&mut buf, self.tab_width));
                     } else {
                         match key.as_str() {
                             "wide_bar" => {
@@ -362,6 +366,15 @@ impl ProgressStyle {
                 None => mem::take(&mut cur),
             })
         }
+    }
+}
+
+struct TabRewriter<'a>(&'a mut dyn fmt::Write, usize);
+
+impl Write for TabRewriter<'_> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.0
+            .write_str(s.replace('\t', &" ".repeat(self.1)).as_str())
     }
 }
 
@@ -691,12 +704,99 @@ enum Alignment {
     Right,
 }
 
+/// Trait for defining stateful or stateless formatters
+pub trait ProgressTracker: Send {
+    /// Creates a new instance of the progress tracker
+    fn clone_box(&self) -> Box<dyn ProgressTracker>;
+    /// Notifies the progress tracker of a tick event
+    fn tick(&mut self, state: &ProgressState, now: Instant);
+    /// Notifies the progress tracker of a reset event
+    fn reset(&mut self, state: &ProgressState, now: Instant);
+    /// Provides access to the progress bar display buffer for custom messages
+    fn write(&self, state: &ProgressState, w: &mut dyn fmt::Write);
+}
+
+impl Clone for Box<dyn ProgressTracker> {
+    fn clone(&self) -> Self {
+        self.clone_box()
+    }
+}
+
+impl<F: Fn(&ProgressState, &mut dyn fmt::Write) + Send + Clone + 'static> ProgressTracker for F {
+    fn clone_box(&self) -> Box<dyn ProgressTracker> {
+        Box::new(self.clone())
+    }
+
+    fn tick(&mut self, _: &ProgressState, _: Instant) {}
+
+    fn reset(&mut self, _: &ProgressState, _: Instant) {}
+
+    fn write(&self, state: &ProgressState, w: &mut dyn fmt::Write) {
+        (self)(state, w)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::sync::Arc;
 
     use super::*;
-    use crate::state::{AtomicPosition, ProgressState, TabExpandedString};
+    use crate::state::{AtomicPosition, ProgressState};
+    use std::sync::Mutex;
+
+    #[test]
+    fn test_stateful_tracker() {
+        #[derive(Debug, Clone)]
+        struct TestTracker(Arc<Mutex<String>>);
+
+        impl ProgressTracker for TestTracker {
+            fn clone_box(&self) -> Box<dyn ProgressTracker> {
+                Box::new(self.clone())
+            }
+
+            fn tick(&mut self, state: &ProgressState, _: Instant) {
+                let mut m = self.0.lock().unwrap();
+                m.clear();
+                m.push_str(format!("{} {}", state.len().unwrap(), state.pos()).as_str());
+            }
+
+            fn reset(&mut self, _state: &ProgressState, _: Instant) {
+                let mut m = self.0.lock().unwrap();
+                m.clear();
+            }
+
+            fn write(&self, _state: &ProgressState, w: &mut dyn fmt::Write) {
+                w.write_str(self.0.lock().unwrap().as_str()).unwrap()
+            }
+        }
+
+        use crate::ProgressBar;
+
+        let pb = ProgressBar::new(1);
+        pb.set_style(
+            ProgressStyle::with_template("{{ {foo} }}")
+                .unwrap()
+                .with_key("foo", TestTracker(Arc::new(Mutex::new(String::default()))))
+                .progress_chars("#>-"),
+        );
+
+        let mut buf = Vec::new();
+        let style = pb.clone().style();
+
+        style.format_state(&pb.state().state, &mut buf, 16);
+        assert_eq!(&buf[0], "{  }");
+        buf.clear();
+        pb.inc(1);
+        style.format_state(&pb.state().state, &mut buf, 16);
+        assert_eq!(&buf[0], "{ 1 1 }");
+        pb.reset();
+        buf.clear();
+        style.format_state(&pb.state().state, &mut buf, 16);
+        assert_eq!(&buf[0], "{  }");
+        pb.finish_and_clear();
+    }
+
+    use crate::state::TabExpandedString;
 
     #[test]
     fn test_expand_template() {
@@ -706,8 +806,14 @@ mod tests {
         let mut buf = Vec::new();
 
         let mut style = ProgressStyle::default_bar();
-        style.format_map.insert("foo", |_| "FOO".into());
-        style.format_map.insert("bar", |_| "BAR".into());
+        style.format_map.insert(
+            "foo",
+            Box::new(|_: &ProgressState, w: &mut dyn Write| write!(w, "FOO").unwrap()),
+        );
+        style.format_map.insert(
+            "bar",
+            Box::new(|_: &ProgressState, w: &mut dyn Write| write!(w, "BAR").unwrap()),
+        );
 
         style.template = Template::from_str("{{ {foo} {bar} }}").unwrap();
         style.format_state(&state, &mut buf, WIDTH);
@@ -730,7 +836,10 @@ mod tests {
         let mut buf = Vec::new();
 
         let mut style = ProgressStyle::default_bar();
-        style.format_map.insert("foo", |_| "XXX".into());
+        style.format_map.insert(
+            "foo",
+            Box::new(|_: &ProgressState, w: &mut dyn Write| write!(w, "XXX").unwrap()),
+        );
 
         style.template = Template::from_str("{foo:5}").unwrap();
         style.format_state(&state, &mut buf, WIDTH);
--- a/examples/download.rs
+++ b/examples/download.rs
@@ -1,8 +1,8 @@
-use std::cmp::min;
 use std::thread;
 use std::time::Duration;
+use std::{cmp::min, fmt::Write};
 
-use indicatif::{ProgressBar, ProgressStyle};
+use indicatif::{ProgressBar, ProgressState, ProgressStyle};
 
 fn main() {
     let mut downloaded = 0;
@@ -11,7 +11,7 @@ fn main() {
     let pb = ProgressBar::new(total_size);
     pb.set_style(ProgressStyle::with_template("{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})")
         .unwrap()
-        .with_key("eta", |state| format!("{:.1}s", state.eta().as_secs_f64()))
+        .with_key("eta", |state: &ProgressState, w: &mut dyn Write| write!(w, "{:.1}s", state.eta().as_secs_f64()).unwrap())
         .progress_chars("#>-"));
 
     while downloaded < total_size {


1Dragoon:
Note I moved the estimator to it's own module. It had to be public for this to work, and I'm not sure whether you guys would want to make the state module public, and it didn't really seem to fit anywhere else.

1Dragoon:
The more I think about this, the more I wonder what other use cases somebody might have for needing this level of access to the API, and I might be over-engineering it as a result. This might be better done with allowing the user to specify whether estimations are done per tick, as the default estimator does, or over n seconds, milliseconds, etc. I think my example merely demonstrates a case where the default estimator simply doesn't work at all (basically the numbers are completely meaningless to the user.)

I think the efficiency improvements I've made in my alternative estimator might even make it more efficient than the existing one, but at the same time it probably shouldn't outright replace the existing one as I could see many people preferring that per-tick estimation that is currently used, and per-tick should probably remain the default as it is much more precise.

Let me know what your thoughts are, or you can outright tell me that I'm totally bonkers and I'll accept that too :)

1Dragoon:
The last push I did gets everything working the way I need it to. Please see the rate_smoothing example for the overall effect of what it is I'm trying to do, with example real world use cases mentioned in the comments.

I'm also tempted to implement a lock-free incrementing function, but I need to evaluate whether that will offer much benefit in what I'm working on.

1Dragoon:
That's odd, I compiled against rustc 1.53 and there were no lints, and all tests passed. I even added some very aggressive clippy lints and they were all clear. I also ran the cargo format check exactly as those checks do. All of that was done before being pushed. I'll look at it again to see if something slipped by somehow.

djc:
It's very possible that clippy has gotten more strict since 1.53, especially if you enable all those extra strict stuff. 1.53 is pretty old at this point, maybe upgrade to current stable?

1Dragoon:
Oh I thought those tests were specifically against 1.53? Sure, I'll use the latest stable, I used rustup to override to 1.53 so that I could guarantee compatibility with 6 months prior as you had asked in my other PR.

djc:
We run the tests against 1.53 to ensure the library works well, but all the linking happens on stable.

djc:
For this to be mergable, I'd like to:

* Have the map be a `std::collections::HashMap`, not a `DashMap`
* Move the map back into `ProgressStyle`
* Keep the `Estimator` as is for now
* If the `Default` parameter doesn't work, we can add a `default()` method to the trait that yields a trait object

As I've tried to explain a few times, this should be a minimal change.

1Dragoon:
Not sure what you mean move it 'back'? The format_map is still in progressstate. That particular map I made into a dashmap had to be that way because it originally did live in progressstate, because it provides internal mutability (it was Arc<DashMap>, hence why those variables were named 'adm'. It doesn't need to be dashmap if it's in progressstate though.

While I can put some sort of default in there, I'm not sure where you are wanting it to be called from.

The only other thing I can do is require that statefulformatter is immutable with internal mutability, which complicates things a lot from the user side as it will probably need atomics even if they're not really appropriate for the use case, or it forces them into using locks. I also don't really know how to meet my own use case while falling within those constraints. I'm looking it over to see what can be done, but right now I don't know.

djc:
> Not sure what you mean move it 'back'? The format_map is still in progressstate.

Sorry, I meant we should not add a map to `BarState` and only replace the `format_map` with something more capable.

> While I can put some sort of default in there, I'm not sure where you are wanting it to be called from.

From the `impl` for `ProgressStyle::clone()`.

> The only other thing I can do is require that statefulformatter is immutable with internal mutability

Why? `BarState::tick()` has mutable access to the `ProgressStyle`, right?

1Dragoon:
> Why? `BarState::tick()` has mutable access to the `ProgressStyle`, right?

Right but in order to live inside of progressstyle it has to be Clone. I don't know how to do that while also making it thread safe by adding Arc, which removes mutability. Unless progressstyle isn't meant to be cloned across threads?

djc:
You can remove the derived Clone impl in favor of manual Clone impl, that gives you the flexibility to do something different for the format_map values.

1Dragoon:
Ah I see. I'll chew on this for a bit.

1Dragoon:
I'm still not really understanding the use case for cloning ProgressStyle. When ProgressStyle is cloned, how do we resume the state from the existing instance? How do we keep the new one syncronized?

1Dragoon:
Well whatever the use case is, it doesn't seem to impact mine at all. Still a draft, there are a few things I need to clean up, but overall let me know if this works.

1Dragoon:
I think I'm mostly satisfied with what I have there. I'd really like the ability to remove or at the very least suppress the built-in estimator as it will be running (I think? unless the compiler sees it as dead code somehow and removes it anyways) even though my use case doesn't need it at all. Also I still don't understand the need for cloning ProgressStyle, but it doesn't interfere with anything I'm doing in any way so no problem there.

.rusty-hook.toml
```diff
@@ -0,0 +1,2 @@
+[hooks]
```
djc, line 1: I don't want to add this to the repo.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ once_cell = "1"
 rand = "0.8"
 structopt = "0.3"
 tokio = { version = "1", features = ["time", "rt"] }
+rusty-hook = "0.11.2"
```
djc, line 28: So this should also be removed.

examples/rate_smoothing.rs
```diff
@@ -0,0 +1,218 @@
+#![warn(clippy::pedantic)]
```
djc, line 1: I'd prefer not to add an example for this right now.

src/lib.rs
```diff
@@ -243,3 +246,9 @@ pub use crate::term_like::TermLike;
 
 #[cfg(feature = "rayon")]
 pub use crate::rayon::ParallelProgressIterator;
+
+pub trait StatefulFormatter: Display + Send {
```
djc, line 250: This should live near the bottom of the `style` module, please rename it to `ProgressTracker`.

src/lib.rs
```diff
@@ -243,3 +246,9 @@ pub use crate::term_like::TermLike;
...
+    fn default(&self) -> Box<dyn StatefulFormatter>;
+    fn tick(&mut self, state: &mut ProgressState, now: Instant);
```
djc, line 252: I think `tick()` and `reset()` should take `&ProgressState`, not `&mut`.

src/progress_bar.rs
```diff
@@ -242,6 +241,16 @@ impl ProgressBar {
         self.state().reset(Instant::now(), Reset::Elapsed)
     }
 
+    /// Set elapsed time for i.e. resuming where the previous elapsed time is known
```
djc, line 244: These methods are unrelated.

src/state.rs
```diff
@@ -63,6 +63,10 @@ impl BarState {
         let _ = self.draw(true, now);
     }
 
+    pub(crate) fn set_elapsed(&mut self, duration: Duration) {
```
djc, line 66: Remove this.

src/state.rs
```diff
@@ -72,6 +76,14 @@ impl BarState {
             self.state.started = now;
         }
 
+        if let Reset::Formatters | Reset::All = mode {
```
djc, line 79: Let's not do a separate `Reset` variant for this, just reset the formatters for `All`.

src/style.rs
```diff
@@ -4,16 +4,32 @@ use std::fmt::{self, Write};
...
 #[cfg(feature = "unicode-segmentation")]
 use unicode_segmentation::UnicodeSegmentation;
 
 use crate::format::{
     BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, HumanCount, HumanDuration,
 };
 use crate::state::ProgressState;
+use crate::StatefulFormatter;
+
+pub(crate) enum FormatterVariant {
```
djc, line 17: No need to have this enum, just make everything stateful. Might be nice to add a blanket impl for `impl<F: Fn(&ProgressState) -> String)> ProgressTracker for F` if possible.

src/style.rs
```diff
@@ -4,16 +4,32 @@ use std::fmt::{self, Write};
...
+        match self {
+            FormatterVariant::Stateless(f) => FormatterVariant::Stateless(*f),
+            FormatterVariant::Stateful(f) => FormatterVariant::Stateful(f.default()),
+        }
+    }
+}
 
-/// Controls the rendering style of progress bars
 #[derive(Clone)]
+/// Controls the rendering style of progress bars
```
djc, line 32: Please revert this change, as it's unrelated.

src/style.rs
```diff
@@ -712,7 +747,9 @@ mod tests {
         let mut buf = Vec::new();
 
         let mut style = ProgressStyle::default_bar();
-        style.format_map.insert("foo", |_| "XXX".into());
+        style
+            .format_map
+            .insert("foo", FormatterVariant::Stateless(|_| "XXX".into()));
 
         style.template = Template::from_str("{foo:5}").unwrap();
         style.format_state(&state, &mut buf, WIDTH);
```
djc, line 755: Please add a tes for a stateful `ProgressTracker`.

1Dragoon:
The hook is entirely for my own convenience, it will go once we've got everything else sorted out.

To a lesser extent, the example is as well. It mainly works to ensure that my use case remains valid without needing me to screw with my other repos in the meantime, and ultimately will provide a working model after.

Just let me know once we're down to these two and I'll take them out last

src/lib.rs
1Dragoon: done

examples/rate_smoothing.rs
1Dragoon: this will be removed last, let's just make sure everything else is good first

src/style.rs
djc: We're about to release 0.17 which IIRC is the first release to introduce this anyway. This change does not need to be additive.

1Dragoon: Oh is my change (tentatively) intended to be part of that release? I figured this would be delayed for something after.

djc: No, the 0.17.0-rc releases are from main.

djc:
You could just keep the example in your local workdir uncommitted, similar for the rusty-hook config (though that should really just be a Git pre-push hook probably). Putting stuff in your PR that is not slated to be merged makes reviewing harder.

1Dragoon:
Git always complains when I do that, there's probably more to its usage that I need to figure out

djc:
@1Dragoon do you still want to finish this? Would be nice; I think we're going to push out another rc soon and hopefully release 0.17.0 soon after.

1Dragoon:
I've been out on vacation, I'll get back to it some time next week

djc:
Great, thanks!

1Dragoon:
One thing I'm having difficulty making time for is that making ProgressTracker a single type rather than having the variant type is that I need to refactor all of the example code, but other than that what I have locally works

1Dragoon:
I'm a bit stuck. I put a TODO comment in style.rs showing where the problem is. Basically in order to fall within your requirement of having clone and having both stateful and stateless trackers as a single trait.

It is nice being able to toss either a simple closure or a more complex tracker type as the same parameter, but I'm not certain how to pull that off given the need for clone to be implemented. Little help anybody?

djc:
Sorry, did you mean on line 727? I didn't see an actual `TODO` anywhere. I think it's fine to ditch the closure option for now and require an instantce of a `ProgressTracker`-implementing type.

1Dragoon:
Oh sorry I accidentally committed before adding the todo comment. It's line 25 of style.rs. What I put there is basically just a filler. I just pushed the corrected one with the comment added.

djc:
Hmm, I think an `Fn` implementation must be stateless anyway, so we can just reuse the `Self` value? Otherwise I would just keep out the `impl<F: Fn(..) -> ..>` for now, let's tackle one thing at a time.

1Dragoon:
> Hmm, I think an `Fn` implementation must be stateless anyway, so we can just reuse the `Self` value? Otherwise I would just keep out the `impl<F: Fn(..) -> ..>` for now, let's tackle one thing at a time.

I'm not sure what you mean. Trait definitions can't return Self, or am I misunderstanding?

Anyways I just pushed a build that just relies on a single trait definition

1Dragoon:
Got it!

Can you review all of the code again? I can't recall what else I needed to do.

Cargo.toml
1Dragoon: Removed

src/style.rs
1Dragoon: Done

1Dragoon:
Merged in the latest upstream commits and resolved the conflict.

One thing though -- I don't really like the default() method name because it's a bit misleading, and it may conflict if the user had a struct that they actually wanted to implement the real default trait for. Not sure what else to call it though, maybe respawn()?

examples/rate_smoothing.rs
djc: Let's remove this, you can just keep it in your local working dir for testing.

src/state.rs
```diff
@@ -111,6 +116,12 @@ impl BarState {
         let pos = self.state.pos.pos.load(Ordering::Relaxed);
         self.state.est.record(pos, now);
         let _ = self.draw(false, now);
+
+        // No real way to do this as a normal iterator without doing some hacky stuff
```
djc, line 120: I think we can leave out these comments, it doesn't seem very important/useful.

src/style.rs
```diff
@@ -2,8 +2,10 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt::{self, Write};
 use std::mem;
+use std::time::Instant;
 
 use console::{measure_text_width, Style};
+
```
djc, line 8: Let's keep this as it was.

src/style.rs
```diff
@@ -235,10 +249,9 @@ impl ProgressStyle {
                     alt_style,
                 } => {
                     buf.clear();
-                    if let Some(formatter) = self.format_map.get(key.as_str()) {
-                        buf.push_str(&formatter(state));
-                    } else {
-                        match key.as_str() {
+                    match self.format_map.get(key.as_str()) {
```
djc, line 252: I'd prefer to minimize the changes here to make this easier to review.

src/style.rs
```diff
@@ -674,12 +687,99 @@ enum Alignment {
     Right,
 }
 
+/// Trait for defining stateful or stateless formatters
```
djc, line 690: I don't think we should elaborate on "stateful" vs "stateless" here, this trait simply facilitates both. People can see the impl for `Fn(&ProgressState) -> String` so should be clear that stateless is an option. Instead, should talk more about the core idea this enables, keeping your own state based on the basic `ProgressState` and generating output from that.

Maybe also call out how `default()` is used.

src/style.rs
```diff
@@ -134,9 +135,22 @@ impl ProgressStyle {
         self
     }
 
-    /// Adds a custom key that references a `&ProgressState` to the template
-    pub fn with_key(mut self, key: &'static str, f: fn(&ProgressState) -> String) -> ProgressStyle {
-        self.format_map.insert(key, f);
+    /// Adds a custom key that owns a [`ProgressTracker`] to the template. These trackers can be either stateful or stateless. Stateful trackers simply need to implement the `ProgressTracker` trait.
```
djc, line 138: I'd like to keep the first line as it was, without a period, and end there. Any further elaboration can go after an empty line. Here too, I don't think the documentation's focus should be on stateful vs stateless, but instead focus on the capability this enables.

src/style.rs
```diff
@@ -674,12 +687,99 @@ enum Alignment {
...
 
+/// Trait for defining stateful or stateless formatters
+///
+/// Stateless formatters are the more simple type and are implemented as a closure in the form of [`Fn(&ProgressState) -> String`]
+/// Stateful formatters can be anything that implements this trait
+pub trait ProgressTracker: Send {
+    fn default(&self) -> Box<dyn ProgressTracker>;
+    fn tick(&mut self, state: &ProgressState, now: Instant);
+    fn reset(&mut self, state: &ProgressState, now: Instant);
+    fn message(&self, state: &ProgressState) -> String;
```
djc, line 698: If you're still open to working on this, we should change `fn message(&self, state: &ProgressState) -> String` to `fn write(&self, state: &ProgressState, w: &mut dyn fmt::Write) -> fmt::Result`.

djc (commented):
This seems to be in a pretty good state, I'm almost ready to merge!

src/style.rs
1Dragoon: Ah yeah that was leftover from the previous way I was doing it, I agree it's more idiomatic the previous way now

src/style.rs
djc: Yeah, renaming it is good!

examples/rate_smoothing.rs
1Dragoon: I'm going to keep this until everything else looks good. The reason why is because I want to continually evaluate whether the syntax I'm able to use looks correct and idiomatic. It acts as a sanity check, and when I refactor anything crate side, it auto-refactors on my side as well

djc: Yes: which is why I suggested you keep this in your local clone without committing it into your Git commits. This should support your refactoring work without me having to look at the changes, and, since I am close to merging this, have to do an extra round trip to ask you to remove it.

src/progress_bar.rs
```diff
@@ -585,6 +585,7 @@ impl Ticker {
         }
     }
 
+    #[allow(clippy::mutex_atomic)]
```
djc, line 588: Why do we need this?

src/style.rs
```diff
@@ -235,8 +239,8 @@ impl ProgressStyle {
                     alt_style,
                 } => {
                     buf.clear();
-                    if let Some(formatter) = self.format_map.get(key.as_str()) {
-                        buf.push_str(&formatter(state));
+                    if let Some(f) = self.format_map.get(key.as_str()) {
+                        f.write(state, &mut buf).unwrap_or_default(); // Should this silently fail or panic?
```
djc, line 243: Just `.unwrap()` it for now.

src/style.rs
```diff
@@ -674,12 +678,96 @@ enum Alignment {
     Right,
 }
 
+/// Trait for defining stateful or stateless formatters
+pub trait ProgressTracker: Send {
+    fn spawn(&self) -> Box<dyn ProgressTracker>;
```
djc, line 683: Hmm, don't like the `spawn()` name much. What about `cloned()`?

src/style.rs
```diff
@@ -674,12 +678,96 @@ enum Alignment {
...
+impl Clone for Box<dyn ProgressTracker>
+where
+    Self:,
+{
+    fn clone(&self) -> Self {
+        self.spawn()
+    }
+}
+
+impl<F: Fn(&ProgressState) -> String + Send + Clone + 'static> ProgressTracker for F {
```
djc, line 698: Can we change the `Fn` here to take the `fmt::Write`, too?

1Dragoon:
@djc Have you had a chance to review my comments yet?

djc:
Uh, I think I already left a round of comments here? I was waiting for you.

1Dragoon:
Hmm is there some github functionality I'm missing? I just noticed that my comments only appear if I'm logged in:

image

1Dragoon:
Oh I figured it out, have to go to the file changes page and then click the review button. They really ought to add some kind of UI hint that you need to do that before anybody can even see your comments, otherwise there's nothing at all to indicate that you have to do so.

src/style.rs
djc: I mean, currently we have `F` return a `String`. We changed the trait method to write into a `fmt::Formatter<'_>` instead, can we change the `F` bound to be similar (that is, avoid allocating a `String` here)?

src/style.rs
```diff
@@ -674,12 +678,98 @@ enum Alignment {
     Right,
 }
 
+/// Trait for defining stateful or stateless formatters
+pub trait ProgressTracker: Send {
+    fn cloned(&self) -> Box<dyn ProgressTracker>;
```
1Dragoon, line 683: On second thought, what if we called it derive() or derived()? I think that would pretty succinctly describe what it does.

src/style.rs
```diff
@@ -674,12 +678,96 @@ enum Alignment {
...
+impl Clone for Box<dyn ProgressTracker>
+where
+    Self:,
+{
+    fn clone(&self) -> Self {
+        self.cloned()
+    }
+}
+
+impl<F: Fn(&ProgressState, &mut dyn fmt::Write) + Send + Clone + 'static> ProgressTracker for F {
```
1Dragoon, line 698: Like this? I figured not having a return at all is a better option as it lets the user decide how they want to handle any failures rather than allowing a potential panic in the crate.

1Dragoon:
BTW I'd still like to be able to tell the built in estimator to noop if we're not using it.

src/style.rs
djc: It's not an obvious improvement to me... @chris-laplante what do you think?

Also would be good to add some documentation to these trait method definitions.

src/style.rs
djc: Yes, this seems reasonable.

src/style.rs
djc: Yup, let's do clone_box!

1Dragoon: All set

1Dragoon (commented):
Other than my concern about the built-in estimator, I'm satisfied with this.

1Dragoon:
@djc any further concerns with this?

djc (approved):

src/style.rs
```diff
@@ -674,12 +678,100 @@ enum Alignment {
...
+    /// Notifies the progress tracker of a tick event
+    fn tick(&mut self, state: &ProgressState, now: Instant);
+    /// Notifies the progress tracker of a reset event
+    fn reset(&mut self, state: &ProgressState, now: Instant);
+    /// Provides access to the progress bar display buffer for custom messages
+    fn write(&self, state: &ProgressState, w: &mut dyn fmt::Write);
+}
+
+impl Clone for Box<dyn ProgressTracker>
+where
```
djc, line 694: Can you ditch this useless `where` clause?

djc (commented):
No, can you rebase it one more time? Feel free to just squash all the changes in one commit (or I can do that when merging).

chris-laplante:
@1Dragoon want me to take a crack at fixing the conflicts? (since they were caused by my most recent commits). I think I can make the changes right in this PR.

1Dragoon:
Sure, sorry I haven't gotten to this yet because I've been sick

1Dragoon:
I'm feeling a bit better this afternoon so I went ahead and squashed my repo and started to resolve the merge confilcts but this one has me a bit confused:

```rust
                    if let Some(formatter) = self.format_map.get(key.as_str()) {
                        formatter.push_str(&formatter(state).replace('\t', &" ".repeat(self.tab_width)));
                    }
```

I think this might need to be moved to the default impl, still looking at it

1Dragoon:
I'm not sure that I like what I had to do with line 246, but it works though

src/style.rs
1Dragoon: done

chris-laplante:
> I'm not sure that I like what I had to do with line 246, but it works though
> 
> @djc please review when you can

It looks OK to me.

src/style.rs
```diff
@@ -237,8 +241,9 @@ impl ProgressStyle {
                     alt_style,
                 } => {
                     buf.clear();
-                    if let Some(formatter) = self.format_map.get(key.as_str()) {
-                        buf.push_str(&formatter(state).replace('\t', &" ".repeat(self.tab_width)));
+                    if let Some(tracker) = self.format_map.get(key.as_str()) {
+                        tracker.write(state, &mut buf);
+                        buf = buf.replace('\t', &" ".repeat(self.tab_width));
```
djc, line 246: I think we should do a `struct TabRewriter(&mut dyn fmt::Write);` with an `impl fmt::Write` that rewrites the tabs before writing them to the inner `fmt::Write` instead.

Does that seem manageable?

1Dragoon: Not sure I follow. Where would we pass tab_width to that?

1Dragoon:
@djc See the latest commit. I think that's what you're looking for?

Also should probably mark this to resolve #391 and #394, not sure how that works though.

djc:
Great work, thanks for sticking with it!

1Dragoon:
Sure, no problem.

Also not a huge deal, but will it ever show me as a contributor here? Reason I ask is because I am thinking that at some point I might show potential future employers what kind of work I've done, and I'm not sure how I'll be able to do that once I've deleted my forked copy (for cleanliness sake.)

djc:
Yes, you should show up as a contributor for this project (or you can send the URL to this PR).

PR #421, opened 2022-03-29
add naive test for ticker termination

chris-laplante:
I'd say if we can get this test to pass, then we can ensure that `Ticker` is behaving properly on `ProgressBar` drop.


djc:
Closing this in favor of #417.

PR #423, opened 2022-03-31
Expand tabs => configurable number of spaces to fix #150

### LINKED ISSUE #150: Tabs mess up bar width calculation
Indicatif appears to not understand how many columns a tab will take up in a terminal. The result is that "wide messages", for example, will wrap to a new line before they are truncated, and bars will re-render down the screen instead of in-place.

I discovered this while building a program that runs arbitrary external commands and displays their most recent line of stdout or stderr on a progress bar. I was scratching my head for a while until I realized that some of those external commands are printing tabs to stderr!

Observed on Mac, using both iTerm2 and the inbuilt terminal. Not sure if it affects other platforms.

My workaround is to simply `.replace("\t", "    ")` before setting the bar message. I'm guessing there's some way to interrogate a terminal and discover how wide a tab will be, and that could be used in width calculations. But I don't know enough about terminals to know if that's safe enough to rely on across all platforms, or whether replacing tabs with spaces would be a more pragmatic solution. Thoughts?

Minimal-ish reproduction:

```rust
use std::thread;
use std::time::Duration;

const LOREM_IPSUM: &str = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis ac facilisis est. Nulla commodo sit amet nulla non maximus. Ut in magna quis nibh ullamcorper volutpat at et purus. Praesent non libero in magna pellentesque pellentesque a nec nisl. Nulla lorem quam, maximus suscipit dapibus vitae, mattis a felis. In volutpat tellus elit, sed laoreet felis tempus vitae. Aenean aliquam, risus vitae sagittis lobortis, nunc tellus iaculis sapien, vel consectetur neque tellus pharetra tortor. Quisque at semper leo. Praesent tellus lectus, bibendum vitae nisl in, posuere placerat nunc. Aliquam dapibus, lorem dignissim suscipit suscipit, mi urna malesuada nunc, condimentum sollicitudin orci massa et nisi. Nam purus purus, sodales commodo augue eget, suscipit iaculis nulla.";
const LOREM_IPSUM_WITH_TABS: &str = "Lorem ipsum\tdolor sit\tamet, consectetur adipiscing elit. Duis ac facilisis est. Nulla commodo sit amet nulla non maximus. Ut in magna quis nibh ullamcorper volutpat at et purus. Praesent non libero in magna pellentesque pellentesque a nec nisl. Nulla lorem quam, maximus suscipit dapibus vitae, mattis a felis. In volutpat tellus elit, sed laoreet felis tempus vitae. Aenean aliquam, risus vitae sagittis lobortis, nunc tellus iaculis sapien, vel consectetur neque tellus pharetra tortor. Quisque at semper leo. Praesent tellus lectus, bibendum vitae nisl in, posuere placerat nunc. Aliquam dapibus, lorem dignissim suscipit suscipit, mi urna malesuada nunc, condimentum sollicitudin orci massa et nisi. Nam purus purus, sodales commodo augue eget, suscipit iaculis nulla.";

fn main() {
    let pb = ProgressBar::new(1024);
    let style = ProgressStyle::default_bar().template("{pos} {wide_msg}");
    pb.set_style(style);
    for i in 0..1024 {
        pb.inc(1);
        if i < 500 {
            pb.set_message(LOREM_IPSUM);
        } else {
            pb.set_message(LOREM_IPSUM_WITH_TABS);
        }
        thread::sleep(Duration::from_millis(5));
    }
}
```

If you run this, as soon as `i >= 500` the bar will start "smearing" down the screen, presumably because the accidental line wrapping makes Indicatif's math for how many lines it should jump up to re-render the bar incorrect.

If I replace the tabs with emojis, the same thing happens in reverse; indicatif appears to expect the emojis to take up more columns than they do in practice, and so the line is truncated early instead.

chris-laplante:


src/style.rs
```diff
@@ -23,6 +23,7 @@ pub struct ProgressStyle {
     // how unicode-big each char in progress_chars is
     char_width: usize,
     format_map: HashMap<&'static str, fn(&ProgressState) -> String>,
+    tab_width: usize,
```
djc, line 26: If we make this a `tab_str: String` we don't have to do `iter::repeat()` thing in `format_state()`. We can still keep `with_tab_width()` taking `usize` and expand it there.

src/style.rs
```diff
@@ -348,7 +358,9 @@ impl ProgressStyle {
                         },
                     }
                 }
-                TemplatePart::Literal(s) => cur.push_str(s),
+                TemplatePart::Literal(s) => {
+                    cur.push_str(&s.replace('\t', &" ".repeat(self.tab_width)))
```
djc, line 362: How about we pre-replace tabs in literals after parsing? Would probably warrant a warning in the `with_tab_size()` docs that you should call it before `template()`, but would save on the work being done here in a somewhat hot path.

src/style.rs
```diff
@@ -398,7 +410,7 @@ impl<'a> WideElement<'a> {
                 buf.write_fmt(format_args!(
                     "{}",
                     PaddedStringDisplay {
-                        str: &style.message,
+                        str: &style.message.replace('\t', &" ".repeat(style.tab_width)),
```
djc, line 413: We can also pre-replace in messages.

src/style.rs
```diff
@@ -327,6 +335,8 @@ impl ProgressStyle {
                         continue;
                     }
 
+                    buf = buf.replace('\t', &" ".repeat(self.tab_width));
```
djc, line 338: It would be nice to avoid extra allocations here in the common case where there are no tabs at all.

src/style.rs
```diff
@@ -91,6 +92,7 @@ impl ProgressStyle {
             char_width,
             template,
             format_map: HashMap::default(),
+            tab_width: 4,
```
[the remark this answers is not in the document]
chris-laplante: I chose 4 because that's what my (Linux Mint, Ubuntu based) system seems to use. I can make it 8 if you prefer.

src/style.rs
chris-laplante: Perhaps we could save the original template string and re-expand it when with_tab_width is called? That way user doesn't have to worry about the order of calls.

djc: That seems reasonable, yes.

src/style.rs
chris-laplante: In this case we'd also need to store the original message and re-expand it if tab width changes. Are you OK with that complexity?

chris-laplante: (same with prefix)

chris-laplante:
> Thinking about this some more, you replaced tabs at the line level, but most of the formatters can't actually emit tabs. Probably only from literals, message and prefix? So it might pay off to be precise about that.

Yes, I agree. The code changes I am working on now will only replace inside those elements. Just wanted to check if you think it is acceptable complexity to also store away the original message & prefix and re-expand them when tab width changes?

djc:
How about we only do that if message/prefix/template actually contained any tabs?

chris-laplante:
> How about we only do that if message/prefix/template actually contained any tabs?

So like this?

```
enum StringMaybeTabs {
    NoTabs(Cow<'static, str>),
    Tabs {
        original: Cow<'static, str>,
        expanded: String,
    }
}
```

djc:
Something like that, yes. I'll need to see it in context.

djc:
I wonder if we can query the terminal for its tab size?

https://unix.stackexchange.com/questions/46368/set-tab-width-in-gui-terminal

chris-laplante:
> I wonder if we can query the terminal for its tab size?
> 
> https://unix.stackexchange.com/questions/46368/set-tab-width-in-gui-terminal

I looked into that but I don't think there's any way to query it unless you actually print a tab and measure the difference in cursor position (e.g. https://unix.stackexchange.com/a/389266). Plus, who knows how Windows / macOS work.

djc:
Maybe open an issue against console? This seems like the kind of thing that it could provide.

chris-laplante:
> Maybe open an issue against console? This seems like the kind of thing that it could provide.

OK, will do.

chris-laplante:
On further thought, I don't think that console are any better positioned to fix this than we are, so I'm going to push on and attempt to make a fix.

chris-laplante:
OK, here's my latest attempt at fixing #150. Still need to write tests for it.

djc:
Sorry, it's been a very busy week, but this is definitely still on my list.

src/progress_bar.rs
```diff
@@ -517,6 +515,26 @@ impl ProgressBar {
         self.state().draw_target.remote().map(|(_, idx)| idx)
     }
 
+    /// Current message
```
djc, line 529: I don't see a good reason to add all this public API?

src/state.rs
```diff
@@ -7,11 +7,14 @@ use std::{fmt, io};
 use crate::draw_target::ProgressDrawTarget;
 use crate::style::ProgressStyle;
 
+pub(crate) const DEFAULT_TAB_WIDTH: usize = 8;
```
djc, line 10: This should go near the bottom of the module.

src/state.rs
```diff
@@ -93,15 +99,42 @@ impl BarState {
     }
 
     pub(crate) fn set_message(&mut self, now: Instant, msg: Cow<'static, str>) {
-        self.state.message = msg;
+        self.state.message = TabExpandedString::new(msg, self.tab_width);
         self.update_estimate_and_draw(now);
     }
 
+    // Called in builder context
+    pub(crate) fn set_message_without_draw(&mut self, msg: Cow<'static, str>) {
```
djc, line 107: I don't see much point in having this setter, we could just make the necessary fields `pub(crate)`?

src/state.rs
```diff
@@ -93,15 +99,42 @@ impl BarState {
...
+
     pub(crate) fn set_prefix(&mut self, now: Instant, prefix: Cow<'static, str>) {
-        self.state.prefix = prefix;
+        self.state.prefix = TabExpandedString::new(prefix, self.tab_width);
+        self.update_estimate_and_draw(now);
+    }
+
+    // Called in builder context
+    pub(crate) fn set_prefix_without_draw(&mut self, prefix: Cow<'static, str>) {
+        self.state.prefix = TabExpandedString::new(prefix, self.tab_width);
```
djc, line 118: Same here.

src/state.rs
```diff
@@ -93,15 +99,42 @@ impl BarState {
...
+    }
+
+    pub(crate) fn set_tab_width(&mut self, now: Instant, tab_width: usize) {
```
djc, line 121: I don't think we need both with and without draw variants, let's only do the `without_draw()` variant and drop the suffix.

src/state.rs
```diff
@@ -287,6 +320,71 @@ impl ProgressState {
     pub fn set_len(&mut self, len: u64) {
         self.len = Some(len);
     }
+
+    pub fn set_message(&mut self, msg: TabExpandedString) {
```
djc, line 324: Why do we need this public API?

src/state.rs
```diff
@@ -287,6 +320,71 @@ impl ProgressState {
...
+        self.prefix = prefix;
+    }
+
+    pub fn prefix(&self) -> &str {
+        self.prefix.expanded()
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum TabExpandedString {
```
djc, line 342: I don't think this should be exposed in the public API.

src/style.rs
```diff
@@ -232,7 +238,7 @@ impl ProgressStyle {
                 } => {
                     buf.clear();
                     if let Some(formatter) = self.format_map.get(key.as_str()) {
-                        buf.push_str(&formatter(state));
+                        buf.push_str(&formatter(state).replace('\t', &" ".repeat(self.tab_width)));
```
djc, line 241: In #420 this is changed to pass in a `Formatter` directly. How do we change tab expansion in that case?

src/progress_bar.rs
chris-laplante: I added these because I need the ability to re-style the prefix and message at runtime. I can make them a separate PR if you'd like.

src/state.rs
chris-laplante: I made them setters because they have invariants they need to maintain. Namely, when setting the message and prefix, we need to keep the tab widths of everything in sync.

src/progress_bar.rs
djc: Well, I think the `unstyled()` getters are very niche and you should handle that in your project. For `message()` and `prefix()`, can we just return a `&str`?

src/state.rs
djc: That doesn't seem like a big issue because `self.state.prefix` is a `TabExpandedString`, so the type system will force the caller to use the correct type. Maybe `ProgressState` should have an `expand(&self, &str) -> TabExpandedString` method that callers can use?

src/progress_bar.rs
chris-laplante: Sure, I'll remove the `unstyled()` getters. We can't return a `&str` because it's behind a lock.

djc: Ah yes... too bad std doesn't have guard mapping like Tokio.

src/style.rs
```diff
@@ -75,6 +75,11 @@ impl ProgressStyle {
         Ok(Self::new(Template::from_str(template)?))
     }
 
+    pub(crate) fn change_tab_width(&mut self, new_tab_width: usize) {
```
djc, line 78: I think it'd be easier if this was just named `set_tab_width()` as well, to match the public API.

djc (approved):
Approved with the remaining comments addressed. This has turned out quite nice!

src/progress_bar.rs
```diff
@@ -517,6 +515,26 @@ impl ProgressBar {
...
+    pub fn message(&self) -> Cow<'static, str> {
+        self.state().state.message.clone()
+    }
+
+    /// Current message (with ANSI escape codes stripped)
+    pub fn message_unstyled(&self) -> String {
```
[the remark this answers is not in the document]
chris-laplante: Ah my mistake, thought I got rid of them...

chris-laplante:
OK, just going to wait for the tests to complete then I'll merge :)

PR #424, opened 2022-04-27
MultiProgress:insert_{after, before}: grab pb index immediately to avoid deadlock with Ticker

chris-laplante:
Before this change it is possible for a combination of Ticker and
MultiProgress to deadlock due a lock inversion. Consider this code:

    let mp = MultiProgress::new();
    let s = mp.add(ProgressBar::new_spinner());
    s.enable_steady_tick(Duration::from_millis(50));

    let p0 = mp.insert_after(&s, ProgressBar::new(10));

Inside `Ticker`, a lock on `BarState` is acquired and then the state is
drawn to the draw target via `BarState::draw()`. `draw` calls
`self.draw_target.width()` which in the case of `MultiProgress` locks
`MultiState`.

Meanwhile, the call to `insert_after` calls `MultiProgress::internalize`
which locks the `MultiState`. Then `MultiState::insert` (in the case of
`insert_after` and `insert_before`) locks the `BarState` in order to
read the index.

So we have thread 1 locking in this order: `BarState`, `MultiState`
and thread 2 locking in this order:        `MultiState`, `BarState`
== boom

The fix is to change `insert_after` and `insert_before` to immediately
grab the index of the progress bar rather than deferring it to be done
when the `MultiState` is already locked. This makes the locking order of
thread 2 the same as that of 1.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -94,7 +94,7 @@ impl MultiProgress {
     /// remote draw target that is intercepted by the multi progress
     /// object overriding custom `ProgressDrawTarget` settings.
     pub fn insert_before(&self, before: &ProgressBar, pb: ProgressBar) -> ProgressBar {
-        self.internalize(InsertLocation::Before(before), pb)
+        self.internalize(InsertLocation::Before(before.index().unwrap()), pb)
     }
 
     /// Inserts a progress bar after an existing one.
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Are there any structural/type-level changes we could put in place to make this harder to hit? Is it possible to write a test for this? Is this a regression from 0.16?

chris-laplante:
> Are there any structural/type-level changes we could put in place to make this harder to regress? Is it possible to write a test for this? Is this a regression from 0.16?

I think I would be able to come up with a test using our friend the `Barrier` :), though I'm not sure how clean the code would look. 

It may be a regression, I'm not sure - the code for handling draw states was changed after I added the `insert_before`/`insert_after`. 

Not sure about structural changes. I think as long as multiple mutexes are in the play the danger is always there. I'm going to play with using [`ThreadSanitizer`](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/sanitizer.html#threadsanitizer) to see if it could have caught this, though.

PR #425, opened 2022-04-27
make set_message/set_prefix accept anything that impls Display

chris-laplante:
Among other things, by broadening the allowed types it lets you pass a `console::StyledObject` which is convenient.


src/progress_bar.rs
```diff
@@ -4,6 +4,7 @@ use std::sync::MutexGuard;
 use std::sync::{Arc, Mutex, Weak};
 use std::time::{Duration, Instant};
 use std::{fmt, mem};
+use std::fmt::Display;
```
djc, line 7: Should import fmt::self here, too, and remove it from the other place.

djc (approved):
Nice!

src/progress_bar.rs
chris-laplante: Will do

chris-laplante:
I also added accessors to `ProgressBar` to get the current message/prefix.

djc:
I think the accessors should return &str.

chris-laplante:
> I think the accessors should return &str.

Ah, good call. I'll change that.

djc:
So the downside here is that it's no longer possible to pass a `&static str` without it allocating a `String` for you? Not sure that's a good trade-off... You could just call `to_string()` in the caller, right?

chris-laplante:
> So the downside here is that it's no longer possible to pass a `&static str` without it allocating a `String` for you? Not sure that's a good trade-off... You could just call `to_string()` in the caller, right?

Yes, that is true :(. An alternative would be to introduce a trait that we `impl` on both `Cow<'static, str>` and `console::StyledObject`. (I did initially try to go a similar route, where I attempted to `impl` the trait for anything that is `impl Into<Cow<'static, str>>`. but I'm not sure Rust lets you express that.)

djc:
Hmm, I feel like adding yet another trait isn't really worth it for this. I think `impl Into<Cow<'static, str>>` is somewhat of a local optimum, as you can trivially call `to_string()` on any `impl Display` that you've got and pass that.

PR #427, opened 2022-05-05
update rayon dependency

bmc-msft:


djc:
Why is this necessary?

bmc-msft:
If users enable the rayon feature, a sub-dependency of a crossbeam-deque specifies a version which has a CVE (CVE-2021-32810) associated with it.  

From my brief read, I don't believe using the rayon feature in indicatif would trigger the race condition discussed in the CVE.  However, using a handful of different dependency scanners alert on this.

djc:
Which scanners? Do they still trigger after you've run `cargo update`?

djc:
Going to close this for now, please reopen if you still think this is an issue.

PR #429, opened 2022-05-11
Remove rustfmt config again

djc:
This contains unstable options, which results in warnings when running
stable rustfmt. That doesn't seem like a good experience to me.

@chris-laplante what do you think? IMO this is a little too ugly (for outside contributors, too) to keep like this. Alternatives:

* We can enforce nightly rustfmt in CI, but sucks if local rustfmt is different from CI
* We could ignore rustfmt.toml, so you can keep it around and I'll review manually?

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,2 +1,0 @@
-imports_granularity = "Module"
-group_imports = "StdExternalCrate"


chris-laplante:
Ah, darn. I think option 2 might be easiest. I can just leave the rustfmt.toml in my home directory and it will be picked up according to the docs.

PR #430, opened 2022-05-12
fix #341: make is_hidden() work for MultiProgress

### LINKED ISSUE #341: is_hidden always returns true when using MultiProgress
I wrote a custom `Log` implementation which writes log messages using `ProgressBar::println`.

As expected, all my log messages got swallowed up when stderr was redirected to a file. Per the docs for `ProgressBar::println`, this is expected because the draw target is considered hidden (not a TTY).

To work around, I tried to call `ProgressBar::is_hidden` from within my logger. If the method returns true, I bypass `ProgressBar::println` and log directly to stderr with `eprintln!`.

Unfortunately, this technique does not work when using `MultiProgress`.

I believe the culprit is here:

https://github.com/mitsuhiko/indicatif/blob/92ce4c3f9903dbcb30a087d213ca54d041f462d6/src/draw_target.rs#L125-L131

When `self.kind` is `Remote`, we simply return false. Would it make sense to recursively call `is_hidden` on the remote target instead?

chris-laplante:


djc (approved):

djc:
Nice!

PR #434, opened 2022-05-19
Don't conflate ticking with updating estimates 

### LINKED ISSUE #433: Spinners no longer advance
It appears that commit 6697ac3472010e54589c14a485512da543b0c923 has broken spinners. After this commit spinners no longer spin.

Run `cargo run --example yarnish` before and at this commit to see the difference.

chris-laplante:
Fixes #433


src/state.rs
```diff
@@ -77,36 +77,37 @@ impl BarState {
 
     pub(crate) fn update(&mut self, now: Instant, f: impl FnOnce(&mut ProgressState)) {
         f(&mut self.state);
-        self.tick(now);
```
djc, line 80: This seems like a big change in behavior? Why do this?

tests/render.rs
```diff
@@ -271,3 +271,25 @@ fn ticker_drop() {
         "  doing stuff 0\n  doing stuff 1\n  doing stuff 2\n  doing stuff 3\n  doing stuff 4"
     );
 }
+
+#[test]
+fn manually_inc_ticker() {
```
djc, line 276: Yay tests!

src/state.rs
chris-laplante: My thinking is that the whole reason the user is calling `update` is because they want to do something specific with `ProgressState` that isn't already covered by the existing API. I can't imagine they'd want to do something very custom + tick. Though I agree it is a change in behavior, so I could back out this change if you want.

tests/render.rs
chris-laplante: :D

src/state.rs
djc: It doesn't seem like a big deal to me to tick in this particular case? It's not like ticking is expensive, we currently tick for most other kinds of updates IIRC.

chris-laplante: Sounds good, I'll back that change out after dinner tonight.

djc: So I was expecting that you'd change everything back to tick, like before, but you only backed this particular change out. I'm still not sure that makes sense?

djc (approved):

fabricedesre:
It would be nice to get a new published version with this fix. When is the next rc planned?

djc:
We don't really plan rcs, we can just do one now.

fabricedesre:
> @fabricedesre out of curiosity, what are you using indicatif for?

I use it in this command line tool: https://github.com/capyloon/nutria/tree/main/builder to display the download progress of prebuilt binaries. That can be slow when the ipfs gateway is not in great shape, so I'd like to use `[enable_steady_tick](https://docs.rs/indicatif/0.16.2/indicatif/struct.ProgressBar.html#method.enable_steady_tick)` but before this fix it was no t updating remaining time properly when the download was slow / waiting for data.

djc:
#435 has the version bump, will publish once that's merged.

MarijnS95:
`fn tick()` in the documentation still mentions `This automatically happens on any other change to a progress bar.`:

https://github.com/console-rs/indicatif/blob/496a6059839c7e50a17099c4828dd4cb24aeb1fc/src/progress_bar.rs#L162-L163

PR #438, opened 2022-05-26
MultiProgress: prune 'zombie' progress bars; fixes #426; add render tests

### LINKED ISSUE #426: MultiProgress: should prune leading progress bars that have become `DoneVisible` or `DoneHidden`
Consider this small example program:

```rust
use std::thread;
use std::time::Duration;

use indicatif::{MultiProgress, ProgressBar, ProgressFinish};

fn main() {
    let m = MultiProgress::new();
    let height = console::Term::stdout().size().0;

    for i in 0..(height + 1) {
        let p1 = m.add(
            ProgressBar::new_spinner()
                .with_message(format!("p{}", i))
                .with_finish(ProgressFinish::AndLeave),
        );
        p1.enable_steady_tick(Duration::from_millis(50));
        thread::sleep(Duration::from_millis(200));
    }
}
```

It creates one more spinner that there are lines in the terminal. Here is a video of it running:

https://user-images.githubusercontent.com/40474653/166507072-efdbaf25-82e5-4f25-91ac-e6ab51db2314.mp4

Since we never call `MultiProgress::remove` on `ProgressBar`s that are dropped/finished (see #419), bars are always redrawn even when they don't need to be. When the number of bars (_n_) becomes greater than terminal lines (_m_), on each draw `MultiProgress` will clear the last _m_ lines but write _n_ lines, leaving _m - n_ lines in the scrollback history.

I propose the following algorithm. On each `MultiProgress:draw`, iterate over the bars in order (according to `MultiState::ordering`) and `take_while(|bar| bar.is_finished())`. This set of leading progress bars is then pruned by calling `MultiProgress:remove` on each one. This ensures we are not re-drawing bars that don't need to be redrawn and that we are some reclaiming memory. 

The wrinkle is that `MultiState` cannot currently get at the `ProgressState` to check `is_finished()`, so we will need to restructure some code.

chris-laplante:


chris-laplante:
@djc FYI this draft PR so far just introduces a test that demonstrates issue #426. I am working on implementing the fix.

src/multi.rs
```diff
@@ -169,8 +174,8 @@ impl MultiProgress {
 pub(crate) struct MultiState {
     /// The collection of states corresponding to progress bars
     /// the state is None for bars that have not yet been drawn or have been removed
-    draw_states: Vec<Option<DrawState>>,
-    /// Set of removed bars, should have corresponding `None` elements in the `draw_states` vector
+    elements: Vec<MultiStateElement>,
```
djc, line 177: Hmm, not so sure about the `element`/`Element` name. Maybe `members` or `bars`?

Also if there are a bunch of diff hunks that only change this name, might be a good idea to pull that change out into a separate commit.

src/multi.rs
```diff
@@ -169,8 +174,8 @@ impl MultiProgress {
...
+    /// Set of removed bars, should have corresponding `None` elements in the `elements` vector
```
djc, line 178: There aren't really `None` `elements`, maybe refine the wording here?

src/multi.rs
```diff
@@ -203,6 +208,32 @@ impl MultiState {
         extra_lines: Option<Vec<String>>,
         now: Instant,
     ) -> io::Result<()> {
+        // Reap all consecutive 'zombie' progress bars from head of elements
+        let mut zombies = vec![];
```
djc, line 212: So this will allocate any time there are zombies. Should we store it in the `MultiProgressState` so we don't have to reallocate on every draw with zombies, or do we think draws with zombies are relatively rare?

src/multi.rs
```diff
@@ -203,6 +208,32 @@ impl MultiState {
...
+        let mut adjust = 0;
+        for index in self.ordering.iter() {
+            let element = &self.elements[*index];
+            if element.is_zombie {
```
djc, line 216: Can we change this to `if !element.is_zombie { break; }` to reduce rightward drift?

src/multi.rs
```diff
@@ -340,7 +371,23 @@ impl MultiState {
     }
 
     fn len(&self) -> usize {
-        self.draw_states.len() - self.free_set.len()
+        self.elements.len() - self.free_set.len()
+    }
+}
+
+#[derive(Default)]
+struct MultiStateElement {
```
djc, line 379: So with `draw_state: Option<_>` and `is_zombie: bool` we have 4 different states. Do we need all 4? Maybe that would be a useful thing to explain in a comment?

I was surprised `Weak` implements `Default` -- it clearly works but it feels a little ugly to me. If we can collapse some of the states maybe there's something better we can do?

djc (commented):
This looks pretty sensible, commented on some possible improvements.

src/multi.rs
chris-laplante: Well, only three states are necessary:

1. On `MultiProgress::internalize` (and before draw) `draw_state` will be `None` and `is_zombie` will be `false`.
2. On draw, `draw_state` will be `Some(..)` and `is_zombie` will be `false`.
3. When a zombie is detected, `is_zombie` will become `true`. Normally `draw_state` will be `Some(..)` but I suppose it's possible for a pb to be briefly in zombie state if it was never drawn (i.e. you immediately drop it after adding to the `MultiProgress`). 

I'll add a comment to this effect.

src/multi.rs
chris-laplante: Depends on the usage pattern. If users are creating and dropping pbs a lot, then it's going to allocate a lot. Seems like a premature optimization to stuff it into `MultiState` for now - would you be ok with us waiting for someone to complain?

src/multi.rs
chris-laplante: One option would be to have an enum, but I think that would make the code even uglier.

chris-laplante:
OK, this should be a little better (hopefully)

src/multi.rs
djc: By "suppose it's possible", I think you're suggesting that we actually do have 4 states?

djc:
Would you mind squashing the review feedback commit back into the originating commit? Would make reviewing easier.

src/multi.rs
chris-laplante: Yes, there are technically four states. But state 3 and 4 (with 4 being "is_zombie == true" and "draw_state is None") are handled the same.

PR #441, opened 2022-06-06
fix: first style of wide elements was ignored

### LINKED ISSUE #431: Examples don't have the main bar colored anymore
It used to be that the progress bar template style supported coloring, but this behavior seems to have broken by commit 5038ba8019bad296b5a0314770db5d05c050a7dd according to git bisect.

Before that commit, `cargo run --release --example download` would print a progress bar where the hashes (`#`) were colored in blue. Starting at that commit, they are printed as white. I don't think this is supposed to happen.

LussacZheng:
Fix #431.

I did not delve into the inner code, so I'm not sure whether it's possible to simply remove this code without causing other bugs.
I ran `cargo test` locally and all tests passed. `cargo run --example download` now works fine too.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -321,12 +321,6 @@ impl ProgressStyle {
                         }
                     };
 
-                    if buf == "\x00" {
-                        // Don't expand for wide elements
-                        cur.push_str(&buf);
-                        continue;
-                    }
-
                     match width {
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks for working on this, LGTM!

Issue #442, opened 2022-06-10
Support for multi-line progress messages


happenslol:
First of all, thanks for the awesome library! It's been a joy to use so far.

Recently, I've had a few usecases where my CLI apps are running bash scripts or something similar in the background. Of course, I could just show an indeterminate progress spinner here to show that a script is running. However, after using buildkit, which basically shows the last few lines of output for each step being run, and clears them when the step is done, I thought about how to achieve this with this library.

The simplest way to show recent output for progress bars would be to just have multiline progress bars. I could then take e.g. the last three lines of the script output and display them below the progress message. I tried this but newlines are not handled at all by indicatif, resulting in duplicated lines whenever a newline is printed.

I would be open to working on this if this functionality is something you would consider adding in this library. If you need more details or examples of this behavior in other CLI tools, I can provide that.

djc:
Which version did you try? It might be helpful to try the latest 0.17 rc. I'm fairly certain there is at least some newline handling in indicatif, although I haven't really tried it myself. I would definitely be open to have something like this in the library, happy to answer questions about the code base if you have any.

happenslol:
I actually checked out the latest main branch, and tried inserting newlines into a bunch of the examples, which just resulted in a whole lot of output :-P I suppose it's possible that it works for a single progress bar though, I'm not sure I tried that.

I'll have a closer look at the codebase and see if I can whip something up. Thanks!

happenslol:
I've tried to implement this, see #443. I'd be happy about any feedback!

djc:
This was merged, going to close this.

PR #443, opened 2022-06-13
Implement multi-line progress message support

happenslol:
This PR adds support for multi-line progress messages, following #442.

The changes mostly update the drawing functions to split progress messages before counting them, so that the correct number of lines are cleared when they are redrawn. I'm not sure I 100% understood how orphan lines work, so I'd appreciate if someone could take a closer look at how I handle them.

The existing examples all run as they did previously, and I've added a new example (`multi-log`) showing the style I was trying to implement when I opened the issue.

One thing I noticed is that calling `clear` on a multiline progress doesn't always seem to clear all progress messages, a random amount stays on the screen even when a message is printed afterwards. I first thought that I introduced that bug, but switching back to main it still appeared in the examples I tried (`yarnish` and `multi`). Is this a known bug?

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -359,23 +359,30 @@ impl ProgressStyle {
                     }
                 }
                 TemplatePart::Literal(s) => cur.push_str(s.expanded()),
-                TemplatePart::NewLine => lines.push(match wide {
-                    Some(inner) => {
-                        inner.expand(mem::take(&mut cur), self, state, &mut buf, target_width)
-                    }
-                    None => mem::take(&mut cur),
-                }),
+                TemplatePart::NewLine => {
+                    self.push_line(lines, &mut cur, state, &mut buf, target_width, &wide)
+                }
             }
         }
 
         if !cur.is_empty() {
-            lines.push(match wide {
-                Some(inner) => {
-                    inner.expand(mem::take(&mut cur), self, state, &mut buf, target_width)
-                }
-                None => mem::take(&mut cur),
-            })
+            self.push_line(lines, &mut cur, state, &mut buf, target_width, &wide);
         }
+    }
+
+    fn push_line(
+        &self,
+        lines: &mut Vec<String>,
+        cur: &mut String,
+        state: &ProgressState,
+        buf: &mut String,
+        target_width: u16,
+        wide: &Option<WideElement>,
+    ) {
+        lines.push(match wide {
+            Some(inner) => inner.expand(mem::take(cur), self, state, buf, target_width),
+            None => mem::take(cur),
+        });
     }
 }
 
--- a/src/style.rs
+++ b/src/style.rs
@@ -379,10 +379,23 @@ impl ProgressStyle {
         target_width: u16,
         wide: &Option<WideElement>,
     ) {
-        lines.push(match wide {
+        let expanded = match wide {
             Some(inner) => inner.expand(mem::take(cur), self, state, buf, target_width),
             None => mem::take(cur),
-        });
+        };
+
+        // If there are newlines, we need to split them up
+        // and add the lines separately so that they're counted
+        // correctly on re-render.
+        for (i, line) in expanded.split('\n').enumerate() {
+            // No newlines found in this case
+            if i == 0 && line.len() == expanded.len() {
+                lines.push(expanded);
+                break;
+            }
+
+            lines.push(line.to_string());
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


happenslol:
> If lines does not contain one element per line, that seems like a problem to me. Should we split up multi-line messages earlier in the process to deal with this?

Could you provide an example of this happening? It would probably be possible to render completely on a line-by-line basis, since the render target is not really aware of the content of the lines and just redraws everything every time. I'm not sure I understand the issue you're referring to here though.

> I'm honestly not sure I 100% understand them, myself. chris-laplante do you? I think they're used to keep track of how many lines we drew in the previous iteration, so that we can correct if we're drawing fewer lines in the current iteration.

From what I could glean from the code, it's about cleaning up old lines when the progress is dropped, as well as when lines are printed manually by the user and lines that were moved up as a result need to be forgotten. So I defaulted to counting multiline strings in there as well, since they could potentially contain multiline messages set by the user. Not sure if this is necessary though.

> Would you mind filing a separate issue about this?

Sure, will do!

chris-laplante:
The code is a little hard to follow, but what I understand is that `orphan_lines` are only used to implement `println`. I'll try to go through and add some comments and/or clean up the code to make it easier to understand.

happenslol:
Cool, feel free to make changes or add suggestions on my PR.

chris-laplante:
> Cool, feel free to make changes or add suggestions on my PR.

Will do, but BTW I meant that the code in indicatif currently is a little hard to follow, not your code :)

happenslol:
Haha, don't worry about it, that's not how I took it. I didn't add enough code for it to be too complicated anyways :-P

happenslol:
Is there anything I can help with to move this along? I'm depending on my fork with my own projects for now, but of course I'd prefer if I can go back to upstream. I don't mind this taking longer, but I'd be happy to help if I can speed up the process :-)

djc:
I'm still wondering if there's a good reason not to have one element per line in `DrawState::lines`, rather than potentially multiple lines per element due to nested `\n`. IMO we should deal with that in a different way, by splitting up the message in `set_message()`.

happenslol:
Just had a more thorough look at this. From what I can see, the different parts of the status (msg, prefix and so on) are set separately, and then they are appended to `DrawState::lines` in `ProgressStyle::format_state`. This means that if we want to already split multiline messages in `set_message`, we would need to have `Vec`s for all of the parts of the status, since both prefix and message (and maybe even the bar?) might have multiple lines.

This already introduces overhead (`Vec` instead of `String` for every part that can contain multiple lines), but also gets very complicated since the last line of each part needs to be joined together with the first line of the next part. Also, having newlines in the template is not even possible like this.

I think my approach is not optimal, but it makes the rendering of newlines in any part of the progress state a part of the rendering process, which allows any amount of newlines in any part of the template or message.

Maybe I misunderstood you and I'm missing an easier solution here... What do you think?

djc:
Hmm, maybe it would be better to split stuff inside `ProgressStyle::format_state()`?

happenslol:
Alright, that turned out to be a great idea! I'm now splitting the buffer right before any new lines are appended, and running the same code as before if no newlines are contained. This should handle all cases, even when the newlines are split over different parts of the template.

This also enabled me to add some tests for the cases I added. Have a look and tell me if there's anything else I can add.

src/style.rs
```diff
@@ -349,22 +349,43 @@ impl ProgressStyle {
...
+        wide: &Option<WideElement>,
+    ) {
+        let expanded = match wide {
+            Some(inner) => inner.expand(mem::take(cur), self, state, buf, target_width),
+            None => mem::take(cur),
+        };
+
+        // If cur doesn't contain any newlines now, we can add
+        // the line as is.
+        if !expanded.contains('\n') {
```
djc, line 379: This is somewhat less efficient than it could be. I think what we want here is to skip this extra `if` and instead, in the loop below, add `enumerate()` and check on `i == 0 && line.len() == expanded().len()` to avoid the allocation in that case.

examples/multi-log.rs
```diff
@@ -0,0 +1,81 @@
+use std::thread;
```
djc, line 1: I'd prefer not to add a separate example for this.

djc (commented):
This is looking pretty good! I'd like it if you squashed all your changes into two commits: one that extracts the `push_line()` method and one that adds the new functionality and the test. (I usually squash using `git rebase -i`.)

happenslol:
Alright, all done. Please let me know if there's anything else I can do.

djc (approved):
This looks good to me! I think there is some potential for optimization, but this will do as a first take. @chris-laplante any thoughts?

chris-laplante:
> This looks good to me! I think there is some potential for optimization, but this will do as a first take. @chris-laplante any thoughts?

It looks good to me too and some light testing on my own code gives correct results :). The only thing I'm wondering is whether we've tested what happens when you change the number of newlines in messages using `ProgressBar::set_message`. Especially when part of `MultiProgress`. I want to make sure it doesn't under/over-erase lines.

djc:
You could submit some tests for the case Chris mentioned in his last comment, I think?

happenslol:
Alright, I've added additional tests that render a `MultiProgress`, set multiline messages and then single line messages again using `set_message`, increase the progress and then finish the progress using `finish_with_message`.

Please tell me if these tests cover the cases you were thinking about.

chris-laplante:
Thanks for adding the tests! This looks good to me. @djc alright to merge?

djc (approved):

djc:
Great, thanks!

PR #444, opened 2022-06-16
leverage GenerationalTokenList to simplify multi code

chris-laplante:
I _did_ have to enable the use of 'unsafe' code in `GenerationalTokenList` in order to use `iter_mut`. Nothing is reported by `MIRIFLAGS=-Zmiri-disable-isolation cargo miri test`. 

I'm fairly confident the code is correct. Basically, I think that as long as the `GenerationalTokenList` maintains the linked list properly (no cycles) nothing bad can happen. 

Still, I understand if you are reticent to accept the unsafe code. I thought I'd present what I have here anyway :).

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -2,6 +2,7 @@ use std::fmt::{Debug, Formatter};
 use std::io;
 use std::sync::{Arc, Mutex, RwLock, Weak};
 use std::time::Instant;
+use generational_token_list::{GenerationalTokenList, ItemToken};
 
 use crate::draw_target::{DrawState, DrawStateWrapper, ProgressDrawTarget};
 use crate::progress_bar::ProgressBar;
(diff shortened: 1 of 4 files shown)


chris-laplante:
One additional benefit of this that I forgot to mention is it reduces the number of states that `MultiStateMember` can take on. `MultiStateMember::pb` will always have a valid reference, and also the meaning of `MultiStateMember::draw_state` being `None` is no longer overloaded to mean both never been drawn and in the free set.

src/multi.rs
```diff
@@ -459,6 +426,19 @@ mod tests {
         pb.finish();
     }
 
+    macro_rules! check_order {
```
djc, line 429: Could this go in a separate PR, or does it depend on the generational token list?

djc (commented):
How is generational_token_list different from slotmap or indexmap? Is this a truly novel thing?

src/multi.rs
chris-laplante: It probably could. I'll make it so.

chris-laplante:
> How is generational_token_list different from slotmap or indexmap? Is this a truly novel thing?

It is (mostly) novel, inspired strongly by https://github.com/steveklabnik/indexlist. The main difference between generational_token_list and indexlist being that the latter rolls its own generational arena, whereas in the former I just use https://github.com/fitzgen/generational-arena. 

generational_token_list and indexlist are both doubly-linked lists. Compare that to slotmap where iteration order is arbitrary. indexmap maintains insertion order, but it just deals with indices which are obviously not stable when elements are inserted or removed. generational_token_list, indexlist, and slotmap solve that problem with tokens.

chris-laplante:
Closing this for now, as I would tend to agree that it doesn't add much value currently.

djc:
It's a cool idea! Just for this particular use case I feel like it has limited benefit.

PR #446, opened 2022-06-22
modify zombie handling code so it doesn't need to alloc

chris-laplante:


src/multi.rs
```diff
@@ -209,23 +209,18 @@ impl MultiState {
         now: Instant,
     ) -> io::Result<()> {
         // Reap all consecutive 'zombie' progress bars from head of the list
-        let mut zombies = vec![];
         let mut adjust = 0;
-        for index in self.ordering.iter() {
-            let member = &self.members[*index];
+        while let Some(index) = self.ordering.first().cloned() {
```
djc, line 213: Should use `copied()` here instead of `cloned()`?

djc (commented):
Nice!

djc (approved):

PR #458, opened 2022-08-03
Only tick if the ticker is disabled

### LINKED ISSUE #457: Spinner running faster than "steady tick" duration in v0.17
When upgrading to v0.17, I noticed that the spinner may update *faster* than the rate set in `.enable_steady_tick(…)`.

Background: I use `.enable_steady_tick(…)` to get a consistent look of my progress bar spinner animation, independent from the actual speed of `.inc(…)` calls (which I don't know in advance - it might be faster or slower than the rate set for the steady tick). If I update the bar at a low rate, everything is fine. But if I call `.inc(…)` *faster* than the rate set in `.enable_steady_tick(…)`, I would still expect the spinner to update at the low refresh rate - instead of flickering arbitrarily fast in v0.17. At least this is how it worked in v0.16, and the name of the function (steady tick) seems to indicate that this is the expected outcome, even if the [documentation](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.enable_steady_tick) refers to a use case where the bar is updated at a low speed.

To reproduce this, compare the output of the two programs below, with v0.16 and v0.17 as respective `indicatif` versions:

## v0.16 version (works fine, spinner updates every 200ms)

```rs
use std::thread;
use std::time::Duration;

fn main() {
    let steps = 10000;

    let progressbar_style = ProgressStyle::default_spinner()
        .template(" {spinner} {wide_bar}");

    let bar = ProgressBar::new(steps);
    bar.set_style(progressbar_style);
    bar.enable_steady_tick(200);

    for _ in 0..steps {
        bar.inc(1);
        thread::sleep(Duration::from_millis(1));
    }
}
````

## v0.17 version (spinner updates very fast, presumably every 1ms)
```rs
use std::thread;
use std::time::Duration;

fn main() {
    let steps = 10000;

    let progressbar_style = ProgressStyle::default_spinner()
        .template(" {spinner} {wide_bar}")
        .unwrap();

    let bar = ProgressBar::new(steps);
    bar.set_style(progressbar_style);
    bar.enable_steady_tick(Duration::from_millis(200));

    for _ in 0..steps {
        bar.inc(1);
        thread::sleep(Duration::from_millis(1));
    }
}

```

djc:
Fix a regression from #417 to fix #457.

Of the 4 callers of `BarState::tick()`, changed three to no longer tick if the ticker is enabled:

* `ProgressBar::inc()`
* `ProgressBar::set_position()`
* `ProgressBar::update()`

The fourth is `ProgressBar::tick()`, which already only ticked if the ticker was disabled.

### DIFF
--- a/tests/render.rs
+++ b/tests/render.rs
@@ -339,7 +339,7 @@ fn ticker_drop() {
                 .with_finish(ProgressFinish::AndLeave)
                 .with_message(format!("doing stuff {}", i)),
         );
-        new_spinner.enable_steady_tick(Duration::from_millis(50));
+        new_spinner.enable_steady_tick(Duration::from_millis(100));
         spinner.replace(new_spinner);
     }
 
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -191,9 +191,13 @@ impl ProgressBar {
     ///
     /// This automatically happens on any other change to a progress bar.
     pub fn tick(&self) {
+        self.tick_inner(Instant::now());
+    }
+
+    fn tick_inner(&self, now: Instant) {
         // Only tick if a `Ticker` isn't installed
         if self.ticker.lock().unwrap().is_none() {
-            self.state().tick(Instant::now())
+            self.state().tick(now)
         }
     }
 
@@ -201,8 +205,8 @@ impl ProgressBar {
     pub fn inc(&self, delta: u64) {
         self.pos.inc(delta);
         let now = Instant::now();
-        if self.pos.allow(now) && self.ticker.lock().unwrap().is_none() {
-            self.state().tick(now);
+        if self.pos.allow(now) {
+            self.tick_inner(now);
         }
     }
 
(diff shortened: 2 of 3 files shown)


tests/render.rs
```diff
@@ -339,7 +339,7 @@ fn ticker_drop() {
                 .with_finish(ProgressFinish::AndLeave)
                 .with_message(format!("doing stuff {}", i)),
         );
-        new_spinner.enable_steady_tick(Duration::from_millis(50));
+        new_spinner.enable_steady_tick(Duration::from_millis(100));
```
djc, line 342: There was a test failure on Windows only, so I'm guessing that's just a timing artefact?

djc:
Seems like the performance impact here is negligible (based on running `cargo run --release --example fastbar`), although it's a little ugly that we have to lock another mutex.

tests/render.rs
chris-laplante: Yes, I think so

chris-laplante (approved):

PR #459, opened 2022-08-03
Updated single example to expose refresh rate

jqnatividad:
also shows that the performance overhead is minimal, and if anything, the main reason to change refresh rate is to reduce flicker.

Also fixes previous example as it did not properly show "done", as the default template doesn't use the `{msg}` placeholder.


djc (approved):

jqnatividad:
This is very old, so just closing this.

PR #462, opened 2022-08-10
Some small tweaks I noticed while working on zombie handling

chris-laplante:


src/state.rs
```diff
@@ -161,14 +161,15 @@ impl BarState {
 
     pub(crate) fn draw(&mut self, mut force_draw: bool, now: Instant) -> io::Result<()> {
         let width = self.draw_target.width();
+
+        // `|= self.is_finished()` should not be needed here, but we used to always for draw for
```
djc, line 165: While you're in here, mind correcting the "for draw for" here?

djc (approved):

src/state.rs
chris-laplante: Done

PR #468, opened 2022-08-15
Fix #467

### LINKED ISSUE #467: indicatif::style::ProgressTracker is not `Sync`
Is ```Sync``` bound missing by mistake? After updating my project's dependency from "0.17.0-rc.8" to "0.17.0" it doesn't compile anymore because of this:
```rust
lazy_static! {
    static ref IDLE_STYLE: ProgressStyle = ProgressStyle::with_template("{prefix:.dim} {msg:.cyan.dim}").unwrap();
}
```
```
486 | / lazy_static! {
487 | |     static ref IDLE_STYLE: ProgressStyle = ProgressStyle::with_template("{prefix:.dim} {msg:.cyan.dim}").unwrap();
488 | | }
    | |_^ `(dyn ProgressTracker + 'static)` cannot be shared between threads safely
```
Considering this struct from [`test_stateful_tracker()`](https://github.com/console-rs/indicatif/blob/20a857a4ee42e29664e5fd1259b72a5563add8bd/src/style.rs#L749):
```rust
#[derive(Debug, Clone)]
struct TestTracker(Arc<Mutex<String>>);
```
doesn't this mean it needs to be ```Sync``` anyway? I'm new at this Rust thread safety thing, I'm sorry if I'm missing something obvious.

rctlmk:
Possible fix for #467.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -733,7 +733,9 @@ pub(crate) struct ArcTracker {
 
 impl ArcTracker {
     fn new<S: ProgressTracker + 'static>(tracker: S) -> Self {
-        Self { tracker: Arc::new(Mutex::new(Box::new(tracker))) }
+        Self {
+            tracker: Arc::new(Mutex::new(Box::new(tracker))),
+        }
     }
 
     pub(crate) fn reset(&mut self, state: &ProgressState, now: Instant) {
@@ -745,7 +747,7 @@ impl ArcTracker {
         let mut tracker = self.tracker.lock().expect("mutex was poisoned");
         tracker.tick(state, now);
     }
-    
+
     pub(crate) fn write(&self, state: &ProgressState, w: &mut dyn fmt::Write) {
         let tracker = self.tracker.lock().expect("mutex was poisoned");
         tracker.write(state, w);
@@ -754,7 +756,9 @@ impl ArcTracker {
 
 impl Clone for ArcTracker {
     fn clone(&self) -> Self {
-        Self { tracker: self.tracker.clone() }
+        Self {
+            tracker: self.tracker.clone(),
+        }
     }
 }
 
@@ -773,7 +777,8 @@ mod tests {
         impl ProgressTracker for TestTracker {
             fn tick(&mut self, state: &ProgressState, _: Instant) {
                 self.0.clear();
-                self.0.push_str(format!("{} {}", state.len().unwrap(), state.pos()).as_str());
+                self.0
+                    .push_str(format!("{} {}", state.len().unwrap(), state.pos()).as_str());
             }
 
             fn reset(&mut self, _state: &ProgressState, _: Instant) {
--- a/src/style.rs
+++ b/src/style.rs
@@ -708,6 +708,8 @@ enum Alignment {
 
 /// Trait for defining stateful or stateless formatters
 pub trait ProgressTracker: Send {
+    /// Creates a new instance of the progress tracker
+    fn clone_box(&self) -> Box<dyn ProgressTracker>;
     /// Notifies the progress tracker of a tick event
     fn tick(&mut self, state: &ProgressState, now: Instant);
     /// Notifies the progress tracker of a reset event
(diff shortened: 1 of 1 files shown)


rctlmk:
It is possible to use [`parking_lot::Mutex`](https://amanieu.github.io/parking_lot/parking_lot/struct.Mutex.html) to avoid a poisoning problem. Or it doesn't really matter, considering a poisoned mutex implies that something already went wrong elsewhere.

rctlmk:
I thought of implementing `Deref` and `DerefMut` for `ArcTracker`, but then decided against it. Unfortunately, I forgot to remove those before restoring some stuff (which I removed without thinking too much), hence the force-push.

djc:
Uh, I thought we'd just add a `Sync` bound to the `ProgressTracker` trait. Why can't we do that instead?

rctlmk:
I forgot to mention it. I'm basing this off [`test_stateful_tracker()`](https://github.com/console-rs/indicatif/blob/217291afa4e4da84c68629311adc3b8c6175493d/src/style.rs#L752). The test fails if we change `TestTracker`'s declaration to `struct TestTracker(String)`. I assumed that we are always supposed to share mutable references, and wrapping all `ProgressTracker` objects into Arc/Mutex is one way to do that.

djc:
Well, we should wrap `TestTracker` in an `Arc<Mutex<_>>` then (or its inner `String`, I guess?).

rctlmk:
I don't know. Is it supposed to work like it does in that test? Or is the test written like that because the inner `String` of the  wrapped in Arc/Mutex? Cloning a `ProgressBar` essentially gets us the same bar,

djc:
Not sure what you mean exactly. The point here is that I don't want to impose `Arc` (much less `Arc<Mutex>`) on the implementer of the trait, so instead we just require the bound for the implementer, and then we have to fix our test type to make sure it can match the constraint.

rctlmk:
I'm sorry if I'm not making sense, I probably don't get how it should to work. Let's suppose there is this `ProgressTracker` implementation:
```Rust
#[derive(Debug, Clone)]
struct TestTracker(String);

impl ProgressTracker for TestTracker {
    fn clone_box(&self) -> Box<dyn ProgressTracker> {
        Box::new(self.clone())
    }

    fn tick(&mut self, state: &ProgressState, _: Instant) {
        self.0.clear();
        self.0.push_str(format!("{} {}", state.len().unwrap(), state.pos()).as_str());
    }

    fn reset(&mut self, _state: &ProgressState, _: Instant) {
        self.0.clear();
    }

    fn write(&self, _state: &ProgressState, w: &mut dyn fmt::Write) {
        w.write_str(self.0.as_str()).unwrap()
    }
}
```

And, like you said, Arc/Mutex is not imposed on the implementor right now. If we run this:
```Rust
use crate::ProgressBar;

let pb = ProgressBar::new(1);
pb.set_style(
    ProgressStyle::with_template("{{ {foo} }}")
        .unwrap()
        .with_key("foo", TestTracker(String::default()))
        .progress_chars("#>-"),
);

let mut buf = Vec::new();
let style = pb.clone().style();

style.format_state(&pb.state().state, &mut buf, 16);
assert_eq!(&buf[0], "{  }");
buf.clear();
pb.inc(1);
style.format_state(&pb.state().state, &mut buf, 16);
assert_eq!(&buf[0], "{ 1 1 }");
pb.reset();
buf.clear();
style.format_state(&pb.state().state, &mut buf, 16);
assert_eq!(&buf[0], "{  }");
pb.finish_and_clear();
```
The first `assert_eq!` is going to pass, but the second one is going to fail. Does it fail because it is expected to do so, or any `ProgressTracker` should be able to pass it, regardless of its implementation?

rctlmk:
> Does it fail because it is expected to do so, or any `ProgressTracker` should be able to pass it, regardless of its implementation?

[Probably the former](https://github.com/console-rs/indicatif/blob/217291afa4e4da84c68629311adc3b8c6175493d/src/progress_bar.rs#L61):
```Rust
/// Get a clone of the current progress bar style.
pub fn style(self) -> ProgressStyle {
    self.state().style.clone()
}
```
Well, then. If the `!Sync` trackers are still a concern, then instead of a `Sync` bound for `ProgressTracker` we can implement `Sync` for `ProgressStyle` manually.

chris-laplante:
We can close this then, right?

PR #469, opened 2022-08-16
Fix the wrong link about `template`

### LINKED ISSUE #465: Where is the list of template keys
The doc tells me that I should review [the list of template keys](https://docs.rs/indicatif/0.17.0/indicatif/style/index.html#templates), but I got nothing.

Jedsek:
Fix #465


djc (approved):

djc:
Thanks!

PR #473, opened 2022-08-26
Correct mis-naming of variables in wrap_async_read

octylFractal:
This was slightly confusing when I saw it in IDE.


djc (approved):
Thanks!

PR #478, opened 2022-09-03
Remove `write_all` from `ProgressBarIter`'s `Write` trait

hexagonal-sun:
By implementing `write_all` we lose granularity of progress updates
that are captured from `write`.  Therefore, remove this function.  This also brings the `ProgressBarIter`'s `Write` trait into line with the `Read` trait since that doesn't provide an implementation of `read_all`.


djc (approved):
Makes sense, thanks!

PR #483, opened 2022-09-23
Reset estimator of progress rate on backwards movement

### LINKED ISSUE #480: ProgressBarIter and finding stream length via seeking breaks rate estimation
MCVE:

```rust
use std::io::{Read, Seek, SeekFrom};
use std::io::Cursor;

use std::thread::sleep;
use std::time::Duration;

pub fn seek_len(seekable: &mut dyn Seek) -> u64 {
    let old_pos = seekable.stream_position().unwrap();
    let len = seekable.seek(SeekFrom::End(0)).unwrap();
    if old_pos != len {
        seekable.seek(SeekFrom::Start(old_pos)).unwrap();
    }
    // return
    len
}

fn main() {
    let mut dummy_buf: [u8; 4] = [0x00; 4];
    let reader = Cursor::new([0xff; 4*100]);

    let pb_style = ProgressStyle::default_bar()
            .template("{wide_bar} | {pos}/{len} | {per_sec}")
            .unwrap();

    let pb = ProgressBar::new(100*4);
    pb.set_style(pb_style);
    let mut pb_reader = pb.wrap_read(reader);
    seek_len(&mut pb_reader);
    for _i in 0..100 {
        pb_reader.read(&mut dummy_buf).unwrap();
        sleep(Duration::from_millis(40));
    }
    pb.finish();
}
```

Expected behavior: rate hovers slightly below 100/s

Actual behavior: rate is stuck at an absurdly high constant while progress bar counts upwards and drops to slightly below 100/s when the progress bar completes

This behavior could occur when a `ProgressBarIter` is passed to a generic function taking a `Read + Seek` object.

This is a regression from v0.16.

rlee287:
Implements a heuristic for ensuring that length finding via seeks does not disrupt the progress estimator.

Closes #480 .

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -385,6 +385,11 @@ impl Estimator {
     fn record(&mut self, new: u64, now: Instant) {
         let delta = new.saturating_sub(self.prev.0);
         if delta == 0 || now < self.prev.1 {
+            // Reset on backwards seek to prevent breakage from seeking to the end for length determination
+            // See https://github.com/console-rs/indicatif/issues/480
+            if new < self.prev.0 {
+                self.reset(now);
+            }
             return;
         }
 


djc:
Sorry, the test failure is unrelated (see #482).

djc (approved):

PR #491, opened 2022-11-05
Fix percent initial value when there is no length

### LINKED ISSUE #489: `percent` in ProgressStyle template starts from 100% instead of 0%
Hi and thank you for maintaining this awesome crate!

After upgrading from `0.16.2` to `0.17.x` I noticed that my spinner `percent` was always starting from 100% instead of 0%.

[indicatif-percent-problems.webm](https://user-images.githubusercontent.com/48568026/197635571-3c025fac-5ffe-415a-b108-db1e7ea25d67.webm)

(I'm not sure if this is a related problem, but also the `eta` stopped working and always displays `0`).

I created a simple repo to reproduce the issue: https://github.com/devmatteini/indicatif-percent-problems

Let me know if I can help somehow :smile: 

---

OS: Ubuntu 22.04
Rust: 1.63.0

devmatteini:
Fixes #489

### DIFF
--- a/src/in_memory.rs
+++ b/src/in_memory.rs
@@ -67,7 +67,7 @@ impl TermLike for InMemoryTerm {
                 .state
                 .lock()
                 .unwrap()
-                .write_str(&*format!("\x1b[{}A", n)),
+                .write_str(&format!("\x1b[{}A", n)),
         }
     }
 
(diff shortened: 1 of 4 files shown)


djc (approved):

djc:
Seems reasonable, thanks for looking into it!

devmatteini:
The lint errors seem unrelated, but I can fix them if you want

djc:
Would be great, thanks!

PR #495, opened 2022-12-02
Update estimate when using a steady ticker

### LINKED ISSUE #493: Custom keys on ProgressStyle do not work together with `enable_steady_tick`
It seems as though there are the following two interactions in indicatif 0.17.2, which were not present in 0.17.0-rc4:

- If `set_style` is used _before_ `enable_steady_tick`, the steady tick takes effect, but any custom keys that are set on the ProgressStyle using its `with_key` method are never updated.
- If `set_style` is used _after_ `enable_steady_tick`, then the custom keys work but the steady tick is disabled.

aawsome:
closes #493

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -640,8 +640,7 @@ impl TickerControl {
                 break;
             }
 
-            state.state.tick = state.state.tick.saturating_add(1);
-            state.draw(false, Instant::now()).ok();
+            state.tick(Instant::now());
 
             drop(state); // Don't forget to drop the lock before sleeping
             drop(arc); // Also need to drop Arc otherwise BarState won't be dropped


src/progress_bar.rs
```diff
@@ -641,7 +641,7 @@ impl TickerControl {
             }
 
             state.state.tick = state.state.tick.saturating_add(1);
-            state.draw(false, Instant::now()).ok();
+            state.update_estimate_and_draw(Instant::now());
```
aawsome, line 644: Actually, we could simply call `state.tick(Instant::now())` here.

aawsome: I changed this.

tobloef:
Awesome 🎉  Could we get this merged?

djc:
@chris-laplante do you want to review this?

chris-laplante (approved):
LGTM!

PR #498, opened 2022-12-11
address a bunch of lints

danieleades:


djc:
I'd prefer to leave out "avoid using `match` on bools" and "remove neeedless pass by value", please remove those commits.

danieleades:
> I'd prefer to leave out "avoid using `match` on bools" and "remove neeedless pass by value", please remove those commits.
> 
> Is this clippy at pedantic level or something?

it's combination of pedantic and nursery lints, as well as some standard lints triggered by edits made in fixing other lints.

i've dropped those two commits

djc (approved):

djc:
Thanks!

PR #504, opened 2023-01-01
Remove `"fs"` feature in tokio dependency.

azriel91:
Heya, this allows `indicatif` to compile to WASM. I've checked and `tokio::fs` is not used in the project.


djc (approved):
LGTM!

azriel91:
ah, it's used in doctests, let me update

PR #510, opened 2023-01-16
docs(lib): update style reference for alignment characters

### LINKED ISSUE #509: Add some documentation on what the alignment characters do
Currently the documentation lists `[<^>]           for an optional alignment specification` but does not specify what each character does and where it aligns to.

I would expect there to only be 2 characters, plus maybe a center character but am not sure which are which and what the others do.

hasezoey:
This PR updates the Templates sections's style characters for Alignment, to have a consistent style guide (removing characters that are not actually part of the style) and adding a clarification on what each character does 

fixes #509

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,7 +126,7 @@
 //! where the `options` part is optional.  If provided the format is this:
 //!
 //! ```text
-//! [<^>]           for an optional alignment specification
+//! <^>             for an optional alignment specification (Left, Center and Right respectively)
 //! WIDTH           an optional width as positive integer
 //! !               an optional exclamation mark to enable truncation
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -126,7 +126,7 @@
 //! where the `options` part is optional.  If provided the format is this:
 //!
 //! ```text
-//! [<^>]           for an optional alignment specification
+//! <^>             for an optional alignment specification (Left, Center and Right respectively)
```
djc, line 129: Style nit: I'd prefer to not capitalize the words.

djc (commented):
Thanks!

djc (approved):

PR #516, opened 2023-02-02
Fix `MultiProgress` alignment handling and migrate from structopt => clap

chris-laplante:


djc:
Yeah, clap is at 1.64 MSRV right now. Maybe we should pin it? It's only a dev-dep for us, anyway.

chris-laplante:
We could, though then we'd still have `atty` and maybe `ansi_term` :/. Right now our github actions is using 1.54. Maybe it's time to bump that up as well? (Though not necessarily all the way to 1.64)

On a side note, I may also open a PR to move off of action-rs/toolchain and onto https://github.com/dtolnay/rust-toolchain, since the former is unmaintained. A benefit is that we can use a toolchain expression like this: https://github.com/dtolnay/rust-toolchain#toolchain-expressions

djc:
I'm fine bumping the MSRV to 1.56, maybe 1.58?

Moving to dtolnay/rust-toolchain sounds good, but I'd rather make MSRV bumps an explicit choice.

chris-laplante:
OK I think this is ready

djc (approved):

djc:
Nice!

Issue #521, opened 2023-02-25
`inc` is call before actual work in examples


tatref:
Hi,

In some examples, `inc` is call before "work", for example in the `single.rs` example, we call `inc`, then `sleep` the next line:

https://github.com/console-rs/indicatif/blob/fb8d0e0c912104d30554d3a1ce09990f3bd430a8/examples/single.rs#L9

Should it be called after the heavy lifting?

djc:
Probably. Want to submit a PR?

tatref:
Sure! I just wanted to be sure that my assumption was correct

chris-laplante:
Closing since PR was merged

PR #522, opened 2023-02-26
fix: inc after work

tatref:
See #521


djc (approved):
I guess this makes sense. I hesitate to spend much more time on this, but I do wonder if some of these `set_message()` calls belong before the `sleep()`/work?

tatref:
I guess the effect of `set_message` is only seen after calling `inc`? So this should probably be just after `inc`?

I missed it in the finebars example:

```
thread::sleep(wait);
pb.inc(1);
pb.set_message(format!("{:3}%", 100 * i / 512));
```

djc (approved):

PR #523, opened 2023-03-07
Fix docs for `ProgressDrawTarget`

tillarnold:
The docs for `ProgressDrawTarget` were not updated when the API changed in 0.17.

The docs where still talking about the `to_term` function and the new `term` function no longer takes a `Option<u8>`.


djc:
Thanks!

PR #524, opened 2023-03-15
Color % symbol in template

eatradish:
Add percent% to convenience set % symbol color

Before:
```
{percent:>3.blue}" + &style("%").blue().to_string()
```

After:
```
 {percent%:>3.blue}"
```


djc:
I don't think I want to add this to the code base, since I don't think it's an important concern whether the % appears colored or not.

PR #526, opened 2023-03-21
Allow rate-limiting TermLike targets

akx:
This adds an optional `rate_limiter` to custom `TermLike` targets, so they don't need to implement rate limiting by hand.


djc (commented):
Seems reasonable to me, thanks! @chris-laplante any thoughts?

djc (approved):

chris-laplante (approved):
Looks good to me too!

Issue #525, opened 2023-03-21
Machine-readable (JSON) output


akx:
I know this is a bit leftfield, maybe, but I have a project where I'd like to use Indicatif's TTY progress bars when human eyes are operating the application, but with an option to use JSON ([NDJSON/jsonlines](http://ndjson.org/)) output instead when working as a subprocess, and I'd like to do this without duplicating the ETA, etc. calculations and rate limiting since Indicatif is already doing a great job at that.

What would be a good approach for this? 

Since `ProgressStyle` isn't a trait, it's not easy to pass in a JSON-formatting `ProgressStyle` from userland (or at least I couldn't figure it out); just passing in a JSON-esque `template` for a regular ProgressStyle isn't great since it doesn't do JSON encoding for strings – and even so, the TTY-related code (moving cursors and clearing progress bars and all that) wouldn't be necessary; you'd just always print a newline after a message.

To clarify, I'm envisioning something like `ProgressBar::new_json(["pos", "len", "eta"])` outputting e.g.

```
{"pos": 1, "len": 100, "eta": 14.3}
{"pos": 39, "len": 100, "eta": 7.7}
{"pos": 100, "len": 100, "eta": 0}
```
(with maybe a `complete` marker afterwards? That's a detail anyway.)

If there's no great way to do this from userspace, would a PR implementing something like this be appreciated and/or approved? 😄

PR #527, opened 2023-03-21
Add option for JSON escaping for placeholder values

akx:
Refs #525 (I went ahead and wrote some code...).

This PR

* adds an option (`json_strings`) to have `ProgressStyle` escape all of the dynamic strings it outputs as JSON strings
* a convenience helper `with_json_keys` to create a `ProgressStyle` with a template that outputs a well-formed JSON object

IOW, a `ProgressStyle` with `json_strings(true)` and a template
```
{{"pos": {pos}, "len": {len}, "prefix": {prefix}, "msg": {msg}}}
```
(this is what `ProgressStyle::with_json_keys(&["pos", "len", "prefix", "msg"])` would internally generate) will result in progress lines like 
```
{"pos": "1", "len": "10", "prefix": "hello", "msg": "something"}
```
which can then be sent to a `TermLike` target (#354, #526) to implement #525 :)

---

If you like, I can put this behind a `feature` gate to avoid the `json` dependency – but what should `indicatif` do when trying to do JSON output when it's not compiled in? I'm pretty sure we wouldn't want it to just output non-escaped JSON...

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -10,6 +10,8 @@ use instant::Instant;
 #[cfg(feature = "unicode-segmentation")]
 use unicode_segmentation::UnicodeSegmentation;
 
+use json;
+
 use crate::format::{
     BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, HumanCount, HumanDuration,
     HumanFloatCount,
@@ -25,6 +27,7 @@ pub struct ProgressStyle {
     char_width: usize,
     tab_width: usize,
     pub(crate) format_map: HashMap<&'static str, Box<dyn ProgressTracker>>,
+    json_strings: bool,
 }
 
 #[cfg(feature = "unicode-segmentation")]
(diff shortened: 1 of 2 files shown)


src/style.rs
```diff
@@ -340,17 +356,41 @@ impl ProgressStyle {
...
+                                    ))),
+                                    None => cur.push_str(&json::stringify(buf.clone())),
+                                }
+                            } else {
+                                match style {
+                                    Some(s) => {
+                                        cur.write_fmt(format_args!("{}", s.apply_to(&buf))).unwrap()
+                                    }
+                                    None => cur.push_str(&buf),
+                                }
```
akx, line 391: (I'm personally not 100% happy with the complexity here, so if someone has better ideas...)

djc:
Sorry for being slow to respond -- while I think your use case is something we'd like to accomodate, I think this PR is probably not the right way to go.

In thinking some more about it, do you think it would be feasible to write a custom [`ProgressTracker`](https://docs.rs/indicatif/latest/indicatif/style/trait.ProgressTracker.html) that outputs the JSON? It seems like a viable path to me, in that the `impl` could take care of all the particular JSON needs.

djc:
Going to close this for now, feel free to reopen if you want to work on this again (or open a new one).

PR #529, opened 2023-04-07
Mention the prefix and message placeholders in the `with_` docs

### LINKED ISSUE #528: `ProgressBar::with_message` does nothing?
I couldn't find an example or an existing issue about this. What does the message do?

Minimal example:

```rust
use std::{thread, time::Duration};

use indicatif::ProgressBar;

fn main() {
    let pb = ProgressBar::new(100).with_message("Frobbing the widget");
    for _ in 0..100 {
        thread::sleep(Duration::from_millis(30));
        pb.inc(1);
    }
}
```

lnicola:
Closes #528

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -80,6 +80,9 @@ impl ProgressBar {
     }
 
     /// A convenience builder-like function for a progress bar with a given prefix
+    ///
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #531, opened 2023-04-16
Add `contents_formatted` to `InMemoryTerm`

dfaust:
This method is almost identical to `contents`, but it includes VT-100 codes.

image

### DIFF
--- a/src/in_memory.rs
+++ b/src/in_memory.rs
@@ -53,6 +53,45 @@ impl InMemoryTerm {
         rows.reverse();
         rows.join("\n")
     }
+
+    pub fn contents_formatted(&self) -> Vec<u8> {
+        let state = self.state.lock().unwrap();
+
+        // For some reason, the `Screen::contents` method doesn't include newlines in what it
+        // returns, making it useless for our purposes. So we need to manually reconstruct the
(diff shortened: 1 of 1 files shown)


djc:
Why do you need this?

dfaust:
For testing. This allows seeing the colored output directly in the tests.

djc (commented):
Sorry for being so slow to follow-up. Adding this API seems reasonable, though I'd prefer to add it in a way that doesn't duplicate as much code. How about we move most of the logic into a new `InMemoryTermState::contents()` method which can take a `formatted: bool` argument?

dfaust:
I just pushed an updated version, which adds a reset code at the end of each line.
This is necessary because `rows_formatted` returns each line in an "undefined state".

Yes, there is some code duplication, the problem is, that `contents` returns a String, while `contents_formatted` returns a Vec<u8>. And this starts basically with the first call, since `rows` returns a String and `rows_formatted` returns a Vec<u8>.

One way would be for the `contents` method to turn the strings into vecs, process them and at the end turn the result vec back into a string.

djc (approved):

djc:
Okay, thanks!

PR #532, opened 2023-04-16
Implement Clone for ProgressDrawTarget

dfaust:
Uses [dyn-clone](https://github.com/dtolnay/dyn-clone) to make `ProgressDrawTarget` and `TermLike` cloneable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ rust-version = "1.58"
 
 [dependencies]
 console = { version = "0.15", default-features = false, features = ["ansi-parsing"] }
+dyn-clone = "1.0.11"
 number_prefix = "0.4"
 portable-atomic = "1.0.0"
 rayon = { version = "1.1", optional = true }
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -18,7 +18,7 @@ use crate::TermLike;
 /// The draw target is a stateful wrapper over a drawing destination and
 /// internally optimizes how often the state is painted to the output
 /// device.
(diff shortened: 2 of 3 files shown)


djc:
What's your use case for this?

dfaust:
This allows creating multiple progress bars within one handler, if a ProgressDrawTarget is provided.
In my case, I'm watching a directory for changes and want to draw a new progress bar whenever changes are detected.

djc:
Why not use a `MultiProgress` for that?

dfaust:
In the case I mentioned, this would be possible.
But it requires to pass the progress bar down to the functions that use it, instead of the draw target.
I don't think that's a clean solution. I'm passing the draw target as part of a context object, and I don't want to create a progress bar just in case I might need it at some point.
Another use case (that I'm not using right now) would be, to print messages between progress bars.

Are you concerned about the additional dependency? In that case I could make it optional.

djc:
So why do you need to pass the `ProgressDrawTarget` anyway?

dfaust:
> So why do you need to pass the `ProgressDrawTarget` anyway?

I use it for testing.

djc:
Okay, that doesn't do anything to convince me that this is the right solution for the problem you're trying to address. Despite me asking several questions, you've yet to state a clear picture of what it is you're trying to achieve that you think necessitates passing around a `ProgressDrawTarget` which doesn't seem like a common way to use our API. If you would still like to have changes made in this project, please open an issue that outlines, at a conceptual level, what you are trying to do and why other solutions aren't acceptable for this use case.

Issue #520, opened 2023-02-20
Duplicate bars if the message is too long


kmanc:
I haven't been able to nail down exactly what the cause is for this, so a little bit of trial and error might be required to reproduce, but I've noticed that if I set a bar message to something that is "too long" my MultiProgress / ProgressBar combination starts behaving oddly. Is there anyone who can shed some light on why that is?

Rough setup

```
use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
use std::sync::Arc;

let mp = Arc::new(MultiProgress::new());
let bar = mp.clone().add(ProgressBar::new(0));
let style = ProgressStyle::with_template("{msg}").unwrap();
bar.set_style(style);
let message = "SOMETHING HERE";
bar.set_message(message.clone());
let message = format!("{message} LONGER");
bar.finish_with_message(message);
```

I've found that if I keep making "message" longer, eventually the bars start double or triple printing. I've poked around in the codebase but I'm not sure why that is, so any insight would be appreciated.

Thanks!

eggyal:
Just ran into this myself and, at least in my case, it was because something else was being written to the terminal "behind" the MultiProgress, which caused it to redraw (in the wrong place).  Using `MultiProgress::suspend` for that other output resolved the issue for me.

djc:
@kmanc have you tried #533? Would be good to know if that fixes your issue.

azriel91:
Heya @djc, I've tried the latest code on `main` and it fixes the issue for me :v:.

djc:
Thanks! Going to optimistically close this, please reopen if you're still seeing this with current main.

lgarron:
> Thanks! Going to optimistically close this, please reopen if you're still seeing this with current main.


I just tried 7c9465b75b2d8a284cde24bcf57636aa8489e2e0, and while https://github.com/console-rs/indicatif/pull/533 seems to be well resolved when the terminal is too narrow, this issue still seems to occur if the terminal is not tall enough.

OS: macOS Venture 13.4
cargo: 1.68.2
Repro branch: https://github.com/lgarron/folderify/tree/indicatif-v0.17.4

Issue #144, opened 2019-12-27
Long progress bar results in new-lines


ceranco:
# Description
* **OS**: *Windows 10*
* **Indicateif version**: 0.13.0
* **Shells**: *CMD*, *Git Bash* and *Powershell*

When using a `ProgressBar` that is longer than the width of the terminal, the progress bar doesn't clear itself, but instead prints a new line for each update.

### CMD
![cmd](https://user-images.githubusercontent.com/22302978/71531980-f922da00-28f9-11ea-89a1-edeab12e0bea.gif)

### Git Bash
![git-bash](https://user-images.githubusercontent.com/22302978/71531984-fc1dca80-28f9-11ea-8528-c952d07d8009.gif)

### Powershell
![powershell](https://user-images.githubusercontent.com/22302978/71532000-fde78e00-28f9-11ea-9a91-211545a570fa.gif)

The above gifs show the following code:
```rust
use std::cmp::min;
use std::thread;
use std::time::Duration;

use indicatif::{ProgressBar, ProgressStyle};

fn main() {
    let mut downloaded = 0;
    let total_size = 231231231;

    let pb = ProgressBar::new(total_size);
    pb.set_style(ProgressStyle::default_bar()
        .template("{msg:>12.cyan.bold} {spinner:.green} [{elapsed_precise}] [{bar:40.cyan/blue}] {bytes}/{total_bytes} ({bytes_per_sec}, {eta})")
        .progress_chars("#>-"));

    pb.set_message("Downloading");
    while downloaded < total_size {
        let new = min(downloaded + 223211, total_size);
        downloaded = new;
        pb.set_position(new);
        thread::sleep(Duration::from_millis(1));
    }

    pb.set_style(
        ProgressStyle::default_bar().template("{msg:>12.green.bold} downloading {total_bytes:.green} in {elapsed:.green}"),
    );
    pb.finish_with_message("Finished");
}

```

P.S. - This library is awesome! Thanks for making it 😄

mibac138:
You can use `wide_bar` instead of `bar` to make it automatically adjust it's size to the terminal size so that the line doesn't overflow. FYI there are currently two `wide` elements: 
- `wide_bar` which adjusts the bar's width to avoid line overflows
- `wide_msg` which truncates the message

Keep in mind that there can only be one `wide` element per `ProgressBar`.

ceranco:
Thanks for your reply @mibac138.

[...]

PR #533, opened 2023-04-17
Handle terminal line wrap to avoid new line

RDruon:
Terminal usually wrap line in case it can't fit the current terminal width. 
This behavior is not taken in account by indicatif and create some visual artifacts.

The issue was reported as #144 (closed with a workaround) and #520.

Please find below a demo **without** the patch:
WindowsTerminal_wwpvPvmRKY

Demo **with** the patch:
WindowsTerminal_Q7z1mdXh3O

Using #144 example:
WindowsTerminal_HVhbcstdMB

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -494,7 +494,13 @@ impl DrawState {
         };
 
         let len = self.lines.len();
+        let mut real_len = 0;
         for (idx, line) in self.lines.iter().enumerate() {
+            // Calculate real length based on terminal width
+            // This take in account linewrap from terminal
+            real_len +=
+                (console::measure_text_width(line) as f64 / term.width() as f64).ceil() as usize;
+
             if idx + 1 != len {
                 term.write_line(line)?;
             } else {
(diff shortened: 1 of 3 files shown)


djc (approved):
Seems like a nice improvement to me! @chris-laplante what do you think?

Would be nice to have some tests for this.

chris-laplante:
> Seems like a nice improvement to me! @chris-laplante what do you think?
> 
> Would be nice to have some tests for this.

Makes sense to me too! I agree some tests would be nice. @RDruon are you able to implement a render test or two for this?

RDruon:
> > Seems like a nice improvement to me! @chris-laplante what do you think?
> > Would be nice to have some tests for this.
> 
> Makes sense to me too! I agree some tests would be nice. @RDruon are you able to implement a render test or two for this?

I added 2 render tests, one with `ProgressBar` and one with `MultiProgress`

RDruon:
@chris-laplante could I get another review? It seems several projects are interested in this change.

djc (approved):

djc:
I think this is looking good, thanks! @chris-laplante feel free to follow-up if you have more thoughts.

chris-laplante:
Sorry for the delay! Looks good to me as well

Issue #534, opened 2023-04-23
With burst-heavy writes, eta and bitrate show incorrect values when using enable_steady_tick


afontenot:
TL;DR: if you use `enable_steady_tick`, and `set_position` on every write, a burst-oriented workload will show incorrect values for ETA and bitrate - not just heavily fluctuating, but **impossibly low** values given the true speed of the connection.

-------------------------------

I have the following two assumptions about how the progress bar behaves. These might need to be corrected.

1. When `enable_steady_tick` is enabled, indicatif *only* ticks on the clock you set. Based on [the docs](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.enable_steady_tick).
2. Rate-dependent metrics like ETA and bitrate are determined by looking at a sliding window of the last 15 ticks. That is to say, every time there's a tick, the progress is stored. The rate is just the amount of progress made made in the last 15 ticks, divided by the amount of time it took for 15 ticks to elapse. Based on [this contributor comment](https://github.com/console-rs/indicatif/issues/394#issuecomment-1344760265).

I was debugging a progress bar issue with [magic-wormhole.rs](https://github.com/magic-wormhole/magic-wormhole.rs) that was hard to reproduce. The issue seems to stem from how burst-heavy my system was. I regularly see the program fill the outgoing TCP buffer, and then wait for half a second or more before continuing. The way the program works is that it calls `set_position` every time it writes into the buffer. It sets the steady tick in indicatif to 250 ms. Based on this, I expect the progress bar to fluctuate a decent amount, but *on average* it should be close to the true bitrate of the file transfer.

In fact, I see *extremely strong* fluctuations, and the indicatif progress bar is *almost always* significantly below the real speed of the transfer (based on the file size and transfer time).

Based on the assumptions I gave above, I recreated a moving average (using matplotlib in Python) across writes from the last `15 * 250 ms`.

![Figure_1](https://user-images.githubusercontent.com/3870410/233815391-b62101fe-8bc8-43dc-a92a-da33fc4e4bda.png)

As you can see, there's some undesirable variability here (it would be nice if we could set indicatif to average over 60 samples or something like that), but there are no extreme dips in the moving average. On the contrary, indicatif frequently shows me bitrates under 1 Mbps for this connection, and appears to tick irregularly as well! Something seems to have gone badly wrong with the estimation.

[...]

PR #535, opened 2023-04-24
Fix incorrect seconds_per_step calculation

afontenot:
seconds_per_step looks at a ring buffer with samples taken each tick. Each sample is

    <duration of the tick> / <number of progress steps>

This results in an incorrect seconds_per_step, when this is calculated as a raw average of the buffer values. This is because we have no reason to assume any tick is a more representative sample than any other tick, but this will treat ticks with very few steps as *more* representative than steps with many ticks, because the value of the denominator is lost in the sample.

The result is a very poor ETA / bitrate estimation when step production follows a burst-wait pattern, since ticks with few or no steps are to be expected.

A better estimate could be achieved by averaging over the step-rate of each tick, but this would still not be ideal since it would treat short ticks as equally representative as long ticks in the average.

Instead, we make two changes:

First, we change the buffer to store the total number of steps rather than the difference between steps. This necessitates changing the new / reset code to set the first buffer item to zero.
    
Second, we create a second ring buffer to store the timestamp of each tick. As a result, we can calculate the `seconds_per_step` rate by simply looking at the earliest values in the buffer, and returning `(now() - first_time) / (last_steps - first_steps)`.

By using `now()` instead of `last_time` in the above equation, we also fix an additional issue. Since we only add items to the buffer when new steps occur, the progress bar freezes and the ETA does not update when progress stalls. By always using the current time, we get a more accurate `seconds_per_step` estimate with no additional overhead.

Fixes https://github.com/console-rs/indicatif/issues/534.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -370,7 +370,8 @@ impl TabExpandedString {
 /// Ring buffer with constant capacity. Used by `ProgressBar`s to display `{eta}`,
 /// `{eta_precise}`, and `{*_per_sec}`.
 pub(crate) struct Estimator {
-    steps: [f64; 16],
+    steps: [u64; 16],
+    step_instants: [Instant; 16],
     pos: u8,
     full: bool,
     prev: (u64, Instant),
@@ -379,16 +380,16 @@ pub(crate) struct Estimator {
 impl Estimator {
     fn new(now: Instant) -> Self {
         Self {
-            steps: [0.0; 16],
-            pos: 0,
+            steps: [0; 16],
+            step_instants: [now; 16],
+            pos: 1,
             full: false,
             prev: (0, now),
         }
     }
 
     fn record(&mut self, new: u64, now: Instant) {
-        let delta = new.saturating_sub(self.prev.0);
-        if delta == 0 || now < self.prev.1 {
+        if new <= self.prev.0 || now < self.prev.1 {
             // Reset on backwards seek to prevent breakage from seeking to the end for length determination
             // See https://github.com/console-rs/indicatif/issues/480
             if new < self.prev.0 {
@@ -397,14 +398,9 @@ impl Estimator {
             return;
         }
 
-        let elapsed = now - self.prev.1;
-        let divisor = delta as f64;
-        let mut batch = 0.0;
-        if divisor != 0.0 {
-            batch = duration_to_secs(elapsed) / divisor;
-        };
+        self.steps[self.pos as usize] = new;
+        self.step_instants[self.pos as usize] = now;
 
-        self.steps[self.pos as usize] = batch;
         self.pos = (self.pos + 1) % 16;
         if !self.full && self.pos == 0 {
             self.full = true;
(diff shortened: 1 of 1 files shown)


afontenot:
Side note: there's a related bug that this PR doesn't currently fix. If you have a tick with no steps, nothing gets added to the buffer. There are two problems this causes:

 * If you're using a steady tick, the progress bar briefly stalls until it gets some more data, which looks bad.
 * During lengthy stalls where no steps are added to the progress bar, the ETA does not update.

If you remove the check so that the buffer gets updated on a tick with no steps, you can end up with an empty buffer. This causes a panic downstream of the fact that `<duration> / <steps>` involves a division by zero.

Rather than add a complicated workaround for this, I wonder how you feel about the following patch, which applies on top of this PR:

```diff
diff --git a/src/state.rs b/src/state.rs
index 171b485..b782fc8 100644
--- a/src/state.rs
+++ b/src/state.rs
@@ -423,7 +423,7 @@ impl Estimator {
         let len = self.len();
         let total_steps = self.steps[0..len].iter().sum::<u64>();
         let total_duration = self.step_durations[0..len].iter().sum::<f64>();
-        total_duration / total_steps as f64
+        (total_duration + duration_to_secs(Instant::now() - self.prev.1)) / total_steps as f64
     }
 
     fn len(&self) -> usize {
```

Rather than allowing the buffer to clear, we change `seconds_per_step` to include not just the time elapsed during the buffer-setting ticks, but *also* the time that elapsed during the *last* buffer-setting tick, since this time will eventually be added to the *next* tick that sets the buffer. This allows downstream values like ETA to continue updating during a stall.

The only question is whether this is an okay place in the code to handle the problem, or whether it should be refactored and placed elsewhere.

afontenot:
The following two changes required modifying the tests:

> changing the new / reset code to set the first buffer item to zero.

Means that a newly initialized ring buffer contains 1 element, not 0.

> using `now()` instead of `last_time` in the above equation

Depending on the latest time when calculating `seconds_per_step` mean that the test which checks that the rate calculation is correct can't be exact any more. Instead, I check that it is between the rates expected for two very close Instant values - immediately before and after the call to `seconds_per_step`.

djc:
I'm not able to review this in detail for the next week or two, but here is some early feedback:

- Given the sensitivity of this code, it would be great if you could add a whole bunch more comments to the parts of the `Estimator` that you changed to elaborate what is going on and why.
- It looks like you modified existing tests to work but didn't add much test coverage? In particular, it would be nice to explicitly test behavior for the motivating use case of ticks with more steps weighing more than ticks with fewer steps as well as the additional issue you noted (potential freezes).
- Tests for any overflow/underflow/divide-by-zero scenarios would also be great to have.

Thanks again for working on this, your contribution is much appreciated!

afontenot:
@djc thanks, I'll add some comments and improve the tests

Quick note on this:

> Tests for any overflow/underflow/divide-by-zero scenarios would also be great to have.

Yep, the status quo here is a bit of a mess actually. You can easily end up dividing by zero if e.g. time passes but no steps occur. It actually relies ensuring that the numerator in the rate calculation is always zero if the denominator is, so that you get a NaN float, and then explicitly tests for this elsewhere in the code.

I believe I imitated the existing behavior in this area exactly, which is the reasoning behind the strange code here:

        let mut duration: f64 = 0.0;
        if steps != 0 {
            duration = duration_to_secs(now - self.step_instants[first_pos]);
        }

But it would be good to test for this, or maybe change the behavior to not rely on NaN floats. Of course the latter would require having an opinion on the correct value to show for the ETA in the case that time has passed but no progress has been made. As I didn't have an opinion, I fell back on maintaining the status quo.

afontenot:
Just a heads up that I have put improvements for this PR on hold until my work on an exponentially weighted average can see some review. If we end up switching the estimator to use that approach, this PR will become irrelevant and need to be closed.

afontenot:
Going to close this for now as it looks like we will be able to merge some form of an exponentially weighted average, which solves this problem in a different way.

Issue #536, opened 2023-04-29
Resetting the time elapsed does not also reset the ETA


afontenot:
I think this is just an oversight, but I want to confirm that instead of sending a PR without justification.

I can see how it would make sense to want to reset the ETA but *not* the time elapsed. E.g. maybe progress stalled for a while, or something, and you want a clean estimate without resetting the time elapsed counter. On the other hand, I can't imagine wanting to reset the time elapsed for the whole progress bar without expecting that to also reset the ETA. It doesn't make very much sense for the progress bar to say it's only been active 1 minute, when the ETA / bitrate is based on data from the last two minutes.

My thinking is that the `Reset::Elapsed` mode should also reset `::Eta`, but not vice versa, and that both of these should be independent of `Reset::All` (because you only want to recreate the whole progress bar in specific situations).

chris-laplante:
Your reasoning makes sense to me; I think it was just an oversight.

djc:
Agreed. Want to submit a PR?

PR #538, opened 2023-05-07
resetting the elapsed time also resets ETA

afontenot:
It makes sense to reset the ETA without resetting the time elapsed back to zero, but when setting the time elapsed back to zero, the ETA should also be reset as this should make the progress bar behave as if it had just been activated (in a partially complete state).

Fixes https://github.com/console-rs/indicatif/issues/536.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -69,9 +69,9 @@ impl BarState {
     }
 
     pub(crate) fn reset(&mut self, now: Instant, mode: Reset) {
-        if let Reset::Eta | Reset::All = mode {
-            self.state.est.reset(now);
-        }
+        // Always reset the estimator; this is the only reset that will occur if mode is
+        // `Reset::Eta`.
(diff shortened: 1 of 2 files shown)


src/state.rs
```diff
@@ -75,6 +75,7 @@ impl BarState {
 
         if let Reset::Elapsed | Reset::All = mode {
             self.state.started = now;
+            self.state.est.reset(now);
```
djc, line 78: This minimal change doesn't really make sense, since we're now unconditionally resetting the `Estimator` in all modes, so we can just pull it out of the first `if` and remove it from the second one?

afontenot: Thanks, I should have looked more closely here. I'll force push to the branch to fix this.

djc (approved):
Thanks!

djc:
@chris-laplante further feedback welcome!

PR #539, opened 2023-05-08
Switch Estimator to use an exponential weighting 

afontenot:
This is an implementation of an exponentially weighted running average with a tuning parameter (currently a constant in this implementation):

 * the `ews` parameter is a number of seconds. Progress the `Estimator` has observed that is older than this value will receive a total weight of 0.1 in the rate estimation, and newer values will receive a total of 0.9. The default is 15 seconds.

This implementation does double smoothing by applying the running average to itself. The result avoids undesirable instantaneous movements in the estimate when large updates occur.

The exponential estimator works by keeping a running tally, where an existing tally that has aged `age` seconds is reweighted such that

    weight ^ (ewa / age) = 0.1

For instance, data aged 5 seconds with a 15 second weight parameter would receive `weight = 0.1 ^ (5/15) = 0.464`. If it then ages another 10 seconds, it would receive `weight = 0.1 ^ (10/15) = 0.215`. After being multiplied by both weights, it would have a weight of `0.464 * 0.215 = 0.1`, as expected.

A couple of basic features are also implemented for higher quality estimates:

 * We divide out any weight given to data before the estimator was     initialized, since these values are meaningless.

 * When returning an estimate, we include the time since the last updated was received as non-progress, which means that the estimator does not freeze when progress stalls.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -384,7 +384,8 @@ pub(crate) struct Estimator {
     steps: [f64; 16],
     pos: u8,
     full: bool,
-    prev: (u64, Instant),
+    prev_steps: u64,
+    prev_time: Instant,
 }
 
 impl Estimator {
@@ -393,22 +394,23 @@ impl Estimator {
             steps: [0.0; 16],
             pos: 0,
             full: false,
-            prev: (0, now),
+            prev_steps: 0,
+            prev_time: now,
         }
     }
 
-    fn record(&mut self, new: u64, now: Instant) {
-        let delta = new.saturating_sub(self.prev.0);
-        if delta == 0 || now < self.prev.1 {
+    fn record(&mut self, new_steps: u64, now: Instant) {
+        let delta = new_steps.saturating_sub(self.prev_steps);
+        if delta == 0 || now < self.prev_time {
             // Reset on backwards seek to prevent breakage from seeking to the end for length determination
             // See https://github.com/console-rs/indicatif/issues/480
-            if new < self.prev.0 {
+            if new_steps < self.prev_steps {
                 self.reset(now);
             }
             return;
         }
 
-        let elapsed = now - self.prev.1;
+        let elapsed = now - self.prev_time;
         let divisor = delta as f64;
         let mut batch = 0.0;
         if divisor != 0.0 {
@@ -421,13 +423,15 @@ impl Estimator {
             self.full = true;
         }
 
-        self.prev = (new, now);
+        self.prev_steps = new_steps;
+        self.prev_time = now;
     }
 
     pub(crate) fn reset(&mut self, now: Instant) {
         self.pos = 0;
         self.full = false;
-        self.prev = (0, now);
+        self.prev_steps = 0;
+        self.prev_time = now;
     }
 
     /// Average time per step in seconds, using rolling buffer of last 15 steps
@@ -448,7 +452,8 @@ impl fmt::Debug for Estimator {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Estimate")
             .field("steps", &&self.steps[..self.len()])
-            .field("prev", &self.prev)
+            .field("prev_steps", &self.prev_steps)
+            .field("prev_time", &self.prev_time)
             .finish()
     }
 }
--- a/src/state.rs
+++ b/src/state.rs
@@ -1,9 +1,9 @@
 use std::borrow::Cow;
+use std::io;
 use std::sync::Arc;
 use std::time::Duration;
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::Instant;
-use std::{fmt, io};
 
 #[cfg(target_arch = "wasm32")]
 use instant::Instant;
@@ -230,13 +230,14 @@ pub struct ProgressState {
 
 impl ProgressState {
     pub(crate) fn new(len: Option<u64>, pos: Arc<AtomicPosition>) -> Self {
+        let now = Instant::now();
         Self {
             pos,
             len,
             tick: 0,
             status: Status::InProgress,
-            started: Instant::now(),
-            est: Estimator::new(Instant::now()),
+            started: now,
+            est: Estimator::new(now),
             message: TabExpandedString::NoTabs("".into()),
             prefix: TabExpandedString::NoTabs("".into()),
         }
@@ -276,7 +277,7 @@ impl ProgressState {
 
         let pos = self.pos.pos.load(Ordering::Relaxed);
 
-        let sps = self.est.steps_per_second();
+        let sps = self.est.steps_per_second(Instant::now());
 
         // Infinite duration should only ever happen at the beginning, so in this case it's okay to
         // just show an ETA of 0 until progress starts to occur.
@@ -298,7 +299,7 @@ impl ProgressState {
     /// The number of steps per second
     pub fn per_sec(&self) -> f64 {
         if let Status::InProgress = self.status {
-            self.est.steps_per_second()
+            self.est.steps_per_second(Instant::now())
         } else {
             let len = self.len.unwrap_or_else(|| self.pos());
             len as f64 / self.started.elapsed().as_secs_f64()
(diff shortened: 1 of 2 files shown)


afontenot:
For the sake of having a draft PR to discuss and test, I put the `now` arguments back in the `Estimator` functions so that it would be trivial for me to fix / add tests without having to worry about time.

Couple of notes on things that might need changing before this is merged:

1. I added a few functions to the Estimator to allow configuring it. You can change the weighting and also disable / enable the double smoothing (using an exponentially weighted average of an exponentially weighted average). These aren't exposed in any way currently, and I wasn't sure what the right way to do that might be.

2. The tests need a look over to make sure everything important is covered. I kept a sanity check, added a correctness check for both single and double EWA, added a test to make sure resetting the ETA behaves as expected, and reworked one test that makes sure the estimator behaves correctly when the position is rewound.

3. It's not clear to me exactly what issue https://github.com/console-rs/indicatif/issues/480 is talking about. I think the behavior I have in this area preserves or improves on the status quo, but I'm not totally sure. Basically, I think it makes sense to force a estimator reset when the position is rewound - but you don't want to set the previous position to zero, because that would result in a *huge* event rate if the true position isn't zero. Instead, I only reset the *estimate* and trust that the `update` function has been called with the correct value.

src/state.rs
```diff
@@ -365,78 +370,126 @@ impl TabExpandedString {
...
+        // We achieve this by diluting with the equation `weight ^ (ews / delta_t) = 0.1`.
+        10_f64.powf(-age / self.exponent_weight_seconds)
     }
 
-    fn len(&self) -> usize {
-        match self.full {
-            true => 16,
-            false => self.pos as usize,
+    /// Average time per step in seconds, using single or double exponential smoothing
+    fn steps_per_second(&self) -> f64 {
```
djc: Let's have this take a `now`?

src/state.rs
```diff
@@ -365,78 +370,126 @@ impl TabExpandedString {
...
             return;
         }
 
-        let elapsed = now - self.prev.1;
-        let divisor = delta as f64;
-        let mut batch = 0.0;
-        if divisor != 0.0 {
-            batch = duration_to_secs(elapsed) / divisor;
-        };
+        let delta_steps = new - self.prev.0;
```
djc, line 426: One trick I used in a previous job was to just straight up set the `smoothed_steps_per_sec` (and `double_smoothed_steps_per_sec`) to the new value if they were uninitialized. Maybe that would be a decent approach here? Should those values become `Option` in that case?

src/state.rs
```diff
@@ -365,78 +370,126 @@ impl TabExpandedString {
...
-            false => self.pos as usize,
+    /// Average time per step in seconds, using single or double exponential smoothing
+    fn steps_per_second(&self) -> f64 {
+        let now = Instant::now();
+
+        // reweight to account for the passage of time since last tick
+        let delta_t = duration_to_secs(now - self.prev.1);
+        let reweight = self.weight(delta_t);
+
+        // The Estimator is initialized with zero values, which are included in the underlying
```
djc, line 471: See my other comment about avoiding the pathologies of an uninitialized estimate.

src/state.rs
```diff
@@ -365,78 +370,126 @@ impl TabExpandedString {
...
-        match self.full {
-            true => 16,
-            false => self.pos as usize,
+    /// Average time per step in seconds, using single or double exponential smoothing
+    fn steps_per_second(&self) -> f64 {
+        let now = Instant::now();
+
+        // reweight to account for the passage of time since last tick
+        let delta_t = duration_to_secs(now - self.prev.1);
+        let reweight = self.weight(delta_t);
```
djc, line 469: This could use some comments on how/why we're reweighting for the passage of time since the last `record()`.

src/state.rs
```diff
@@ -365,78 +370,126 @@ impl TabExpandedString {
     }
 }
 
-/// Estimate the number of seconds per step
+/// Estimate the number of steps per second
+///
+/// Exponentially smoothed estimate with an adjustable weighting
```
djc, line 376: Would be good to comment a bit more on the how and why of the double smoothing trick.

djc:
This is looking great, thanks for working on it!

> For the sake of having a draft PR to discuss and test, I put the `now` arguments back in the `Estimator` functions so that it would be trivial for me to fix / add tests without having to worry about time.

Nice. The reason these are here is (a) for performance, calling `Instant::now()` means doing a syscall which can be somewhat expensive relative to the other stuff we do, and (b) for consistency, we usually want to use a consistent timestamp for all of a given invocation of the public API. If we want to deviate from these principles for some reason, there should be a clear comment explaining why the deviation is necessary/useful.

> A few notes on things that might need changing before this is merged:
> 
>     1. I added a few functions to the Estimator to allow configuring it. You can change the weighting and also disable / enable the double smoothing (using an exponentially weighted average of an exponentially weighted average). These aren't exposed in any way currently, and I wasn't sure what the right way to do that might be.

I've addressed this in my review feedback.

>     2. The tests need a look over to make sure everything important is covered. I kept a sanity check, added a correctness check for both single and double EWA, added a test to make sure resetting the ETA behaves as expected, and reworked one test that makes sure the estimator behaves correctly when the position is rewound.
> 
>     3. It's not clear to me exactly what issue [ProgressBarIter and finding stream length via seeking breaks rate estimation #480](https://github.com/console-rs/indicatif/issues/480) is talking about. I think the behavior I have in this area preserves or improves on the status quo, but I'm not totally sure. Basically, I think it makes sense to force a estimator reset when the position is rewound - but you don't want to set the previous position to zero, because that would result in a _huge_ event rate if the true position isn't zero. Instead, I only reset the _estimate_ and trust that the `update` function has been called with the correct value.

Right, I think this makes sense.

>     4. I didn't come up with an appropriate debug fmt for the Estimator after making my changes.

Maybe we can go back to just deriving `Debug`? I don't think we deeply care about the `Debug` values, though likely we'd want something that makes your life (as a contributor) or my life (as a maintainer) easier when debugging the `Estimator`.

src/state.rs
```diff
@@ -365,78 +370,126 @@ impl TabExpandedString {
...
 pub(crate) struct Estimator {
-    steps: [f64; 16],
-    pos: u8,
-    full: bool,
+    smoothed_steps_per_sec: f64,
+    double_smoothed_steps_per_sec: f64,
     prev: (u64, Instant),
+    start: Instant,
+    exponent_weight_seconds: f64,
+    double_smoothing: bool,
```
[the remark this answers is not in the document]
afontenot: That sounds fine to me. Though as some users in the comments pointed out, the behavior might become quite bad with rare events (on the order of an event every 30 seconds, say), because the smoothing would be happening over a (relatively) short period of time. 

I was also assuming some users would prefer single smoothing over double. But agreed on the importance of good defaults.

src/state.rs
afontenot: I think there are two separate questions here:

1. What should the estimator return when a value is requested but it is in the uninitialized state?
2. What the should original (uninitialized) *internal* value of the estimator be?

Because the weight of data in the estimator is always positive (there's no horizon at which the weight = 0), the only sensible answer to (2) is 0. That's because you have no way to pick the right value ahead of time, so you will want to divide out the effect of the original estimator value (I call this debiasing). This is much easier to do if it is 0, because then you just divide the naive estimate by `weight(age)`.

I don't really have an opinion on (1) so I'm happy to have it do whatever you like. If you'd like to have an uninitialized estimator return `None`, that's fine with me.

BTW in case it's unclear, after the first `update` the estimator value *will* be equal to the value of the first update (for both smoothing and double smoothing), it's just that prior to the first update there's not a correct rate to return.

afontenot:
Most outstanding issues resolved. Heads up that I will be force pushing in a moment to split out the `seconds_per_step()` to `steps_per_second()` change as requested.

djc:
@chris-laplante have opinions on this stuff? I've felt like #394 has been the biggest open issue and this feels like a good step forward.

afontenot:
@djc I've fixed all the issues I had with the code, so I've marked it ready for review. Happy to make additional changes you want to see. Wasn't totally sure how you wanted the comments on the weight function formatted - I made them doc comments but code comments would maybe make more sense, as these are implementation details.

The remaining nit I'd like to poke at is trying to extract the debiasing from the `update` function so it doesn't have to be done every time. I think it should be possible to do it only once, when generating an estimate (even for the double EWA), but it's mathematically more complicated. This could always wait for a future PR.

chris-laplante:
> @chris-laplante have opinions on this stuff? I've felt like #394 has been the biggest open issue and this feels like a good step forward.

I've been slammed at work, but I will try to take some time later this week to go over this in detail.

djc:
@chris-laplante what's the chance that you'll get to it this week? Otherwise let's just go ahead and I'll take care of it and you can follow up later?

src/state.rs
djc: > I think there are two separate questions here:
> 
> 1. What should the estimator return when a value is requested but it is in the uninitialized state?
> ..
> I don't really have an opinion on (1) so I'm happy to have it do whatever you like. If you'd like to have an uninitialized estimator return `None`, that's fine with me.

Having an unitialized estimator return `None` sounds good to me in theory at least, though it might depend a bit on how much downstream complexity that requires to make work? Off the top of my head it seems like it might not be too much and would force us to write something like `N/A` in the progress bars, which seems good?

> 2. What the should original (uninitialized) _internal_ value of the estimator be?
> 
> Because the weight of data in the estimator is always positive (there's no horizon at which the weight = 0), the only sensible answer to (2) is 0. That's because you have no way to pick the right value ahead of time, so you will want to divide out the effect of the original estimator value (I call this debiasing). This is much easier to do if it is 0, because then you just divide the naive estimate by `weight(age)`.

Fair enough.

> BTW in case it's unclear, after the first `update` the estimator value _will_ be equal to the value of the first update (for both smoothing and double smoothing), it's just that prior to the first update there's not a correct rate to return.

Okay, good!

src/state.rs
```diff
@@ -370,80 +371,109 @@ impl TabExpandedString {
...
+        self.smoothed_steps_per_sec =
+            self.smoothed_steps_per_sec * weight + new_steps_per_second * (1.0 - weight);
 
-    pub(crate) fn reset(&mut self, now: Instant) {
-        self.pos = 0;
-        self.full = false;
-        self.prev = (0, now);
-    }
+        // Get an unbiased estimate of `smoothed_steps_per_sec` to serve as the data source for the
+        // double smoothed estimate. See comment on debiasing in `steps_per_second` for details.
```
djc, line 432: Even after consuming the `steps_per_second` documentation, I feel like "unbiased" and "debias" are bit too jargony/abstract/vague for me. Could you restate the relevant commits in more concrete/simpler terms?

afontenot: Sure; I'm doing my best to make it as comprehensive as possible without overloading the comments with paragraphs of information. Tell me what you'd like to see in the comments, and I'll do my best to make some changes.

Here's what's happening:

An exponentially weighted estimate is a single value that is iteratively updated over time. At each update, the previous value of the estimate is re-weighted according to its age, with new data receiving the rest of the weight. The raw value of the estimate therefore always reflects the *assumption* that all previous data in the estimate has been weighted according to its age. This is a false assumption.

When the first bit of data comes in, e.g. at t=1, then because the update process is iterative, the previous value of the estimator (which is arbitrary) will receive a weight appropriate for data that has aged by one second. With the default exponent weight of 15s, this means that the arbitrary initial estimator value will have a weight of ~86% in the stored estimate value. This is wrong - it should have weight zero in the estimate - because whatever value the estimator has before t=0 is arbitrary.

Another way of expressing this is to say that a naive iteratively updated exponentially weighted estimator is *biased* towards its initial value.

Fortunately, there is a very easy way to correct for this bias, and return a corrected (debiased) estimate. Included in the naive (biased) estimate is an initial value `x` with a total weight of `estimator_weight(delta_t_start)`. Thinking creatively, if we set the initial value of the estimate to 0, then removing the weight that this value has in the estimate is as trivial as dividing the estimate by the weight of all the data *since* `t=0`! So that's what we do. Whenever we need an estimate that treats the initial estimator value as having zero weight, we *debias* the raw estimator value.

-------------------

Do you think that sufficiently explains the idea? I could include the entire text in a comment if you think it's worth it.

-------------------

Here's a graph of the weight that the initial value has for an estimate of age `t`:

initial_value_weight

As you can see, there's no point at which this weight equals zero, but debiasing is most useful in the first 10-15 seconds. Otherwise you start with a value that is potentially very wrong, and slowly correct it. (Or, following a suggestion you made in a previous comment, you set the initial value of the estimator to be equal to the value of the first piece of data that comes in. But this first piece of data is usually *very* biased in my experience with file transfers, and getting a true estimate by dividing out the initial weight is cheap enough that it makes sense to just do that.)

afontenot:
@djc I'd like to replace the `self.prev` tuple with independent `self.prev_steps` and `self.prev_time` for better readability. That okay with you?

djc:
Yup, go ahead.

afontenot:
> Yup, go ahead. Please make a separate commit that only does that, though.

I've now done so, and force pushed so that it is grouped with the other refactoring change before my exponential weighting work.

I also added some more substantial commentary on debiasing to the `ewa` commit, so let me know if you think that reads clearly enough now.

chris-laplante:
> @chris-laplante what's the chance that you'll get to it this week? Otherwise let's just go ahead and I'll take care of it and you can follow up later?

Sorry, was on vacation last week. I will follow up with it later. Thanks for understanding!

chris-laplante:
I'm no statistician, but I looked this over and I think it makes sense. Your explanation of "debiasing" also make sense. I do also wonder if there is a better term than "debias" we could use - I tried searching for it in the context of exponential smoothing and it doesn't seem to come up in literature much. But I also don't think it's a big deal now that you've added plenty of comments to explain the approach. This definitely seems like a great step forward, so thanks again for your work!

src/state.rs
djc: I think adding a lot of what you've written here as comments is exactly right. I still feel like the use of "debias" is jargony and maybe unnecessary if we can use something less specific like "corrected" instead, but I'm also fine with keeping debias if you think it's better for the text you're writing here.

Looking forward to merging this soon!

afontenot:
> I do also wonder if there is a better term than "debias" we could use - I tried searching for it in the context of exponential smoothing and it doesn't seem to come up in literature much.

I agree that "debias" is not an especially accurate description of what's happening. Strictly speaking, the problem with the naive estimate is that it assumes a non-zero weight for the initial value of the estimator, not that it is "biased" towards that initial value.

Maybe a better way of talking about this would be to call debiased estimates "normalized estimates" instead. After all, if $t_0$ is the current time, with positive time going back into the past, the estimate $E$ is a weighted average of samples for some weighting function $W(t)$:

```math

E = (W(t_0) - W(t_1)) \cdot s_1 + (W(t_1) - W(t_2)) \cdot s_2 + \ldots + (W(t_{f-1}) - W(t_f)) \cdot s_f

```

where $s_f$ is the first collected sample.

This makes the problem fairly obvious. The sum of the weights in our weighted average isn't 1! As you can see from [my plot of the weighting function](
https://user-images.githubusercontent.com/3870410/241409733-f2f80395-7b06-4105-b998-c2a164d199fa.png) the weights in the weighted average will only sum to 1 if the samples continue back into the infinite past.

Since we set the initial value of the estimator to 0, rather than getting a bogus estimate, all the weights for times prior to $t_f$ are canceled out instead. So we have an "underweight" weighted average, and need to [normalize](https://en.wikipedia.org/wiki/Weighted_arithmetic_mean) it so that the total weight is 1. Simply dividing by the total weight in the estimator, we retain the desired weight ratio between samples of various ages!

```math

E_{norm} = \frac{E}{W(t_0) - W(t_f)}

```

@chris-laplante @djc Do you like this way of describing the concepts better? I can rewrite my explanatory comment to use the language of normalization rather than calling this debiasing. Or we could just call it a "correction", though that's a little vague.

Also, please let me know if there's anything you'd still like to see, as I think this is otherwise ready to be merged.

<details>

<summary>Side note on the statistical literature and why I initially chose the "debiasing" language:</summary><br>

I developed the approach here independently; a quick skim of the literature suggests that this is usually discussed as an "initialization" problem. The assumption is that the estimator *has* to take some initial value, and so you figure out a method for doing so. Either you just use the first sample, or the sample mean, or a MSE minimizing value, etc etc.

Trying to set a reasonable initial value like this probably makes sense when doing a post-hoc analysis on data that *has* an (unknown) value before `t=0`, but makes less sense when producing a live estimate where there *is* no value before t=0 (because that point marks the start of `step` production).

After developing this method, I went looking for anyone who had taken a similar approach to the problem, and found this one StackOverflow [answer](https://stackoverflow.com/a/23617678) which handles the initialization issue in the same way. I borrowed the "debiasing" language from that answer, as I hadn't come up with a better way to describe it.

</details>

afontenot:
I've pushed a new commit that does as I suggested in the comment above, and hopefully includes enough commentary to clearly justify the behavior. I kept it as a second commit for now so you can see the change, but I'll squash it into the previous commit and force push if you want to clean up the history before merging.

djc (approved):
Looks great! If you squash these last two commits, I'll be happy to merge this.

Would you mind adding a commit that bumps the version to 0.17.5?

afontenot:
> Looks great! If you squash these last two commits, I'll be happy to merge this.
> 
> Would you mind adding a commit that bumps the version to 0.17.5?

Done and done. Needed to rebase because the PR was based on 0.17.3.

djc:
Thanks for all your work on this! 0.17.5 has made it out to crates.io.

PR #540, opened 2023-05-11
Handle newline in msg and empty msg

RDruon:
Using `println()` with a new line at the beginning of the message trigger some artifacts, see below:

`newline_artifact.rs`
```
use indicatif::ProgressBar;

fn main() {
    let pb = ProgressBar::new(1024);
    pb.inc(1);
    pb.println("\nShould be a new line before this message and no duplicate progress bar");
}
```
Output on `main`:
image
Output with this patch:
image

-----

Using `println()` with an empty string does not trigger new line:
`empty_msg.rs`
```
use indicatif::ProgressBar;

fn main() {
    let pb = ProgressBar::new(1024);
    pb.inc(1);
    pb.println("");
    pb.println("There should an empty line before this message");
}
```
Output on `main`:
image

Output with this patch:
image

New test in `render.rs` showcase both issue

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -496,11 +496,15 @@ impl DrawState {
         let len = self.lines.len();
         let mut real_len = 0;
         for (idx, line) in self.lines.iter().enumerate() {
-            // Calculate real length based on terminal width
-            // This take in account linewrap from terminal
-            real_len +=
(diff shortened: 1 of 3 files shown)


djc (approved):
This makes sense to me, modulo the formatting issue found in CI.

RDruon:
> This makes sense to me, modulo the formatting issue found in CI.

I fixed the format issue

djc (approved):

PR #548, opened 2023-06-02
Add builder-like `with_finish` method to `ProgressBarIter`.

### LINKED ISSUE #545: Feature Request: Add builder-like `with_finish` method to `ProgressBarIter` to set finish behavior
`ProgressBarIter` has a `with_style` method that lets you set the progress bar's style. But if you want to change the finish behavior, i.e. whether the progress bar remains on screen or is erased when it completes, you have to create a new `ProgressBar` and pass it to `progress_with`.

To keep the progress bar on screen when it finishes, I can do the following:

```
let v = vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
v.iter()
        .progress_with(ProgressBar::new(v.len() as u64).with_finish(ProgressFinish::AndLeave));
```

However, it's a bit verbose and needs to explicitly pass the length of the vector to `ProgressBar::new`.

I think it would be convenient to have the more concise form:

```
v.iter().progress().with_finish(ProgressFinish::AndLeave)
```

If this is acceptable from an API perspective, I can write the PR.

matthewhchan:
This lets you more concisely set the finish behavior of the progress bar when using `ProgressBarIter`.

For example, to show a progress bar and keep it displayed when it finishes, you can do

```
v.iter().progress().with_finish(ProgressFinish::AndLeave)
```

instead of

```
v.iter()
        .progress_with(ProgressBar::new(v.len() as u64).with_finish(ProgressFinish::AndLeave));
```

Closes #545

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -13,6 +13,7 @@ use tokio::io::{ReadBuf, SeekFrom};
 
 use crate::progress_bar::ProgressBar;
 use crate::style::ProgressStyle;
+use crate::state::ProgressFinish;
 
 /// Wraps an iterator to display its progress.
 pub trait ProgressIterator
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

PR #550, opened 2023-06-12
Add `decimal_bytes_per_sec` state formatter

Hirevo:
This PR simply adds a `decimal_bytes_per_sec` state formatter.  

Since `bytes_per_sec` and `binary_bytes_per_sec` are both formatted using power-of-two units, it felt like a `decimal_bytes_per_sec` formatter (for using SI units) was missing.  

This follows in the same spirit as the `bytes` and `total_bytes` formatters, which both have `binary` and `decimal` variants.

This PR also updates the docs to mention these `binary` and `decimal` variants being available, which were missing even for `bytes` and `total_bytes`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -175,9 +175,17 @@
 //! * `len`: renders the amount of work to be done as an integer
 //! * `human_len`: renders the total length of the bar as an integer, with commas as the thousands
 //! separator.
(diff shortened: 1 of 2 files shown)


djc:
I'm not that inclined to merge more formatters, given that we now have an extension point for this in the `ProgressTracker` trait.

PR #551, opened 2023-06-22
feat: add futures::Stream support

aatifsyed:
I've mostly just copied the existing patterns for iterator and asyncread

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ rust-version = "1.62.1"
 
 [dependencies]
 console = { version = "0.15", default-features = false, features = ["ansi-parsing"] }
+futures-core = { version = "0.3", default-features = false, optional = true }
 number_prefix = "0.4"
 portable-atomic = "1.0.0"
 rayon = { version = "1.1", optional = true }
@@ -27,6 +28,7 @@ clap = { version = "4", features = ["color", "derive"] }
 once_cell = "1"
 rand = "0.8"
 tokio = { version = "1", features = ["fs", "time", "rt"] }
+futures = "0.3"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 instant = "0.1"
@@ -35,6 +37,7 @@ instant = "0.1"
 default = ["unicode-width", "console/unicode-width"]
 improved_unicode = ["unicode-segmentation", "unicode-width", "console/unicode-width"]
 in_memory = ["vt100"]
+futures = ["dep:futures-core"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ clap = { version = "4", features = ["color", "derive"] }
 once_cell = "1"
 rand = "0.8"
 tokio = { version = "1", features = ["fs", "time", "rt"] }
-futures = "0.3"
+futures = "0.3" # so the doctest for wrap_stream is nice
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 instant = "0.1"
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -277,6 +277,26 @@ impl<W: tokio::io::AsyncBufRead + Unpin + tokio::io::AsyncRead> tokio::io::Async
     }
 }
 
+#[cfg(feature = "futures")]
+#[cfg_attr(docsrs, doc(cfg(feature = "futures")))]
+impl<S: futures_core::Stream + Unpin> futures_core::Stream for ProgressBarIter<S> {
+    type Item = S::Item;
+
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let item = std::pin::Pin::new(&mut this.it).poll_next(cx);
+        match &item {
+            std::task::Poll::Ready(Some(_)) => this.progress.inc(1),
+            std::task::Poll::Ready(None) => this.progress.finish_using_style(),
+            std::task::Poll::Pending => {}
+        }
+        item
+    }
+}
+
 impl<W: io::Write> io::Write for ProgressBarIter<W> {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         self.it.write(buf).map(|inc| {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version = "1.62.1"
 
 [dependencies]
 console = { version = "0.15", default-features = false, features = ["ansi-parsing"] }
+futures = { version = "0.3", default-features = false, optional = true }
```
djc, line 17: If we only need `Stream`, can we depend on futures-core instead?

src/iter.rs
```diff
@@ -277,6 +277,28 @@ impl<W: tokio::io::AsyncBufRead + Unpin + tokio::io::AsyncRead> tokio::io::Async
...
+    type Item = S::Item;
+
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let item = std::pin::Pin::new(&mut this.it).poll_next(cx);
+
+        if item.is_ready() {
```
djc, line 292: This seems wrong? As documented `poll_next()` will yield `Ready` many times, with the `Ready` including `Some` while still going and then yielding `Ready(None)` when the stream has been exhausted. So I think this should be:

```rust
match &item {
    Poll::Ready(Some(_)) => this.progress.inc(1),
    Poll::Ready(None) => this.progress.finish_using_style(),
    Poll::Pending => {}
}
```

src/progress_bar.rs
```diff
@@ -511,6 +511,30 @@ impl ProgressBar {
         }
     }
 
+    #[cfg(feature = "futures")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "futures")))]
```
djc, line 515: Please put the `cfg()` attributes below the docstring.

aatifsyed: Ah, sorry, I was just copying the `wrap_async_read` blocks, would you like me to change the above blocks to match

src/iter.rs
aatifsyed: Haha, that's a bit embarassing - I'll make that change

src/progress_bar.rs
djc: Would be nice -- please do so in a different commit.

Cargo.toml
```diff
@@ -27,6 +28,7 @@ clap = { version = "4", features = ["color", "derive"] }
 once_cell = "1"
 rand = "0.8"
 tokio = { version = "1", features = ["fs", "time", "rt"] }
+futures = "0.3"
```
djc, line 31: This doesn't seem to be used?

src/progress_bar.rs
```diff
@@ -511,6 +511,30 @@ impl ProgressBar {
...
+    /// let stream = pb.wrap_stream(stream::iter(1..=10));
+    /// assert_eq!(
+    ///     stream.count().await,
+    ///     10,
+    /// );
+    /// # }
+    /// ```
+    #[cfg(feature = "futures")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "futures")))]
+    pub fn wrap_stream<S: futures_core::Stream>(&self, stream: S) -> ProgressBarIter<S> {
```
djc, line 531: Style nit: would prefer `impl futures_core::Stream` to avoid the indirection to a type argument here.

djc (approved):
Thanks, this looks much better!

Cargo.toml
aatifsyed: It's for the doctest - I'd imagine most users import `futures`, not `futures_core` 

```
/// use futures::stream::{self, StreamExt};
```

Happy to change if needed!

djc: Oh, I'm okay with using `futures_core` for the doctest.

aatifsyed: Whoops, I think I'd also have to change the doctest, because I wouldn't be able show `futures::stream::iter`:
```rust
# use indicatif::ProgressBar;
use futures::stream::{self, StreamExt};
# async fn test() {
let stream = pb.wrap_stream(stream::iter(1..=10));
assert_eq!(
    stream.count().await,
    10,
);
# }
```

djc: Ah yeah, that's annoying. Okay, can you just add a comment in the `Cargo.toml` for the futures dev-dep why it's there?

src/progress_bar.rs
aatifsyed: I've applied this change, but think I will revert. As the code currently stands, we lose the type of the streamed item.
```rust
    pub fn wrap_stream(
        &self,
        stream: impl futures_core::Stream,
    ) -> ProgressBarIter<impl futures_core::Stream> {
        ProgressBarIter {
            progress: self.clone(),
            it: stream,
        }
    }

```

We've not told `rustc` that the item is the same, so this code doesn't compile:
```rust
fn main() {
    use futures::StreamExt;
    let collected = futures::executor::block_on(
        indicatif::ProgressBar::new(100)
            .wrap_stream(futures::stream::iter(0..100))
            .inspect(|_| std::thread::sleep(std::time::Duration::from_millis(100)))
            .collect::<Vec<usize>>(),
    );
    println!("{collected:?}")
}
```
```
error[E0277]: the trait bound `std::vec::Vec<usize>: std::iter::Extend<<impl futures::Stream as futures::Stream>::Item>` is not satisfied
   --> examples/stream.rs:4:9
    |
3   |       let collected = futures::executor::block_on(
    |                       --------------------------- required by a bound introduced by this call
4   | /         indicatif::ProgressBar::new(100)
5   | |             .wrap_stream(futures::stream::iter(0..100))
6   | |             .inspect(|_| std::thread::sleep(std::time::Duration::from_millis(100)))
7   | |             .collect::<Vec<usize>>(),
    | |____________________________________^ the trait `std::iter::Extend<<impl futures::Stream as futures::Stream>::Item>` is not implemented for `std::vec::Vec<usize>`
    |
```

So we'd at least need to write this:
```rust

    pub fn wrap_stream<I>(
        &self,
        stream: impl futures_core::Stream<Item = I>,
    ) -> ProgressBarIter<impl futures_core::Stream<Item = I>> {
```

Which means we might as well keep the `<S: futures_core::Stream>` bound.

In addition, not erasing the type might be important - maybe the `impl Stream` is also `impl Send`, but we force `rustc` to forget that fact.

djc: Ahhh sorry, yes, because we're reusing the argument type for the return type this doesn't work.

djc (approved):

Issue #485, opened 2022-10-08
Unnecessary hard coded precision in `HumanFloatCount`


ReagentX:
The following is impossible to disable:

https://github.com/console-rs/indicatif/blob/14b5ef279e8de045912a449676ccb32f7ce2d37f/src/format.rs#L157

This precision should be customizable. There is also nothing in the [template docs](https://docs.rs/indicatif/latest/indicatif/index.html#templates) that explain this.

djc:
Sure, would you be able to submit a PR?

PR #552, opened 2023-06-24
Adding method to specify floating-point precision

cloppingemu:
Addresses #485

Precision can now be specified for percent, per_sec, elapsed_precise and eta_precise. Default precision values for percent, per_sec, elapsed_precise and eta_precise are 0, 4 (previous default), 0 and 0 respectively.

Precision can specified by the `-` separator as `percent:-1` or `per_sec:-2`.


djc:
I'm actually no longer sure I'm interested in adding a feature for this. This breaks compatibility, uses an unintuitive marker character (`-`, presumably since `.` is already used for the alternate template style?) and we now have support for custom `ProgressTracker` instances where you can fully customize the desired behavior.

cloppingemu:
Fair enough.

Issue with the marker was unavoidable because of the exisiting use of the more obvious choice. Open to suggestions for choice of character if we decide to merge in some version of the proposed changes.

I'm also surprised it is breaking compatibility. I'll have to have have a look at the CI actions.

I'll have a look at ProgressTracker and if it is easy and self-contained enough we could MAYBE just add an example for it instead.

djc:
It's breaking because you added a field to `FormattedDuration` and `HumanFloatCount`.

I really don't want to add any more examples -- there are too many already. However, better documentation for `ProgressTracker` is probably warranted -- I don't think it's very discoverable right now.

cloppingemu:
Would agree with 
> better documentation for ProgressTracker is probably warranted -- I don't think it's very discoverable right now.

Just going through the style.rs now to see how `ProgressTracker` could be used to customize the printed message. I'll close this PR in the meantime. I'll be back with another using `ProgressTracker`.

Issue #554, opened 2023-06-27
Crash if progress stops


conradoplg:
If progress stops and an ETA is being displayed, it will eventually crash after overflowing.

This modified `download.rs` shows this. It crashes after some minutes.

```
use std::thread;
use std::time::Duration;
use std::{cmp::min, fmt::Write};

use indicatif::{ProgressBar, ProgressState, ProgressStyle};

fn main() {
    let mut downloaded = 0;
    let total_size = 231231231;

    let pb = ProgressBar::new(total_size);
    pb.set_style(ProgressStyle::with_template("{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})")
        .unwrap()
        // .with_key("eta", |state: &ProgressState, w: &mut dyn Write| write!(w, "{:.1}s", state.eta().as_secs_f64()).unwrap())
        .progress_chars("#>-"));

    let mut packet = 1;
    while downloaded < total_size {
        let new = min(downloaded + packet, total_size);
        packet = packet.saturating_sub(1);
        downloaded = new;
        pb.set_position(new);
        thread::sleep(Duration::from_millis(12));
    }

    pb.finish_with_message("downloaded");
}
```

I'll open a PR shortly with a possible solution.

teor2345:
Closed by #555?

djc:
Yup, thanks.

PR #555, opened 2023-06-27
fix overflow crash when displaying ETA

conradoplg:
Possible fix for https://github.com/console-rs/indicatif/issues/554, but I wasn't thorough checking for other possible sources of overflow (in particular in the ETA calculation itself, though from a glance it seems OK)

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -77,7 +77,7 @@ impl fmt::Display for HumanDuration {
         for (i, &(cur, _, _)) in UNITS.iter().enumerate() {
             idx = i;
             match UNITS.get(i + 1) {
-                Some(&next) if self.0 + next.0 / 2 >= cur + cur / 2 => break,
+                Some(&next) if self.0.saturating_add(next.0 / 2) >= cur + cur / 2 => break,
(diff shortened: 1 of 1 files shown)


src/format.rs
```diff
@@ -77,7 +77,7 @@ impl fmt::Display for HumanDuration {
         for (i, &(cur, _, _)) in UNITS.iter().enumerate() {
             idx = i;
             match UNITS.get(i + 1) {
-                Some(&next) if self.0 + next.0 / 2 >= cur + cur / 2 => break,
+                Some(&next) if self.0.saturating_add(next.0 / 2) >= cur + cur / 2 => break,
```
djc, line 80: Do we also need `saturating_add()` for `cur` here?

djc (commented):
Thanks for looking at this!

src/format.rs
conradoplg: I don't think so, `cur` comes from a list of constants and none of them make it overflow.

djc (approved):

djc:
Right, thanks!

PR #562, opened 2023-07-18
Add missing space before unit

Error-42:
*(I'm new to contributing, please be pedantic (but not opinionated) about the way I contribute.)*

The storage unit without prefix (B, but not KB, KiB...) is not separated from the number by a space, which is incorrect ([see 5.3.3 here](https://www.bipm.org/documents/20126/41483022/si_brochure_8.pdf)) and inconsistent with the formatting of prefixed units (KB, KiB...) in this library.

This is a ***breaking change*** and requires the modification of one test.


djc (approved):
This seems good for consistency if nothing else. @chris-laplante, what do you think?

chris-laplante:
> This seems good for consistency if nothing else. @chris-laplante, what do you think?

Makes sense to me.

PR #563, opened 2023-07-19
Only render as many lines as the terminal can actually display

### LINKED ISSUE #496: If the number of progress bar exceed terminal height, newlines appears instead of clearing itself.
Similar to #144 , if progress bars exceeds terminal's height i.e if there are more progress bars than it is able fit in visible at once, newlines are created.
Fix provided in #144 doesn't help either.

# Reproduction steps:

```toml
indicatif = "0.17.2"
```

```rust
use std::thread;
use std::time::Duration;

use indicatif::{MultiProgress, ProgressBar, ProgressStyle};

fn main() {
    let m = MultiProgress::new();
    let sty = ProgressStyle::with_template(
        "[{elapsed_precise}] {bar:40.cyan/blue} {pos:>7}/{len:7} {msg}",
    )
    .unwrap()
    .progress_chars("##-");

    m.println("starting!").unwrap();

    let mut h = vec![];

    for _ in 0..50 {
        let pb = m.add(ProgressBar::new(128));
        pb.set_style(sty.clone());

        let m_clone = m.clone();
        let h1 = thread::spawn(move || {
            for i in 0..128 {
                pb.set_message(format!("item #{}", i + 1));
                pb.inc(1);
                thread::sleep(Duration::from_millis(50));
            }
            m_clone.println("pb1 is done!").unwrap();
        });

        h.push(h1);
    }

    for k in h {
        k.join().unwrap();
    }

    // m.clear().unwrap();
}
```

## Use this code and run it 
- in a terminal with height tall enough  that it is able to display all progress bars at once, no issues.
-  in a terminal with reduced height so that all progress bars are not visible at once,  newlines occurs.

[Example](https://user-images.githubusercontent.com/11157420/205418363-137b5d65-f6d7-4e57-8172-9b0860583edf.webm)

oli-obk:
fixes #496


src/record.rs
oli-obk, line 1: This causes very verbose tests that are hard to update. I could add a blessing scheme that just dumps the expected output in files and updates them automatically locally, but checks them in CI?

chris-laplante:
@oli-obk thank you for working on this, I appreciate it greatly! I will review today or tomorrow :)

djc:
@oli-obk I'm curious about your use case for indicatif?

oli-obk:
I am rewriting compiletest-rs (https://github.com/oli-obk/ui_test). Similar to regular Rust ui tests, it currently has a `--quiet` mode where it prints one `.` per test that was run. I have a [WIP PR](https://github.com/oli-obk/ui_test/pull/108) that switches to indicatif displaying a progress bar. But when I added displaying a spinner per currently running test below that, this became an issue, as the test runner uses all cores to run one test per core, and I tested this on a 96 core cloud machine 😆

oli-obk:
Note that this PR appears to not fully fix the issue (I got the progress bar moved upwards and then everything fixed itself for about 15 times while running >200 tests). I have tried to debug that with the dummy terminal I added, but I think that nothing is actually wrong, and this may just be an issue because I'm running everything across ssh and there are some delays sometimes

src/record.rs
oli-obk: meaning I'd add `ui_test` as a test runner to this crate 😆

src/record.rs
```diff
@@ -0,0 +1,296 @@
+use std::convert::TryFrom;
+use std::fmt::Debug;
+use std::sync::{Arc, Mutex};
+
+use crate::TermLike;
+
+/// A Terminal that records all writes and curor moves to it.
+/// Useful for debugging operations on terminals.
+#[cfg_attr(docsrs, doc(cfg(feature = "in_memory")))]
```
chris-laplante, line 9: Possible typo: should this be 'record' instead of 'in_memory'?

src/record.rs
```diff
@@ -0,0 +1,296 @@
...
+#[derive(Debug, Clone)]
+pub struct RecordTerm {
+    state: Arc<Mutex<RecordTermState>>,
+}
+
+impl RecordTerm {
```
chris-laplante, line 15: In general, I wonder if it makes more sense to just merge the functionality of `RecordTerm` into `InMemoryTerm`?

src/record.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.write_str(s)?;
+        let mut lock = self.state.lock().unwrap();
+        lock.history.push(Move::NewLine);
+
+        lock.cursor_position.1 = 0;
+        lock.cursor_position.0 += 1;
+        Ok(())
+    }
+
+    fn write_str(&self, s: &str) -> std::io::Result<()> {
```
chris-laplante, line 85: I would maybe add a comment here that this doesn't handle escape sequences. The difficulty in doing so is what ultimately led me to use vt100 for `InMemoryTerm`.

src/record.rs
chris-laplante: I don't think the verbose tests are a big deal. It's unfortunate, but kind of a fact of life for a crate like this :).

src/record.rs
chris-laplante: (either way, as per comment on lib.rs this could probably go away anyway)

chris-laplante (commented):
Looks really good! I can tell you've put a ton of work into this :), and we (@djc and I) really appreciate it. I think just a couple small things to clean up.

I tested the multi-many example on my own laptop and it worked fine, even as I was resizing the terminal.

src/record.rs
oli-obk: Yea, that makes sense. I mostly didn't know whether `InMemoryTerm` had any non-test usage

src/record.rs
oli-obk: Oh the verbosity is fine, but the updating is a pain.

I think I have an idea for how to make updating them easier without requiring a new tool. I'll try it out and we'll see if that is better.

src/record.rs
oli-obk: ah... hmm yea. that's an issue

chris-laplante: It's OK, this is just a test mechanism after all. Don't try to implement escape sequence handling - it is nothing but pain, btw.

Cargo.toml
```diff
@@ -29,6 +29,7 @@ once_cell = "1"
 rand = "0.8"
 tokio = { version = "1", features = ["fs", "time", "rt"] }
 futures = "0.3" # so the doctest for wrap_stream is nice
+pretty_assertions = "1.4.0"
```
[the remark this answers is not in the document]
djc: Unless we are sure we are only relying on features in pretty_assertions that were present in 1.0.0, it's common to add new dependencies at their current versions.

oli-obk:
I made the tests simpler (they are just strings now that you can mostly copy out of the assertion failure message).

All other reviews should also have been addressed

Cargo.toml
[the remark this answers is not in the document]
chris-laplante: @oli-obk ah, yes my mistake. Please revert to 1.4.0. 
@djc Gotcha - I was under the impression that looser deps were better since you automatically got minor bugfixes on rebuild. But I am still a relative Rust noob :)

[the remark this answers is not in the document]
djc: 1.4.0 is the same level of loose in this context: Cargo pulls in semver-compatibles by default, any 1.x release with x >= 4 is compatible.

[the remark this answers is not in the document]
chris-laplante: TIL thanks!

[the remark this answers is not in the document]
oli-obk: > Unless we are sure we are only relying on features in pretty_assertions that were present in 1.0.0,

I am sure that we are only relying on 1.0 features, and I set the commit message to reflect that.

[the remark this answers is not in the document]
oli-obk: Now that I'm back on a computer, I removed the last commit.

examples/multi-many.rs
```diff
@@ -0,0 +1,36 @@
+use std::thread;
```
djc, line 1: We already have a lot of examples. Would it be feasible to change one of the existing examples instead of adding yet another one? All of these amount to relatively low-value extra code that needs to be maintained.

src/draw_target.rs
```diff
@@ -514,7 +514,7 @@ impl DrawState {
                 // subtract with overflow later.
                 usize::max(terminal_len, 1)
             };
-            if real_len + diff >= term_height {
+            if real_len + diff > term_height {
```
djc, line 517: I'd prefer for fixes like these to be squashed into the originating commit.

(Also goes for the other changes you've tacked on.)

examples/multi-many.rs
oli-obk: done

bjorn3:
Does this handle shrinking the vertical size of the terminal? When doing so the terminal is temporarily 1 line smaller than indicatif expected when rendering until the next time the size is queried and everything is rendered again. Leaving a line as buffer may help with this.

oli-obk:
> Does this handle shrinking the vertical size of the terminal? When doing so the terminal is temporarily 1 line smaller than indicatif expected when rendering until the next time the size is queried and everything is rendered again. Leaving a line as buffer may help with this.

I tried it out and resizing is indeed an issue. I don't think leaving an extra line as buffer is going to significantly help us here, and I'd rather look into this in general

oli-obk:
> Thanks!
> 
> Minor nit if you want to clean it up: some of the commit messages are now slightly wrong respective to the commit contents: "Add a test terminal" is more like "Add history to the in-memory terminal" and "Edit an existing test" is more like "Edit an existing example".

done

djc:
@chris-laplante do you want to take another look? Would be good to release this along with the other recent fixes.

chris-laplante:
> @chris-laplante do you want to take another look? Would be good to release this along with the other recent fixes.

Yes, I'll take a quick look this morning.

chris-laplante:
Sorry for the delay. Looks great, thanks again!

PR #568, opened 2023-07-30
Fixed progress bar not suspending multiprogress bar

Vrajs16:
When suspending a progress bar which is already inside a multiprogress bar, it should suspend the multiprogress bar.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -161,13 +161,17 @@ impl BarState {
     }
 
     pub(crate) fn suspend<F: FnOnce() -> R, R>(&mut self, now: Instant, f: F) -> R {
-        if let Some(drawable) = self.draw_target.drawable(true, now) {
-            let _ = drawable.clear();
-        }
+        if let Some((state, _)) = self.draw_target.remote() {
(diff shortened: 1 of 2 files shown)


src/state.rs
```diff
@@ -161,13 +161,17 @@ impl BarState {
     }
 
     pub(crate) fn suspend<F: FnOnce() -> R, R>(&mut self, now: Instant, f: F) -> R {
-        if let Some(drawable) = self.draw_target.drawable(true, now) {
-            let _ = drawable.clear();
-        }
+        if let Some((state, _)) = self.draw_target.remote() {
+            state.write().unwrap().suspend(f, now)
```
djc, line 165: Style nit: please return from this `if`, so that we don't need to have the rest of the function in the `else` block (reducing rightward drift).

djc (commented):
@chris-laplante what do you think? Makes sense to me.

src/state.rs
Vrajs16: Alright Changed it

chris-laplante (approved):
Makes sense to me too, thanks!

Issue #566, opened 2023-07-29
using indicatif and the log crate


djugei:
This is not an issue but instructions on how you can log messages and have a progress bar at the same time.
its an issue that comes up sometimes: https://github.com/console-rs/indicatif/issues/366 https://github.com/console-rs/indicatif/issues/92 https://github.com/console-rs/indicatif/issues/474 https://github.com/console-rs/indicatif/issues/27 possibly https://github.com/console-rs/indicatif/issues/474

You create a global MultiProgress, and then you just need to make your logging implementation, in this example env_logger use that (or rather its println function) as write target. to do so we utilize the pipe crate, but it would also be possible to use a ring buffer or something else:

    let (read, write) = pipe::pipe();
    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
        .target(env_logger::Target::Pipe(Box::new(write)))
        .init();

    let multi = Arc::new(MultiProgress::new());

    {
        let m = multi.clone();
        std::thread::spawn(move || {
            let br = std::io::BufReader::new(read);
            for line in br.lines() {
                let line = line.unwrap();
                m.println(&line);
            }
        });
    }

then you simply add all your progress bars to the gobal multi and things just work.

i will probably turn this into an example soon.
it should be somewhat simple to actually have MultiProgress (or a wrapper of it) to implement the [Log](https://docs.rs/log/latest/log/trait.Log.html) trait. i might send a pr for that in a week valve time.

PR #569, opened 2023-07-30
added LogWrapper to bridge indicatif and the log crate

djugei:
this is a much nicer solution than https://github.com/console-rs/indicatif/issues/566 and https://github.com/console-rs/indicatif/pull/567 in my opinion.

this is behind the "log" feature for now.
we could consider making this a default feature, it barely adds any code, does add the log crate as dependency though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ tokio = { version = "1", optional = true, features = ["io-util"] }
 unicode-segmentation = { version = "1", optional = true }
 unicode-width = { version = "0.1", optional = true }
 vt100 = { version = "0.15.1", optional = true }
+log = { version = "0.4.19", optional = true, features = ["std"] }
 
 [dev-dependencies]
 clap = { version = "4", features = ["color", "derive"] }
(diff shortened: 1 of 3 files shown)


chris-laplante:
To be honest, I'm not sure this belongs in indicatif. I think I'd rather see it as a separate crate, e.g. like the other loggers: https://docs.rs/log/latest/log/#available-logging-implementations. I don't have strong feelings though, so I could be convinced otherwise.

djc:
I agree. Would be happy to link to it from the README/documentation, though.

djugei:
yeah i agree, did it here:

https://crates.io/crates/indicatif-log-bridge
https://github.com/djugei/indicatif-log-bridge

chris-laplante:
> yeah i agree, did it here:
> 
> https://crates.io/crates/indicatif-log-bridge https://github.com/djugei/indicatif-log-bridge

Awesome, thanks! Feel free to submit a PR that edits the README to mention the crate.

djugei:
allright i will, but first i gotta polish/add a bit of documentation to it

PR #570, opened 2023-07-31
Add new Integrations section to README, containing a link to indicatif_log_bridge initially

djugei:
discussed in https://github.com/console-rs/indicatif/pull/569

i tried writing an example too, but that leads to a circular dependency (indicatif_log_bridge depends on indicatif) and two "different" versions of MultiProgress being found, so that did not work out sadly.


djc (approved):

djc:
Thanks!

PR #576, opened 2023-08-11
Fix a potential overflow with a saturating add.

### LINKED ISSUE #575: Overflow error
Hi, thanks for your work on `indicatif`, which we've been happily using for years now.

Recently, an overflow bug became apparent to us and has already been fixed (thanks!).  However, I seem to have stumbled across another one.  Here is the link to our CI job as well as the offending line of code:

https://github.com/MWATelescope/Birli/actions/runs/5828640212/job/15806705268#step:4:234
https://github.com/console-rs/indicatif/blob/27c839904e4f60570535adfa72a33e3220fb6d03/src/multi.rs#L293

I'm almost certain the issue stems from using a progress bar somewhere with very little work to do, but I'm sure you agree that these potential issues should be handled, probably like the other overflow bug (i.e. saturating add).  I'm happy to contribute code to fix the bug.

cjordan:
Fixes #575

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -241,7 +241,7 @@ impl MultiState {
             .unwrap_or_default();
 
         // Track the total number of zombie lines on the screen
-        self.zombie_lines_count += line_count;
+        self.zombie_lines_count = self.zombie_lines_count.saturating_add(line_count);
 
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

djc:
The MSRV failure in CI is unrelated, will submit a PR for that separately.

PR #579, opened 2023-08-15
Tick ProgressTrackers before drawing

djc:
@chris-laplante what do you think of this? Seems like a more logical order for things to happen in... (Found after spelunking in the context of #578.)


chris-laplante (approved):
Makes sense to me too

chris-laplante:
@djc I'm working on fixing the linter failure

PR #583, opened 2023-09-02
allow padding HumanBytes Display output

darrell-roberts:
I've updated the `Display` implementation for `HumanBytes` so it can accept formatting parameters.


djc:
So what exactly is this trying to solve? It seems unfortunate that this will need an additional allocation.

darrell-roberts:
The goal was to be able to output `HumanBytes` with formatting parameters.

Ex:

```rust
#[test]
    fn test_human_bytes() {
        use std::fmt::Write;

        let mut s = String::new();
        write!(&mut s, "{:>10}", HumanBytes(10_000)).unwrap();
        assert_eq!(s, "  9.77 KiB");
    }
```

I agree though I would not want to incur an additional allocation cost. If there is a better alternative I'd go with that.

PR #586, opened 2023-09-08
Fix attempt to subtract with overflow (#582)

### LINKED ISSUE #582: Attempt to subtract with overflow when output exceeds terminal height
I am observing `attempt to subtract with overflow` on this line: https://github.com/console-rs/indicatif/blob/40b40d29b6d06ae18c40b829f77e9c43bcebd7af/src/draw_target.rs#L534

`self.lines.len() >= self.orphan_lines_count` holds prior to entering this loop: https://github.com/console-rs/indicatif/blob/40b40d29b6d06ae18c40b829f77e9c43bcebd7af/src/draw_target.rs#L501

And `real_len >= idx` holds at the start of each iteration of that loop.

Those two facts causes me to squint at this `if` block: https://github.com/console-rs/indicatif/blob/40b40d29b6d06ae18c40b829f77e9c43bcebd7af/src/draw_target.rs#L517-L519

In particular, commenting out that `if` block causes my panics to go away.

I don't completely understand the purpose of that `if` block, but maybe it can go away?

(indicatif is a great tool, BTW.)

smoelius:
Fixes #582 

The test `orphan_lines` reproduces the panic. The test `orphan_lines_message_above_progress_bar` tries to ensure that messages are still printed correctly following the fix.

The proposed fix is to effectively ignore `self.orphan_lines_count` many lines from `real_len` when comparing `real_len` to the terminal height.

@oli-obk Would you be able to glance at this? :pray:

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -498,6 +498,7 @@ impl DrawState {
         let len = self.lines.len();
         let mut real_len = 0;
         let mut last_line_filler = 0;
+        debug_assert!(self.orphan_lines_count <= self.lines.len());
         for (idx, line) in self.lines.iter().enumerate() {
             let line_width = console::measure_text_width(line);
             let diff = if line.is_empty() {
@@ -514,7 +515,11 @@ impl DrawState {
                 // subtract with overflow later.
                 usize::max(terminal_len, 1)
             };
-            if real_len + diff > term_height {
+            // Don't consider orphan lines when comparing to terminal height.
+            debug_assert!(idx <= real_len);
+            if self.orphan_lines_count <= idx
+                && real_len - self.orphan_lines_count + diff > term_height
+            {
                 break;
             }
             real_len += diff;
--- a/tests/render.rs
+++ b/tests/render.rs
@@ -1804,13 +1804,15 @@ fn orphan_lines_message_above_progress_bar() {
         let n = 5 + i;
 
         // Test with messages of differing numbers of lines. The messages have the form:
-        // n - 1 newlines followed by the number `n`. The value of n ranges from 5 (less
-        // than the terminal height) to 15 (greater than the terminal height).
-        pb.println(format!("{}{n}", "\n".repeat(n - 1)));
+        // n - 1 newlines followed by n * 11 dashes (`-`). The value of n ranges from 5
+        // (less than the terminal height) to 15 (greater than the terminal height). The
+        // number 11 is intentionally not a factor of the terminal width (80), but large
+        // enough that the strings of dashes should eventually wrap.
+        pb.println(format!("{}{}", "\n".repeat(n - 1), "-".repeat(n * 11)));
 
         // Check that the line above the progress bar is the number `n`.
         assert_eq!(
-            format!("{n}"),
+            format!("{}", "-".repeat(n * 11 % 80)),
             in_mem.contents().lines().rev().nth(1).unwrap(),
         );
     }
--- a/tests/render.rs
+++ b/tests/render.rs
@@ -1780,8 +1780,7 @@ fn orphan_lines() {
 
         let n = 5 + i;
 
-        let msg = std::iter::repeat('\n').take(n).collect::<String>();
-        pb.println(msg);
+        pb.println("\n".repeat(n));
     }
 
     pb.finish();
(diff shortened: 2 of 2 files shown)


djc (approved):
LGTM, thanks for digging into this! Will leave this open for other reviews, feel free to ping me in a week or so.

src/draw_target.rs
```diff
@@ -514,7 +515,11 @@ impl DrawState {
                 // subtract with overflow later.
                 usize::max(terminal_len, 1)
             };
-            if real_len + diff > term_height {
+            // Don't consider orphan lines when comparing to terminal height.
+            debug_assert!(idx <= real_len);
+            if self.orphan_lines_count <= idx
+                && real_len - self.orphan_lines_count + diff > term_height
```
oli-obk, line 521: how does `idx` play into this? It just counts entries, not lines.

smoelius: Please forgive me, I don't know the precise meanings of these terms ("entries" and "lines").

But what I seem to observe is this. Following a call to `println` with a string containing `n` newlines, `self.orphan_lines_count` will be (at least) `n + 1`, and `self.lines` will contain (at least) `n + 1` elements.

The purpose of checking `idx` is see whether all of those orphan lines have been processed.

smoelius: Thanks for clarifying.

> we're trying to keep `real_len < term_height`

Is that really what we want, though? Not `real_len - self.orphan_lines_count < term_height`?

To be more precise, isn't it that we want to keep the vertical space of the displayed progress bars less than `term_height`?

> So I think your issue should also be fixable by comparing against `real_len` instead of `idx`

Here's why that suggestion gives me pause. Suppose we break with `idx < self.orphan_lines_count`. Then some number of orphan lines have not been drawn. In other words, the user called `println`, but some portion of the output has not been shown. Isn't that undesirable?

Wouldn't we want to show those lines (possibly allowing them to scroll off the top of the screen), and then draw just as many progress bars will fit?

oli-obk: ah... right.

In that case, isn't even `idx` causing `println` lines to disappear?
If I understand you correctly, we'll always want to render orphan lines, and ignore them in the height math, acting as if they don't contribute to the terminal space that gets used up.

So... do we actually want `(real_len + diff).saturating_sub(self.orphan_lines_count) > term_height` without any other checks?

djc: @chris-laplante maybe you want to weigh in here as well?

smoelius: > If I understand you correctly, we'll always want to render orphan lines, and ignore them in the height math, acting as if they don't contribute to the terminal space that gets used up.

That's how I think of it, yes.

But I don't see how these follow.
> In that case, isn't even `idx` causing `println` lines to disappear?

> So... do we actually want `(real_len + diff).saturating_sub(self.orphan_lines_count) > term_height` without any other checks?

If we were try to attach "units" to these variables, `self.orphan_lines_count` and `idx` and would count "unformatted lines", but `real_len` would count "formatted (i.e,. narrower than the terminal) lines", correct?

So wouldn't `(real_len + diff).saturating_sub(self.orphan_lines_count)` be a kind of "type error," because it tries to add a formatted line count (`real_len + diff`) to an unformatted line count (`self.orphan_lines_count`)?

chris-laplante: > @chris-laplante maybe you want to weigh in here as well?

I'll try to take a look later this week, sorry, a little short on time for now :(

smoelius: @oli-obk Did you have any further thoughts on this?

oli-obk: I assumed the `orphan_lines_count` was the real printed lines count. If that is not true, then your reasoning is correct and we need a way to fix that.

smoelius: To help increase confidence, I updated the `orphan_lines_message_above_progress_bar` test to generate orphan lines that wrap around the terminal.

From what I can tell, the fix seems to handle those lines correctly.

oli-obk (approved):

chris-laplante:
> To whomever keeps approving the workflows: thank you. 🙏

We gotchu :)

djc (approved):

djc:
@chris-laplante do you want to do a final review before (squash and) merge?

@smoelius thanks for your work in getting this right, and @oli-obk thanks for the reviews!

chris-laplante (approved):
Looks good to me! Thanks for the new tests too :)

PR #592, opened 2023-09-24
documented behaviour of MultiProgress::add

djugei:


djc (approved):
@chris-laplante I believe it makes sense to document this as a guarantee, right?

djc:
Thanks!

chris-laplante (approved):

Issue #597, opened 2023-10-06
Sub-second precision for `elapsed`.


lgarron:
I've been happily using `indicatif` for a few use cases, and just applied it to tracking how long certain dependencies take to build: https://github.com/lgarron/mak

"Unfortunately", I've optimized the builds and tests for one of my projects so well that many of the steps take less than a second.

https://github.com/lgarron/mak/blob/main/readme/demo.gif

Would it be possible to enable sub-second precision for `elapsed` so I can see how long the steps took to run?

I'd be glad for a way to specify 3 digits of precision, since some commands can be in the range to 10ms.

djc:
I'm open to having the logic changed so that it shows milliseconds instead of seconds past some threshold. Are you able to submit a PR for this? I think it might be enough to add millisecond value to `UNITS` here: https://github.com/console-rs/indicatif/blob/main/src/format.rs#L106.

lgarron:
ooh, that definitely was easy!


```
🎯|    3s |   ↙ test-spec-bun
🎯|    5s |   ↙ test-spec-dom
🎯|  17ms |  ↙ test-spec
🎯| 854ms |  ↙ lint-ci
🎯|    7s |  ↙ test-src-tsc
🎯| 353ms |   ↙ build-lib-js
🎯| 953ms |  ↙ test-src-import-restrictions
🎯|  42ms |  ↙ test-src-scripts-consistency
🎯|   9ms | ↙ test-src
🎯| 121ms |  ↙ build-bin
🎯|   14s |  ↙ build-lib-types
🎯| 726ms |   ↙ build-site-twizzle
🎯| 569ms |   ↙ build-site-experiments
🎯|  14ms |  ↙ build-sites
🎯|    8s |  ↙ build-site-docs
🎯|   7ms | ↙ test-build
🎯|    4s |   ↙ test-dist-lib-node-import
🎯|   10s |   ↙ test-dist-lib-node-scramble-all-events
🎯|    4s |   ↙ test-dist-lib-perf
🎯|    4s |   ↙ test-dist-lib-plain-esbuild-compat
🎯|    4s |   ↙ test-dist-lib-build-size
🎯|    2s |   ↙ test-dist-sites-experiments
🎯|   7ms |  ↙ test-dist-lib
🎯|    3s |   ↙ test-dist-bin-shebang
🎯|    3s |   ↙ test-dist-bin-npm-exec
🎯|  21ms |  ↙ test-dist-bin
[...]

PR #598, opened 2023-10-06
Display milliseconds for `elapsed` below 1 second.

lgarron:
Addresses https://github.com/console-rs/indicatif/issues/597

See https://github.com/console-rs/indicatif/issues/597#issuecomment-1751404133 for some thoughts on alternatives.


powerpellet1077:
Just returning from [here](https://github.com/console-rs/indicatif/issues/597#issuecomment-4167510648). Would anyone care to clean up and merge this PR?

PR #599, opened 2023-10-18
docs: Improve formatting and linking.

waywardmonkeys:
Using `clippy::doc_markdown` showed a couple of minor issues with the docs and then the rest were via visual inspection.


djc (approved):
Good stuff, thanks!

PR #600, opened 2023-10-29
Add minimal examples demonstrating Formatters

Chris--B:
This change adds some doc tests to show off what the formatter looks like. This change adds it for just the byte variants:
- `HumanBytes`, which seems to forward to `BinaryBytes`?
- `DecimalBytes`, 1 MB == 1 * 10**6 bytes
- `BinaryBytes`, 1 MiB == 1 * (1 << 20) bytes

I wanted to know what these formatters *look like*, preferably from the docs. This is the simplest way that comes to mind, but I'm open to doing this differently if there's alternatives. I only did the Byte variants with this. I'm happy to work on the others but want to make sure this direction is accepted first.

Here's what my local `rustdoc` docs look like after this change:
image

These also double as tests, and `cargo test` will fail if their behavior changes.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -19,14 +19,47 @@ pub struct FormattedDuration(pub Duration);
 pub struct HumanDuration(pub Duration);
 
 /// Formats bytes for human readability
+///
+/// # Examples
+/// ```rust
+/// # use indicatif::HumanBytes;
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

Issue #601, opened 2023-10-29
per_sec values not retained when iteration exits early with ProgressFinish::Abandon


DanGonite57:
When progression of a progress bar with `.with_finish(ProgressFinish::Abandon)` ends early, the position of the bar is retained at the point that it stopped. However, the per_sec keys (`per_sec`, `bytes_per_sec`, `binary_bytes_per_sec`) are all updated as if the rest of the progress had been completed instantly. I would suggest that these values should also be kept as they were when progression was abandoned.

env:
`rustc 1.73.0 (cc66ad468 2023-10-03)`

cargo.toml:
`indicatif = "0.17.7"`

Arbitrary example:
```rust
use indicatif::{ProgressBar, ProgressFinish, ProgressIterator, ProgressStyle};
use std::{array, thread::sleep, time::Duration};

fn main() {
    let arr: [usize; 10] = array::from_fn(|x| x);

    let progress_bar = ProgressBar::new(10).with_finish(ProgressFinish::Abandon);
    progress_bar.set_style(
        ProgressStyle::default_bar()
            .template(
                "[{elapsed_precise}] {bar:60} {pos:>7}/{len:7} {per_sec:>8} {bytes_per_sec:>8} {binary_bytes_per_sec:>8}",
            )
            .unwrap(),
    );
    arr.iter()
        .map(|x| {
            sleep(Duration::new(1, 0));
            x
        })
        .progress_with(progress_bar)
        .find(|&&x| x == 9);

    let progress_bar = ProgressBar::new(10).with_finish(ProgressFinish::Abandon);
    progress_bar.set_style(
        ProgressStyle::default_bar()
            .template(
                "[{elapsed_precise}] {bar:60} {pos:>7}/{len:7} {per_sec:>8} {bytes_per_sec:>8} {binary_bytes_per_sec:>8}",
            )
            .unwrap(),
    );
    arr.iter()
        .map(|x| {
            sleep(Duration::new(1, 0));
            x
        })
        .progress_with(progress_bar)
        .find(|&&x| x == 4);
}
```

Output:
```
[00:00:10] ████████████████████████████████████████████████████████████      10/10      0.9925/s    0 B/s    0 B/s
[00:00:05] ██████████████████████████████░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░       5/10      1.9868/s    1 B/s    1 B/s
```

djc:
[...]

PR #605, opened 2023-11-20
Use current position for calculating finished progress bar per_sec instead of bar length

DanGonite57:
Fixes https://github.com/console-rs/indicatif/issues/601


djc (approved):
Sorry for the long delay in reviewing this -- I think it makes sense.

For future robustness, please consider submitting another PR that includes a test for this.

Issue #612, opened 2023-12-12
indicatif crashes with assertion failed: self.orphan_lines_count <= self.lines.len() if number of terminal columns is less than 135


bes:
Hello!

The fix for https://github.com/console-rs/indicatif/issues/582 introduced a crash in my code, and I can't figure out why.

My application crashes with this message:

```
thread 'main' panicked at /Users/bes/.cargo/registry/src/index.crates.io-6f17d22bba15001f/indicatif-0.17.7/src/draw_target.rs:501:9:
assertion failed: self.orphan_lines_count <= self.lines.len()
```

If my terminal has enough "width" - 135 columns, it works without crashing
If my terminal has below 135 columns, it crashes!

The number of rows seems to be irrelevant, I've tried everything from 9 to 74 lines.

djc:
Sorry for the regression! Can you try if it works better with the changes from #608?

bes:
@djc nope still crashes unfortunately

```
714/171414thread 'main' panicked at /Users/bes/.cargo/git/checkouts/indicatif-c82e440bcf1e0928/9169539/src/draw_target.rs:501:9:
assertion failed: self.orphan_lines_count <= self.lines.len()
```

smoelius:
@bes Can you share your code, or provide instructions to reproduce?

bes:
@smoelius here is a proof-of-concept https://github.com/bes/indicatif-assert-panic-poc

smoelius:
I can reproduce the crash.

Can you give me a sense for what you expect that program's output to look like (i.e., if everything worked correctly)?

smoelius:
> Can you give me a sense for what you expect that program's output to look like (i.e., if everything worked correctly)?

Oh, I can just run it on a larger width to get that. 🤦

smoelius:
The code in state.rs and multi.rs seem to have different ideas about what `orphan_lines_count` represents.

The code in state.rs seems to regard it a number of `lines` entries:
https://github.com/console-rs/indicatif/blob/8941d5e8f55fe32a629a80872ee20971c00ee53d/src/state.rs#L155

Whereas the code in multi.rs seems to regard it as a visual line count (formerly `real_len`):
https://github.com/console-rs/indicatif/blob/8941d5e8f55fe32a629a80872ee20971c00ee53d/src/multi.rs#L315-L323

[...]

PR #616, opened 2023-12-25
Add `VisualLines` newtype wrapper

smoelius:
An attempt to add a newtype wrapper, as discussed in https://github.com/console-rs/indicatif/issues/612#issuecomment-1853498395 and https://github.com/console-rs/indicatif/issues/612#issuecomment-1860809497.

Notes
1. The wrapper is called `VisualLines` because that was the term introduced in #608, but I am not tied to that name.
2. This PR adds a dependency: [`derive_more`](https://github.com/JelteF/derive_more).
2. I did not modify the `TermLike` trait. It has both "height" and "width" characteristics, and it seemed weird to modify just one.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -1,4 +1,5 @@
 use std::io;
+use std::slice::SliceIndex;
 use std::sync::{Arc, RwLock, RwLockWriteGuard};
 use std::thread::panicking;
 use std::time::Duration;
@@ -6,6 +7,7 @@ use std::time::Duration;
 use std::time::Instant;
 
 use console::Term;
+use derive_more::{Add, AddAssign, Sub};
 #[cfg(target_arch = "wasm32")]
 use instant::Instant;
 
@@ -72,7 +74,7 @@ impl ProgressDrawTarget {
         Self {
             kind: TargetKind::Term {
                 term,
-                last_line_count: 0,
+                last_line_count: VisualLines::zero(),
                 rate_limiter: RateLimiter::new(refresh_rate),
                 draw_state: DrawState::default(),
             },
@@ -84,7 +86,7 @@ impl ProgressDrawTarget {
         Self {
             kind: TargetKind::TermLike {
                 inner: term_like,
-                last_line_count: 0,
+                last_line_count: VisualLines::zero(),
                 rate_limiter: None,
                 draw_state: DrawState::default(),
             },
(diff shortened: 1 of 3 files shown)


chris-laplante:
Thanks! First impressions are that I actually really like this. There are a lot fewer casts than I expected. (I also didn't know about `SliceIndex` - that's pretty cool.)

The real question will be whether or not it makes the code clearer and easier to read though. I'm going to take some time to digest it over the next few days. 

Thanks again for the work!

src/draw_target.rs
```diff
@@ -551,11 +556,48 @@ impl DrawState {
...
+    }
+    pub(crate) fn saturating_add(&self, other: Self) -> Self {
+        Self(self.0.saturating_add(other.0))
+    }
+    pub(crate) fn saturating_sub(&self, other: Self) -> Self {
+        Self(self.0.saturating_sub(other.0))
+    }
+    pub(crate) fn zero() -> Self {
+        Self(0)
+    }
```
djc, line 590: Style nits: let's order these `zero()` (if we're keeping it), the `saturating` methods, then `as_usize()`, each separated by an empty line.

src/draw_target.rs
```diff
@@ -551,11 +556,48 @@ impl DrawState {
...
+}
+
+#[test]
+fn visual_lines_default_is_zero() {
```
djc, line 594: Please move this into the `mod tests` below.

djc (approved):
LGTM modulo nits, thanks for working on this!

src/draw_target.rs
```diff
@@ -72,7 +74,7 @@ impl ProgressDrawTarget {
         Self {
             kind: TargetKind::Term {
                 term,
-                last_line_count: 0,
+                last_line_count: VisualLines::zero(),
```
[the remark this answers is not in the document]
smoelius: > Is there any case where you think that's unreasonable and `zero()` makes for a better alternative?

Not really. And I don't have a strong preference.

src/draw_target.rs
smoelius: Removed `zero` per the [above request](https://github.com/console-rs/indicatif/pull/616#discussion_r1438172468).

smoelius:
@djc I think I addressed all of your comments. Please let me know if there are any I missed.

PR #618, opened 2023-12-30
synchronized output

Funami580:
Added support for synchronized output (https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036). Before, my multi progressbars flickered a lot. But not anymore.

Synchronized output should work on Windows, too, I think. But I don't know how to check whether that terminal feature is supported on Windows, and I don't use Windows anyway, so I added support for Unix only.

I added two new dependencies for Unix: lazy-static and libc. lazy-static could be replaced with something else, if that is preferable.


Cargo.toml
```diff
@@ -31,6 +31,10 @@ tokio = { version = "1", features = ["fs", "time", "rt"] }
 futures = "0.3" # so the doctest for wrap_stream is nice
 pretty_assertions = "1.4.0"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+lazy_static = "1.4.0"
```
smoelius, line 36: Please forgive me for chiming in, but `lazy_static` is effectively [deprecated](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214). Maybe use [`once_cell::sync::Lazy`](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) or [`OnceLock`](https://github.com/rust-lang-nursery/lazy-static.rs/pull/216) instead?

djc: @smoelius exactly what I was going to suggest.

djc:
I think it would be good to support this, but I also think a lot of the supporting code (about detecting and using the sync mechanism) would probably make more sense in the lower-level console crate (which indicatif uses). Maybe start a PR there?

Funami580:
Yes, I can do a separate PR in console. @djc Should the API look like this? 
```rs
let term = Term::stdout();
if term.supports_sync_update() {
	term.begin_sync_update();
}
// do something
if term.supports_sync_update() {
	term.end_sync_update();
}
```
With sync update doing sth like `write_str("\x1b[?2026h")`?

djc:
I would do a guard type that ends the sync on dropping. But, probably good to ask the console maintainer(s) to weigh in on this.

Funami580:
Since the console crate seems a little bit inactive, I made this PR simpler. It now uses once_cell and might work on Windows, too, though I haven't checked (should be checked before merging, though). Note: I added a new method to a public trait which is a breaking change. I'm still not sure, if this is the correct design (`supports_ascii_codes` in the public trait).

src/term_like.rs
```diff
@@ -76,4 +81,44 @@ impl TermLike for Term {
     fn flush(&self) -> io::Result<()> {
         self.flush()
     }
+
+    fn supports_ansi_codes(&self) -> bool {
+        static SUPPORTS_ANSI: OnceCell<bool> = OnceCell::new();
```
Funami580, line 86: I noticed the use of static here is incorrect, and I don't see an easy way around that.

djc:
I pinged a console maintainer on Discord to see if we can get that conversation going. :)

Funami580:
If we were to add a Sync Guard in the console crate, how would we be able to use it in indicatif? I guess via the TermLike trait, but what about the in_memory term then?

chris-laplante:
> If we were to add a Sync Guard in the console crate, how would we be able to use it in indicatif? I guess via the TermLike trait, but what about the in_memory term then?

The in-mem term is just for testing purposes. I don't think we'd need sync guard there necessarily?

chris-laplante:
> Okay, but then what about custom TermLike implementations? indicatif provides the ability to create a ProgressBar via a `Box<TermLike>`. Right now with `supports_ansi_codes`, it will work for these, too, but how would we generalize the TermLike trait to allow for custom implementations and also be able to use the console crate's SyncGuard?

Ah ok. I see what you mean. Let me think about it a bit...

Funami580:
Any update, @chris-laplante?

chris-laplante:
> Any update, @chris-laplante?

Sorry, I finally got time to look at this. I think I was able to put all the pieces together in a way that makes sense for both crates. Please see https://github.com/console-rs/console/pull/205 for the `console` side of things, and my PR here: https://github.com/console-rs/indicatif/pull/625

Funami580:
> Sorry, I finally got time to look at this. I think I was able to put all the pieces together in a way that makes sense for both crates. Please see https://github.com/console-rs/console/pull/205 for the console side of things, and my PR here: https://github.com/console-rs/indicatif/pull/625

Thanks for taking a look at this! These PRs look good to me!

> Still TODO: implement support for WASM, Windows terminals.

Just a note for the future: The Windows 10 terminal can support ANSI codes, if you enable them. If you call `colors_supported()`, it tries to enable ANSI escape sequences and returns true on success. ANSI escape sequences are needed for the `supports_synchronized_output()` function (as well as the `SyncGuard`) to work correctly.

https://github.com/console-rs/console/blob/de2f15a31a8fef0b0e65ef4bdf92cd03c3061dac/src/windows_term/mod.rs#L77

So in the future it might look like this.

```rs
#[inline]
pub fn is_synchronized_output_supported(&self) -> bool {
    #[cfg(unix)]
    {
        supports_synchronized_output()
    }
    #[cfg(not(unix))]
    {
        self.colors_supported() && supports_synchronized_output()
    }
}
```

I think it would be more efficient to cache the result of `colors_supported()`, though, instead of re-doing the operation for every `SyncGuard`.

But right now, the `with_raw_terminal` that `supports_synchronized_output` uses doesn't work for Windows, so it won't work.

In my last update for this PR, I simply removed `supports_synchronized_output` since the ANSI escape sequences for synchronizing are invisible in the terminal, if the terminal does not support synchronization (when the terminal supports ANSI escape sequences).

By the way, this one line that I removed in console (`use std::ptr;`) was wrong, since macOS needs this and the tests for it now fail. It should be possible to put this line in `fn select_fd(...)`, however.

chris-laplante:
> > Sorry, I finally got time to look at this. I think I was able to put all the pieces together in a way that makes sense for both crates. Please see [console-rs/console#205](https://github.com/console-rs/console/pull/205) for the console side of things, and my PR here: #625
> 
> Thanks for taking a look at this! These PRs look good to me!
> 
> > Still TODO: implement support for WASM, Windows terminals.
> 
> Just a note for the future: The Windows 10 terminal can support ANSI codes, if you enable them. If you call `colors_supported()`, it tries to enable ANSI escape sequences and returns true on success. ANSI escape sequences are needed for the `supports_synchronized_output()` function (as well as the `SyncGuard`) to work correctly.
> 
> https://github.com/console-rs/console/blob/de2f15a31a8fef0b0e65ef4bdf92cd03c3061dac/src/windows_term/mod.rs#L77
> 
> So in the future it might look like this.
> 
> ```rust
> #[inline]
> pub fn is_synchronized_output_supported(&self) -> bool {
>     #[cfg(unix)]
>     {
>         supports_synchronized_output()
>     }
>     #[cfg(not(unix))]
>     {
>         self.colors_supported() && supports_synchronized_output()
>     }
> }
> ```
> 
> I think it would be more efficient to cache the result of `colors_supported()`, though, instead of re-doing the operation for every `SyncGuard`.
> 
> But right now, the `with_raw_terminal` that `supports_synchronized_output` uses doesn't work for Windows, so it won't work.
> 
> In my last update for this PR, I simply removed `supports_synchronized_output` since the ANSI escape sequences for synchronizing are invisible in the terminal, if the terminal does not support synchronization (when the terminal supports ANSI escape sequences).
> 
> By the way, this one line that I removed in console (`use std::ptr;`) was wrong, since macOS needs this and the tests for it now fail. It should be possible to put this line in `fn select_fd(...)`, however.

Would you mind copying this comment to the PR in console? That way I can reply to the console specific parts there.

PR #620, opened 2024-01-12
Revert heuristic algorythm

nagos:
Revert ETA heuristics, untill better solution found.

Current version produce large ETA results when cycle time is greater than 15s.

This reverts commit 011c998820a5e7f2a929c18b0aa882491b05d579.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -1,9 +1,9 @@
 use std::borrow::Cow;
-use std::io;
 use std::sync::Arc;
 use std::time::Duration;
 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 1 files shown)


djc:
I'm not interested in reverting this until we have better research showing that this is the best solution. Otherwise we're just stuck trading one person's use cases off against another's, which doesn't seem like an improvement.

afontenot:
I haven't really had time to look at issues for this project, but I want to add that my PR fixed some pretty significant issues with the old approach. If for some reason the approach was found wanting, the appropriate solution would involve fixing the bugs in the previous algorithm and not just blindly reverting.

PR #627, opened 2024-02-07
Fix truncated ANSI

remi-dupre:
Hi!

I noticed that truncated variables in templates were not ANSI aware. So here is a suggestion **based on https://github.com/console-rs/console/pull/206** to fix that :smile:


djc:
Thanks for having a look at this! Going to set it to draft state until there is some movement on the console PR.

Issue #638, opened 2024-03-31
Using ⚙️ emoji in ProgressBar spinner message causes unnessecary line breaks


Silvea12:
For some reason, the width of ⚙️ (a 2-char emoji) seems to have its length mis-calculated, and as a result puts every single refresh on a new line.
As seen in the Dioxus CLI when watching a build. Reproduced in both Konsole 24.02.1 and JetBrains RustRover 2023.3 EAP on Linux with both the Fira Code and Hack fonts.

Note: this only happens with ⚙️, not with ⚙.

Minimal reproduction code:
```toml
[dependencies]
indicatif = "0.17.8"
```

```rust
use std::time::Duration;
use indicatif::ProgressBar;

fn main() {
        let mut bar = ProgressBar::new_spinner();
        bar.enable_steady_tick(Duration::from_millis(200));
        bar.set_message("⚙️ Hi there!");
        std::thread::sleep(Duration::from_secs(5));
}
```

djc:
You probably just need to enable some of the Cargo features for proper Unicode support?

Silvea12:
Gave that a shot - same bug. Seems to happen regardless of if `improved_unicode` is enabled or not.

djc:
Well, maybe you can have a look at how `indicatif::style::measure()` fails with your example string, and if there's something that can be done to improve it?

djc:
> Additionally, in the ratatui issue that links to it, the crate https://github.com/jameslanska/unicode-display-width/ is mentioned as a possible replacement. I haven't tested yet, but it'd need to be moved into the `console` crate instead.

Why would need to be moved into the console crate? I'd be okay with taking on the extra dependency on this (only if `improved_unicode` is enabled).

Silvea12:
Could do, just feels weird to me to use the console crate for so few parts. But it'd just be a matter of replacing all occurrences of `console::measure_text_width()` with the `width()` method from the mentioned crate, maybe putting `console::ansi::strip_ansi_codes()` in before it depending on if you wanted that feature flag - I'm not sure what your intent is with dealing with ANSI there (though I'd assume you would want to strip it)

djc:
It's separation of concerns -- Unicode and dealing with the console are really separate domains so using different dependencies for these makes some amount of sense to me.

Silvea12:
Works for me then. I will try to find some time in the next couple days to whip up an MR.

Silvea12:
[...]

PR #639, opened 2024-04-04
Moved from unicode-width to unicode-display-width for visual grapheme width estimation

Silvea12:
This change had to be made in both crates in order to be updated correctly.
The console-rs crate has been bumped an extra version as a result of a behavior change of `char_width()` with the `unicode-display-width` feature enabled - it now returns always 1 or 2, never 0 as it used to.
The indicatif crate may also need to be bumped a full version as a result of the dependency change, but I'm not sure if you'd consider this a breaking change to the same degree, as it's not an advertised feature.

Related issue: https://github.com/console-rs/indicatif/issues/638
Related PR: https://github.com/console-rs/console/pull/210

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "indicatif"
-version = "0.17.8"
+version = "0.17.9"
 edition = "2021"
 rust-version = "1.63"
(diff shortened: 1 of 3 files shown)


Silvea12:
Pulling in @djc as requested

PR #647, opened 2024-05-25
Improve trimming and alignment with multi-byte characters

### LINKED ISSUE #497: ProgressBar scrolls instead of redraws when `◎` character is present
I am experiencing my ProgressBar scrolling instead of truncating/redrawing when I shrink my terminal's window.

I am running on a remote Ubuntu server. Terminal is Mac's iTerm2, with and without tmux. The ProgressBar's template uses `{wide_msg}`, so I expect the progress bar to truncate text as the window sizes decreases.

This only manifests when I shrink my *past* the `◎` character. If this character is not present in the message, the progress bar works as expected regardless of how small I shrink my terminal's window.

Also, version v0.16.2 does *not* exhibit this behavior; it works as expected.

I've tried v0.17.1 and v0.17.2. With v0.17.2 I also tried using the "improved_unicode" feature. None of these resulted in correct behavior.

Here's more information in an issue I have against my repo where I debugged this:
https://github.com/solana-labs/solana/issues/29158

If it's helpful, I can get a screen capture and/or a set of steps to reproduce the issue (may be slightly involved though). Thanks in advance!

djc:
Intended to fix #497, but this doesn't quite make sense to me yet.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -738,6 +738,57 @@ enum Alignment {
     Right,
 }
 
+impl Alignment {
+    fn trim(self, s: &str, width: usize) -> &str {
+        dbg!(s);
+        let cur = measure_text_width(s);
+        if dbg!(cur) <= width {
+            return s;
+        }
+
+        let mut iter = s.char_indices();
+        let (mut i, mut start, mut end, mut reduced) = (0, 0, s.len(), 0);
+        let mut buf = [0; 4];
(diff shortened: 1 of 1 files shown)


chris-laplante:
See https://github.com/console-rs/indicatif/issues/497#issuecomment-2537582942. Might want to give https://github.com/Aetf/unicode-truncate/tree/master a try.

Issue #143, opened 2019-12-22
MultiProgress bar flickers#2


inv2004:
I did another try of indicatif 0.13.0, and found that as soon as I only one ProgressBar into MultiProgress it starts to flicker the same I described in https://github.com/mitsuhiko/indicatif/issues/42.

Regards,

mibac138:
Please submit a code reproducing the problem and your environment

inv2004:
windows-10. conemu.

It is just a bit modified example/multi.rs
If I change ranges for 10k 100k 10k - it stops flicker.
https://gist.github.com/127aff606d2d5bf5e8ab9197c7a9b71b

Deniskore:
I got the same flickering problem using MultiProgress bar with thread pool.
The code is almost the same as in the example.

Deniskore:
```rust
    let m = MultiProgress::new();
    m.set_move_cursor(true);
```
Solves the problem with flickering

inv2004:
@Deniskore 
Just checked - it helped. Thank you.

PR #648, opened 2024-06-10
Fix move cursor flag not working

SuperTux88:
I had the problem of my progress bars starting to flicker after adding some more bars, especially over an ssh connection. I then found #42 and from that then found #143, but the fix described there to set `m.set_move_cursor(true);` didn't make a difference at all.

After some debugging, I found out, that the move_cursor flag was used in the `draw_to_term()` function, but this is only called from a `Drawable::Term`, while the `set_move_cursor` was only set on the `MultiState` which is only used in the `Drawable::Multi`. So in the current state, it never has any effect when actually drawing to the terminal.

Then after forwarding the flag to the correct DrawState instance, I noticed, that the first bar always starts drawing one line too high on the last char, so the first line was just off-by-one. This was because the cursor is always on the last char of the last line, and then moving up the number of bars, will always end up one line too high. So just move up one line less, but then the cursor also still needs to be moved to the front of the line.

So this now fixes using of `set_move_cursor(true)` again, so it can be used to fix the flickering.


src/multi.rs
```diff
@@ -227,13 +225,16 @@ impl MultiState {
...
             ordering: vec![],
             draw_target,
-            move_cursor: false,
             alignment: MultiProgressAlignment::default(),
             orphan_lines: Vec::new(),
             zombie_lines_count: VisualLines::default(),
         }
     }
 
+    fn set_move_cursor(&mut self, move_cursor: bool) {
```
djc, line 234: Don't see much value in adding this method? Looks like `MultiProgress::set_move_cursor()` can just reach into `MultiState::draw_target` directly.

djc (commented):
Thanks for the investigation!

Would you be able to add a test that can help us avoid regressing this in the future?

src/multi.rs
SuperTux88: You are right, fixed :+1:

SuperTux88:
Thanks for the review.

> Would you be able to add a test that can help us avoid regressing this in the future?

I tried to do that, but failed, because the output looks the same, whether it moves the cursor or uses clear, so the test was always green, even when it was broken (so it wasn't useful). But I just noticed `moves_since_last_check()` where I can check for moves specifically, and added a test using that. So it now fails when it gets broken. :+1:

PR #650, opened 2024-07-11
Fix AtomicPosition::reset storing wrong value

TheJokr:
`AtomicPosition::allow` interprets `self.prev` as a number of nanoseconds since `self.start`, but `AtomicPosition::reset` instead stored a number of milliseconds in there. This is a bug, albeit without significant consequences.

I reckon the bug got introduced because the comments are worded a bit ambiguously, so I cleared those up and added a regression test for the future.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -537,7 +537,7 @@ impl AtomicPosition {
         }
 
         let mut capacity = self.capacity.load(Ordering::Acquire);
-        // `prev` is the number of ms after `self.started` we last returned `true`, in ns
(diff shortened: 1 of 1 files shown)


djc (approved):
Awesome, thanks! How did you find this?

TheJokr:
I was checking out how often `tick()` is scheduled by `ProgressBar::inc()`, so I read through that part of the code and was confused by the comments 😅 And then I noticed the inconsistency with the `reset()` function, which was right below.

PR #657, opened 2024-09-09
initial draft oft hold_max with heuristic

djugei:
alternative or update to #642 

In BufRead/AsyncBufRead fill_buf does not logically consume the reader/advance progress, only a call to consume does. this was the wrong way around for the AsyncBufRead implementation, i have fixed that in this pr.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -62,6 +62,7 @@ where
 pub struct ProgressBarIter<T> {
     pub(crate) it: T,
     pub progress: ProgressBar,
+    pub(crate) dejitter: MaxSeekHeuristic,
 }
 
 impl<T> ProgressBarIter<T> {
@@ -155,25 +156,37 @@ impl<T: FusedIterator> FusedIterator for ProgressBarIter<T> {}
 impl<R: io::Read> io::Read for ProgressBarIter<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         let inc = self.it.read(buf)?;
-        self.progress.inc(inc as u64);
+        self.progress.set_position(
+            self.dejitter
+                .update_seq(self.progress.position(), inc as u64),
+        );
         Ok(inc)
     }
 
     fn read_vectored(&mut self, bufs: &mut [IoSliceMut<'_>]) -> io::Result<usize> {
         let inc = self.it.read_vectored(bufs)?;
-        self.progress.inc(inc as u64);
+        self.progress.set_position(
+            self.dejitter
+                .update_seq(self.progress.position(), inc as u64),
+        );
         Ok(inc)
     }
 
     fn read_to_string(&mut self, buf: &mut String) -> io::Result<usize> {
         let inc = self.it.read_to_string(buf)?;
-        self.progress.inc(inc as u64);
+        self.progress.set_position(
+            self.dejitter
+                .update_seq(self.progress.position(), inc as u64),
+        );
         Ok(inc)
     }
 
     fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
         self.it.read_exact(buf)?;
-        self.progress.inc(buf.len() as u64);
+        self.progress.set_position(
+            self.dejitter
+                .update_seq(self.progress.position(), buf.len() as u64),
+        );
         Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


djugei:
this is now ready for some feedback @djc

djugei:
would love some feedback on this

djugei:
rebased to 1.8

src/iter.rs
```diff
@@ -150,30 +230,78 @@ impl<T: DoubleEndedIterator> DoubleEndedIterator for ProgressBarIter<T> {
     }
 }
 
+impl<T: ExactSizeIterator> ExactSizeIterator for ProgressBarIter<T> {
```
djc, line 233: Please don't move code around.

src/iter.rs
```diff
@@ -57,11 +57,97 @@ where
     }
 }
 
+#[derive(Debug)]
```
djc, line 60: This new code should go somewhere nearer the bottom of the module, since the project is organized in a top-down fashion.

src/iter.rs
```diff
@@ -57,11 +57,97 @@ where
...
+struct RingBuf<const SIZE: usize = 10> {
```
djc, line 61: It doesn't seem like making `SIZE` a generic const is helpful since we're not actually using that in this project?

src/iter.rs
```diff
@@ -57,11 +57,97 @@ where
...
+#[derive(Debug)]
+pub(crate) struct RingBufWrap<const RESET: u8 = 5, const BUFSIZE: usize = 10> {
+    buf: Option<(Box<RingBuf<BUFSIZE>>, u8)>,
+}
+
+impl<const RESET: u8, const BUFSIZE: usize> RingBufWrap<RESET, BUFSIZE> {
+    pub(crate) fn new() -> Self {
+        Self { buf: None }
+    }
+    fn update_seq(&mut self, oldpos: u64, delta: u64) -> u64 {
```
djc, line 118: Nits: empty line for the module. As above wrt the `const` generics. Use let-else with early return to reduce rightward drift in this method.

src/iter.rs
```diff
@@ -57,11 +57,97 @@ where
...
+                buf.max()
+            }
+        } else {
+            newpos
+        }
+    }
+
+    fn update_seek(&mut self, newpos: u64) -> u64 {
+        let (b, seq) = self.buf.get_or_insert_with(|| {
+            let b = Box::new(RingBuf::<BUFSIZE>::new());
```
djc, line 136: Nit: inline the `b` binding.

src/iter.rs
```diff
@@ -57,11 +57,97 @@ where
...
+        self.head = (self.head + 1) % (self.vals.len() as u8);
+    }
+
+    fn max(&self) -> u64 {
+        self.max
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct RingBufWrap<const RESET: u8 = 5, const BUFSIZE: usize = 10> {
```
djc, line 110: `RingBufWrap` should go above `RingBuf`. I don't think these names are very helpful in understanding the types' roles in this project.

src/iter.rs
```diff
@@ -150,30 +230,78 @@ impl<T: DoubleEndedIterator> DoubleEndedIterator for ProgressBarIter<T> {
...
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let item = std::pin::Pin::new(&mut this.it).poll_next(cx);
+        match &item {
+            std::task::Poll::Ready(Some(_)) => this.progress.inc(1),
+            std::task::Poll::Ready(None) => this.progress.finish_using_style(),
+            std::task::Poll::Pending => {}
```
djc, line 265: Import all of the types (`Pin`, `Context`, `Poll`) to avoid the unnecessary clutter from import paths.

djc (commented):
Here's some initial feedback. Needs a bunch of work.

Please squash all of your commits and avoid making changes to the existing structure of the code as much as possible (or isolate such changes in separate commits).

src/iter.rs
djugei: yeah that was to make playing with different sizes easier for me, i can take it out in the final commit.

djugei:
> Please squash all of your commits
will do before a merge, i am keeping it in separate commits for now to make reviews easier (for example its somewhat easier to ignore the code movements by just not looking at that commit etc)

i will make the requested changes over the weekend probably

src/iter.rs
djugei: renamed and moved

src/iter.rs
djugei: un-moved pre-existing code so this does not show up in the diff anymore

djugei:
I have done all(-1) of the requested changes, also split out the bug fix into a separate commit while squashing all other changes.

src/iter.rs
```diff
@@ -203,6 +223,116 @@ impl<S: io::Seek> io::Seek for ProgressBarIter<S> {
...
+    /// therefore amortizing to O(1).
+    ///
+    /// In case there is some linear increase with jitter,
+    ///   as expected in this specific use-case,
+    /// as long as there is one bigger update each HISTORY updates the scan is never triggered at all.
+    ///
+    /// Worst case would be linearly decreasing values, which is still O(1).
+    fn update(&mut self, new: u64) {
+        // % here is a purely for the optimizer
+        let head: usize = usize::from(self.head) % self.vals.len();
```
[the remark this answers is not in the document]
djugei: i have decided doubling down on this, i am altering the docs/comments to talk more about invariants.

i have just rechecked, it does fully eliminate bounds checks. while bounds checks are not _that_ performance intensive, it also eliminates the entire panic path, which has a positive impact on code size. Also it just feels right to feed the compiler additional information.

i was somewhat tempted to use std::hint::assert_unchecked, which has much the same effect, but also eliminates the modulo calculations. i opted to not introduce unsafe code, though its correctness can be somewhat trivially asserted.

src/iter.rs
```diff
@@ -203,6 +223,116 @@ impl<S: io::Seek> io::Seek for ProgressBarIter<S> {
...
+            if *seq >= RESET {
+                self.buf = None;
+                return newpos;
+            }
+
+            buf.update(newpos);
+            buf.max()
+        } else {
+            newpos
+        }
```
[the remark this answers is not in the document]
djugei: not my style but i will addapt.

djugei:
the clippy issue is literally a change you requested, but i have fixed it now.

djc (approved):

Issue #447, opened 2022-07-11
`MultiProgress:println` doesn't behave correctly when no progress bars have been added yet


chris-laplante:
Try this render test:

```rust
#[test]
fn multi_progress_println_newline() {
    let in_mem = InMemoryTerm::new(10, 10);
    let mp =
        MultiProgress::with_draw_target(ProgressDrawTarget::term_like(Box::new(in_mem.clone())));

    mp.println("This line is tooooooo long for the terminal\r\n").unwrap();
    mp.println("but it should wrap correctly\r\n").unwrap();

    assert_eq!(in_mem.contents(), "This line\nis toooooo\no long for\n the termi\nnal\nbut it sho\nuld wrap c\norrectly");
}
```

It fails like this:

```
---- multi_progress_println_newline stdout ----
thread 'multi_progress_println_newline' panicked at 'assertion failed: `(left == right)`
  left: `"This line\nis toooooo\no long for\n the termi\nnalbut it\nshould wra\np correctl\ny"`,
 right: `"This line\nis toooooo\no long for\n the termi\nnal\nbut it sho\nuld wrap c\norrectly"`', tests/render.rs:607:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The issue is that the underlying `DrawState::draw_to_term` method assumes that the last line to be drawn doesn't contain a newline. This is true when progress bars are involved, but not true when just calling `MultiProgress:println` with no progress bars.

I think we need to add some extra logic to `draw_to_term` so that it behaves differently when no progress bars are present. 

The workaround I am using right now is to just use `MultiProgress::suspend` and then use `eprintln!` inside that.

chris-laplante:
FYI this is related to #442 I think

djc:
> I think we need to add some extra logic to `draw_to_term` so that it behaves differently when no progress bars are present.

Seems reasonable.

Ekleog:
FWIW, this is what it manifests as:

![Screenshot from 2023-02-07 00-18-08](https://user-images.githubusercontent.com/411447/217109220-8335ef71-9961-4341-8b3c-f5940073982c.png)

```rust
fn main() {
    let progress = indicatif::MultiProgress::new();
[...]

PR #659, opened 2024-09-26
Fix bar-less text output

spoutn1k:
This is my attempt at fixing #447.

The code in `draw_to_term` is a bit of mess, with double definitions, complex branching, and I tried cleaning it up. I introduce a `LineType` enum to differentiate between text, bars and empty lines. The variables names have been renamed to reflect better what they represent. Some code has been shuffled around. Tests pass and I ran examples at random with no issues.

A sample program to understand what this PR wants to fix:
```rs
static LOREM_SMALL: &str = "Lorem ipsum dolor sit amet.";
static LOREM_BIG: &str= "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec nec viverra massa. Nunc nisl lectus,
auctor in lorem eu, maximus elementum est.";

fn generate_logs(manager: &indicatif::MultiProgress, i: u32) {
    if i % 20 == 0 {
        let _ = manager.println(LOREM_BIG);
    } else if i % 10 == 0 {
        let _ = manager.println(LOREM_SMALL);
    }
}

fn main() {
    let display_manager = indicatif::MultiProgress::new();
    let bar = display_manager.add(indicatif::ProgressBar::new(100));

    for i in 0..100 {
        generate_logs(&display_manager, i);
        std::thread::sleep(std::time::Duration::from_millis(10));
    }

    for i in 0..100 {
        bar.inc(1);
        generate_logs(&display_manager, i);
        std::thread::sleep(std::time::Duration::from_millis(100));
    }
}
```

This is the output of the above before the PR:
[asciicast](https://asciinema.org/a/mwXUYBO3HC0wtYV4jLjjBVvNp)

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -511,7 +511,6 @@ impl DrawState {
             _ => VisualLines::default(),
         };
 
-        let term_height = term.height() as usize;
         let term_width = term.width() as usize;
         debug_assert!(self.orphan_lines_count <= self.lines.len());
         let orphan_visual_line_count =
@@ -540,7 +539,7 @@ impl DrawState {
                 // If so, then `real_len` should be at least `orphan_visual_line_count`.
                 debug_assert!(orphan_visual_line_count <= real_len);
                 // Don't consider orphan lines when comparing to terminal height.
-                if real_len - orphan_visual_line_count + diff > term_height.into() {
+                if real_len - orphan_visual_line_count + diff > term.height().into() {
                     break;
                 }
             }
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -362,7 +362,7 @@ pub(crate) enum LineAdjust {
 
 pub(crate) struct DrawStateWrapper<'a> {
     state: &'a mut DrawState,
-    orphan_lines: Option<&'a mut Vec<String>>,
+    orphan_lines: Option<&'a mut Vec<LineType>>,
 }
 
 impl<'a> DrawStateWrapper<'a> {
@@ -373,7 +373,7 @@ impl<'a> DrawStateWrapper<'a> {
         }
     }
 
-    pub(crate) fn for_multi(state: &'a mut DrawState, orphan_lines: &'a mut Vec<String>) -> Self {
+    pub(crate) fn for_multi(state: &'a mut DrawState, orphan_lines: &'a mut Vec<LineType>) -> Self {
         Self {
             state,
             orphan_lines: Some(orphan_lines),
@@ -460,7 +460,7 @@ const MAX_BURST: u8 = 20;
 #[derive(Clone, Debug, Default)]
 pub(crate) struct DrawState {
     /// The lines to print (can contain ANSI codes)
-    pub(crate) lines: Vec<String>,
+    pub(crate) lines: Vec<LineType>,
     /// The number [`Self::lines`] entries that shouldn't be reaped by the next tick.
     ///
     /// Note that this number may be different than the number of visual lines required to draw [`Self::lines`].
@@ -525,12 +525,13 @@ impl DrawState {
         };
 
         // Accumulate the displayed height in here. This differs from `full_height` in that it will
-        // not reflect the displayed content if the terminal height is exceeded.
+        // accurately reflect the number of lines that have been displayed on the terminal, if the
+        // full height exceeds the terminal height.
         let mut real_height = VisualLines::default();
 
         for (idx, line) in self.lines.iter().enumerate() {
-            let line_width = console::measure_text_width(line);
-            let diff = if line.is_empty() {
+            let line_width = console::measure_text_width(line.as_ref());
+            let diff = if line.as_ref().is_empty() {
                 // Empty line are new line
                 1
             } else {
@@ -561,7 +562,7 @@ impl DrawState {
                 term.write_line("")?;
             }
 
-            term.write_str(line)?;
+            term.write_str(line.as_ref())?;
 
             if idx + 1 == self.lines.len() {
                 // Keep the cursor on the right terminal side
@@ -584,7 +585,7 @@ impl DrawState {
 
     pub(crate) fn visual_line_count(
         &self,
-        range: impl SliceIndex<[String], Output = [String]>,
+        range: impl SliceIndex<[LineType], Output = [LineType]>,
         width: usize,
     ) -> VisualLines {
         visual_line_count(&self.lines[range], width)
@@ -649,6 +650,30 @@ pub(crate) fn visual_line_count(lines: &[impl AsRef<str>], width: usize) -> Visu
     }
 
     real_lines.into()
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum LineType {
+    Text(String),
+    Bar(String),
+    Empty,
+}
+
+impl LineType {}
+
+impl AsRef<str> for LineType {
+    fn as_ref(&self) -> &str {
+        match self {
+            LineType::Text(s) | LineType::Bar(s) => s,
+            LineType::Empty => "",
+        }
+    }
+}
+
+impl PartialEq<str> for LineType {
+    fn eq(&self, other: &str) -> bool {
+        self.as_ref() == other
+    }
 }
 
 #[cfg(test)]
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -543,18 +543,14 @@ impl DrawState {
             let line_height = line.wrapped_height(term_width);
 
             // Check here for bar lines that exceed the terminal height
-            if text_line_count <= idx {
-                // If all the orphan lines have been drawn, then `real_height` should be
-                // at least `orphan_visual_line_count`.
-                debug_assert!(text_height <= real_height);
-
-                // Don't consider orphan lines when comparing to terminal height.
-                if real_height - text_height + line_height > term.height().into() {
+            if matches!(line, LineType::Bar(_)) {
+                // Stop here if printing this bar would exceed the terminal height
+                if real_height + line_height > term.height().into() {
                     break;
                 }
-            }
 
-            real_height += line_height;
+                real_height += line_height;
+            }
 
             // Print a new line if this is not the first line printed this tick
             // the first line will automatically wrap due to the filler below
@@ -573,7 +569,7 @@ impl DrawState {
         }
 
         term.flush()?;
-        *bar_count = real_height - text_height + shift;
+        *bar_count = real_height + shift;
 
         Ok(())
     }
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -505,21 +505,8 @@ impl DrawState {
 
         let term_width = term.width() as usize;
 
-        // The number of text lines that are contained in this draw state
-        let text_line_count = self
-            .lines
-            .iter()
-            .filter(|line| matches!(line, LineType::Text(_) | LineType::Empty))
-            .count();
-
-        // Here we calculate the terminal vertical height each of those groups require when
-        // printing, taking wrapping into account.
-        let text_height = self.visual_line_count(..text_line_count, term_width);
-        let bar_height = self.visual_line_count(text_line_count.., term_width);
+        // Here we calculate the terminal vertical real estate that the state requires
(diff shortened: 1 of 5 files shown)


src/draw_target.rs
```diff
@@ -398,8 +398,13 @@ impl std::ops::DerefMut for DrawStateWrapper<'_> {
 impl Drop for DrawStateWrapper<'_> {
     fn drop(&mut self) {
         if let Some(orphaned) = &mut self.orphan_lines {
-            orphaned.extend(self.state.lines.drain(..self.state.orphan_lines_count));
-            self.state.orphan_lines_count = 0;
+            orphaned.extend(
```
djc, line 401: This commit has a lot going on. Would you mind trying to split it into smaller commits that make a more minimal single logical change?

src/draw_target.rs
```diff
@@ -456,15 +461,60 @@ impl RateLimiter {
 
 const MAX_BURST: u8 = 20;
 
+#[derive(Debug, Clone, PartialEq)]
```
djc, line 464: We have a top-down order here, this should move down below any types that rely on it (`DrawState` at least).

djc (commented):
Thanks for trying to improve this code. Especially because this is pretty gnarly code, it would be good to split these changes into smaller commits that make minimal (logical) changes -- for example, might introduce the `LineType` enum while representing only the types of lines that are currently in lines, and then later introduce the other variants.

spoutn1k:
Thank you for taking the time to review this. My tests omitted the `--all-features` flag and it shows. It did not help that this PR is the result of a caffeine overdose and subsequent insomnia. Comments duly noted this PR will either be force pushed or a new one will be opened with working tests and better commits. Thanks again !

djc:
Thanks! Force pushing is fine IME.

spoutn1k:
https://github.com/console-rs/indicatif/blob/5396704f8e9fe781d21a5475c8d3af292d0155fe/src/draw_target.rs#L554-L556

I think my main issue stems here. For wrapping lines, this calculation is not accurate. However the concept itself seems strange to me. Why fill the line with whitespace and expect the terminal to wrap instead of moving the cursor down when needed ? This means when resizing the terminal the formatting will fall apart.

Any objections to me investigating changing this behaviour ?

spoutn1k:
The new commit passes all the tests but two, and that is because the filler is not only printed when needed (removing one variable definition). On my machine:

```
---- multi_progress_many_bars stdout ----
thread 'multi_progress_many_bars' panicked at tests/render.rs:1230:5:
assertion failed: `(left == right)`
                                                                                                                                                                                 
Diff < left / right > :
 Clear
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("                                                                             ")
 Flush
 Up(1)
 Clear
 Down(1)
 Clear
 Up(1)
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("")
 NewLine
 Str("⠁ 1")
 Str("                                                                             ")
 Flush
 Up(2)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Up(2)
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("")
 NewLine
 Str("⠁ 1")
 Str("")
 NewLine
 Str("⠁ 2")
 Str("                                                                             ")
 Flush
 Up(3)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Up(3)
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("")
 NewLine
 Str("⠁ 1")
 Str("")
 NewLine
 Str("⠁ 2")
>Str("")
 Flush
 Up(3)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Up(3)
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("")
 NewLine
 Str("⠁ 1")
 Str("")
 NewLine
 Str("⠁ 2")
>Str("")
 Flush
 Up(3)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Up(3)
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("")
 NewLine
 Str("⠁ 1")
 Str("")
 NewLine
 Str("⠁ 2")
>Str("")
 Flush
 Up(3)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Down(1)
 Clear
 Up(3)
 Str("░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0/10")
 Str("")
 NewLine
 Str("⠁ 0")
 Str("")
 NewLine
 Str("⠁ 1")
 Str("")
 NewLine
 Str("⠁ 2")
>Str("") <---------------------------------- Those useless prints are avoided
 Flush
```

djc:
> I think my main issue stems here. For wrapping lines, this calculation is not accurate. However the concept itself seems strange to me. Why fill the line with whitespace and expect the terminal to wrap instead of moving the cursor down when needed ? This means when resizing the terminal the formatting will fall apart.
> 
> Any objections to me investigating changing this behaviour ?

I never have objections to other people investigating things. 😄 Suggest reading git blame to figure out why this change was made, pretty sure it wasn't just for funsies.

djc:
> The new commit passes all the tests but two, and that is because the filler is not only printed when needed (removing one variable definition).

Removing the empty writes from the test seems fine to me.

spoutn1k:
> Suggest reading git blame to figure out why this change was made, pretty sure it wasn't just for funsies.

I believe this is working in tandem with this check:
https://github.com/console-rs/indicatif/blob/5396704f8e9fe781d21a5475c8d3af292d0155fe/src/draw_target.rs#L549-L551

This means that for everyline line other than the first one, a newline is added. The first is expected to wrap automatically due to the filler set in my previous comment.

spoutn1k:
This effectively fixes #447. I deeply want to keep going, because there is a lot of further refactoring that can be done, and the current solution of padding the last line will break any resizing. This is for future reference, the watermark to go back to if the PR spins out of control

spoutn1k:
In a rare moment of lucidity I will not change the line padding in this PR. It is now ready for review again @djc !

src/draw_target.rs
```diff
@@ -475,19 +475,37 @@ impl DrawState {
...
         &mut self,
         term: &(impl TermLike + ?Sized),
-        last_line_count: &mut VisualLines,
+        bar_count: &mut VisualLines, // The number of dynamic lines printed at the previous tick
     ) -> io::Result<()> {
         if panicking() {
             return Ok(());
         }
 
+        let term_height = term.height() as usize;
```
djc, line 533: This first commit has too many logically independent changes going on. In general, I don't think I'm a fan of hoisting up all these variable bindings, further up from where they're actually used.

spoutn1k: All the changes are variable renaming or movement. The hoisting up is to have a clear definition of what each variable is and relates to the others, especially as some of those variables were defined twice. It is also temporary and here to make the coming changes clearer.

djc: Especially for variable renaming and movement, I want those in separate commits so reviewing it is easier. I definitely don't want a whole bunch of variable declarations up front, each variable should be defined shortly before its first usage site.

src/draw_target.rs
```diff
@@ -498,8 +498,21 @@ impl DrawState {
...
+
+        // Here we calculate the terminal vertical height each of those groups require when
+        // printing, taking wrapping into account.
+        let orphan_visual_line_count = self.visual_line_count(..text_line_count, term_width);
+        let bar_height = self.visual_line_count(text_line_count.., term_width);
+        let visual_lines = self.visual_line_count(.., term_width);
+
+        // Sanity checks
+        debug_assert!(visual_lines == orphan_visual_line_count + bar_height);
+        debug_assert!(self.orphan_lines_count <= self.lines.len());
```
djc: We need `visual_lines` for the `shift` definition, but I'd prefer to defer the calculation of `orphan_visual_line_count` until after the `shift` definition.

src/draw_target.rs
```diff
@@ -471,6 +471,34 @@ pub(crate) struct DrawState {
     pub(crate) alignment: MultiProgressAlignment,
 }
 
+#[derive(Clone, Debug)]
```
djc, line 475: Please move this down to just before the `tests` module.

src/draw_target.rs
```diff
@@ -471,6 +471,34 @@ pub(crate) struct DrawState {
...
+impl LineType {
+    pub(crate) fn inner(&self) -> &str {
+        match self {
+            LineType::Text(s) | LineType::Bar(s) => s,
+            LineType::Empty => "",
+        }
+    }
+}
+
+impl AsRef<str> for LineType {
```
djc, line 507: Let's inline `inner()` into this impl and use `as_ref()` everywhere.

src/draw_target.rs
```diff
@@ -558,44 +587,34 @@ impl DrawState {
...
-                // This take in account linewrap from terminal
-                let terminal_len = (line_width as f64 / term_width as f64).ceil() as usize;
-
-                // If the line is effectively empty (for example when it consists
-                // solely of ANSI color code sequences, count it the same as a
-                // new line. If the line is measured to be len = 0, we will
-                // subtract with overflow later.
-                usize::max(terminal_len, 1)
-            }
-            .into();
```
djc: Would prefer that changes to this code happen in a separate commit from moving it into a method.

src/draw_target.rs
```diff
@@ -667,21 +686,15 @@ impl Sub for VisualLines {
...
-        real_lines += usize::max(
-            (effective_line_length as f64 / width as f64).ceil() as usize,
-            1,
-        );
-    }
-
-    real_lines.into()
+pub(crate) fn visual_line_count(lines: &[LineType], width: usize) -> VisualLines {
+    lines.iter().fold(VisualLines::default(), |acc, line| {
+        acc.saturating_add(line.wrapped_height(width))
```
djc, line 670: Would also like the changes here to be in a separate commit.

src/draw_target.rs
```diff
@@ -485,6 +485,22 @@ impl LineType {
             LineType::Empty => "",
         }
     }
+
+    fn console_width(&self) -> usize {
```
djc, line 490: Nit: `console_width()` should appear after `wrapped_height()`, since the latter depends on the former.

tests/render.rs
```diff
@@ -1903,3 +1903,29 @@ fn orphan_lines_message_above_progress_bar_test(pb: &ProgressBar, in_mem: &InMem
 
     pb.finish();
 }
+
+#[test]
```
djc, line 1907: Nit: please move the docstring before the attribute.

djc:
Alright, thanks for all your work on this!

spoutn1k:
Thank you, it was a pleasure, I just am travelling a lot at the moment. Cheers !

PR #664, opened 2024-11-07
feat: allow constructing and setting the progress bar len to `None`

cdellacqua:
For situations where the length of a progress bar can only be discovered after some processing, it might be more appropriate to show an empty bar rather than a full one of length 0 (`0/0` is represented with a ratio of `1.0`).

This PR introduces a new constructor, `ProgressBar::empty()`, and a new method, `ProgressBar::unset_length()`. These two allow library users to have a length of `None`, which is represented as a ratio of `0.0`, achieving the empty state.


src/progress_bar.rs
```diff
@@ -47,6 +47,18 @@ impl ProgressBar {
...
 
+    /// Creates a new progress bar without a specified length
+    ///
+    /// This progress bar by default draws directly to stderr, and refreshes a maximum of 20 times
+    /// a second. To change the refresh rate, [set] the [draw target] to one with a different refresh
+    /// rate.
+    ///
+    /// [set]: ProgressBar::set_draw_target
+    /// [draw target]: ProgressDrawTarget
+    pub fn empty() -> Self {
```
djc, line 58: I think the `empty()` name isn't great for this. Maybe `no_length()`?

djc (commented):
I think this makes sense, thanks.

src/progress_bar.rs
cdellacqua: Sure, no problem!

djc (approved):

Issue #665, opened 2024-11-10
The dependency `instant` is no longer maintained - consider switching to `web-time` instead


torokati44:
We just got this notice:

```
error[unmaintained]: `instant` is unmaintained
    ┌─ /github/workspace/Cargo.lock:260:1
    │
260 │ instant 0.1.13 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0384
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0384
    ├ This crate is no longer maintained, and the author recommends using the maintained [`web-time`] crate instead.
      
      [`web-time`]: https://crates.io/crates/web-time
    ├ Solution: No safe upgrade is available!
    ├ instant v0.1.13
      └── indicatif v0.17.8
          ├── exporter v0.1.0
          └── ruffle_scanner v0.1.0
```

See also: https://crates.io/crates/instant

PR #666, opened 2024-11-10
Replace `instant` with `web-time`

torokati44:
Fixes https://github.com/console-rs/indicatif/issues/665. See: https://crates.io/crates/web-time.

Note that unlike `instant`, `web-time` falls back to `std::time` on non-WASM, so this kind of conditional exports are not really necessary:

```rust
#[cfg(not(target_arch = "wasm32"))]
use std::time::Instant;
#[cfg(target_arch = "wasm32")]
use web_time::Instant;
```

Instead, it could unconditionally be `web_time::Instant`, if making `web-time` a (really light, to be fair) dependency on non-WASM platforms as well is acceptable. I can update the PR accordingly, if desired.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ futures = "0.3" # so the doctest for wrap_stream is nice
 pretty_assertions = "1.4.0"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-instant = "0.1"
+web-time = "1.1.0"
 
 [features]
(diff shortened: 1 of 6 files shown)


djc (approved):
LGTM, thanks!

> Instead, it could unconditionally be web_time::Instant, if making web-time a (really light, to be fair) dependency on non-WASM platforms as well is acceptable. I can update the PR accordingly, if desired.

I prefer keeping it like this.

djc:
@torokati44 see #667.

Issue #670, opened 2024-11-26
Regression: text lines print twice


spoutn1k:
We observed a regression after my series of patches. In some configurations the text output would be doubled and result in output of the form:

```
[root@node ~]# ./executable
Message 1/4
Message 1/4
Message 2/4
Message 2/4
⠉ [3s] [=>                  ] 2/100 Progress message
```

I'm fully committed to fixing my mess and have identified the change responsible.

djc:
Thanks for looking at this! Someone else also reported #669 which seems like it might have been caused by your changes.

spoutn1k:
They mention 0.17.9, which does not include my patch, right ?

djc:
> They mention 0.17.9, which does not include my patch, right ?

Hmm, you're right.

infiniteregrets:
hi, I think i am facing this same issue on 0.17.9

i have `Multiprogress` instance containing two bars, when I `abandon` or `finish` either of the bars, it closes with two bars

<img width="417" alt="Screenshot 2024-12-13 at 11 08 22 PM" src="https://github.com/user-attachments/assets/5155e260-9f13-4925-9213-baa79024cab1" />

spoutn1k:
Try the new master, the push just got merged in

infiniteregrets:
> Try the new master, the fix just got merged in

hi, I just tried the main branch and it still persists for me, when I do ctrl + C it gets printed twice ):

spoutn1k:
Then hopefully it is #669 ...

DBY047:
@spoutn1k retested #669 with master and the issue is still present.

spoutn1k:
Hi @DBY047, I just checked with the example in https://github.com/console-rs/indicatif/pull/671 on both d8a066653f37110b9b45a41102cd12f7879e9d1e and master and I can confirm the bug I opened this issue with is fixed:

```
indicatif> cargo run --example issue
   Compiling getrandom v0.3.1
   Compiling zerocopy v0.8.21
   Compiling indicatif v0.17.11 (/Users/jb/Projects/indicatif)
   Compiling rand_core v0.9.3
   Compiling rand_chacha v0.9.0
   Compiling rand v0.9.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.26s
     Running `target/debug/examples/issue`
This should print once
indicatif> git checkout d8a066653f37110b9b45a41102cd12f7879e9d1e                                                            
[...]                                                                                                                              
[...]

PR #671, opened 2024-11-26
Fix double prints

spoutn1k:
Fix for #670.

Will work on a reproducing the issue and creating a test for it, but in the meantime this fixes the issue experimentally.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -398,13 +398,13 @@ impl std::ops::DerefMut for DrawStateWrapper<'_> {
 impl Drop for DrawStateWrapper<'_> {
     fn drop(&mut self) {
         if let Some(orphaned) = &mut self.orphan_lines {
-            orphaned.extend(
-                self.state
-                    .lines
-                    .iter()
-                    .filter(|l| matches!(l, LineType::Text(_) | LineType::Empty))
-                    .cloned(),
-            );
+            let (text, bars): (Vec<_>, Vec<_>) = self
+                .state
+                .lines
+                .drain(..)
+                .partition(|l| matches!(l, LineType::Text(_) | LineType::Empty));
+            self.state.lines = bars;
+            orphaned.extend(text);
         }
     }
 }
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -397,14 +397,19 @@ impl std::ops::DerefMut for DrawStateWrapper<'_> {
 
 impl Drop for DrawStateWrapper<'_> {
     fn drop(&mut self) {
-        if let Some(orphaned) = &mut self.orphan_lines {
-            let (text, bars): (Vec<_>, Vec<_>) = self
-                .state
-                .lines
-                .drain(..)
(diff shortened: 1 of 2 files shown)


spoutn1k:
```rs
use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget};

fn main() {
    let display_manager = MultiProgress::new();
    let progress: ProgressBar = {
        let bar = ProgressBar::with_draw_target(None, ProgressDrawTarget::stderr());
        let bar = display_manager.add(bar);

        bar
    };

    progress.println("This should print once");
}
```

Minimal program reproducing the issue.

spoutn1k:
The failed check comes from a license change that is out of my control. Should I ignore or roll in a fix @djc ?

djc:
> The failed check comes from a license change that is out of my control. Should I ignore or roll in a fix @djc ?

Submitted fixes for these in #674.

src/draw_target.rs
```diff
@@ -398,13 +398,13 @@ impl std::ops::DerefMut for DrawStateWrapper<'_> {
...
-                    .filter(|l| matches!(l, LineType::Text(_) | LineType::Empty))
-                    .cloned(),
-            );
+            let (text, bars): (Vec<_>, Vec<_>) = self
+                .state
+                .lines
+                .drain(..)
+                .partition(|l| matches!(l, LineType::Text(_) | LineType::Empty));
+            self.state.lines = bars;
+            orphaned.extend(text);
```
djc, line 407: I'd like to avoid the intermediate allocations from `drain().partition()`. Can we just write this as a `for`-loop that directly pushes into `self.state.lines` and `orphaned` depending on the condition?

spoutn1k: @djc ping on this one

djc: I guess this is a case where'd want `Vec::extract_if()`. Given that that's still unstable, I still think we can avoid allocating the intermediate `Vec` for `orphaned` since we can `push()` onto that directly while draining `lines`, right?

src/draw_target.rs
```diff
@@ -398,13 +398,23 @@ impl std::ops::DerefMut for DrawStateWrapper<'_> {
 impl Drop for DrawStateWrapper<'_> {
     fn drop(&mut self) {
         if let Some(orphaned) = &mut self.orphan_lines {
-            let (text, bars): (Vec<_>, Vec<_>) = self
+            // Filter out the lines that do not contain progress information
+            // Store the filtered out lines in orphaned
+            let bars: Vec<_> = self
```
djc, line 403: I don't think this is a clear way of writing it. Prefer:

```rust
let mut lines = Vec::new();
for line in self.state.lines.drain(..) {
    match &line {
         LineType::Text(_) | LineType::Empty => orphaned.push(line),
         _ => lines.push(line),
    }
}
self.state.lines = lines;
```

djc (approved):

djc:
Thanks!

spoutn1k:
Thank you !

PR #673, opened 2024-12-02
add tracing-indicatif create to integration list

emersonford:
I wrote this integration between tracing and indicatif a while back, which sees a fair amount of use now, and thought it would be nice to call it out as another integration a la indicatif-log-bridge.


djc (approved):
Makes sense to me, thanks!

CI failures are unrelated, addressed in:

- #674

emersonford:
@djc rebased on top of #674 fixes :)

PR #679, opened 2024-12-12
Implement ProgressBar::enable_steady_tick_with_hook for adding custom tick behaviors

chris-laplante:
See #578 for motivation


djc:
I don't know, this seems like a lot of complexity when we don't really have a good use case right now (the original reporter was able to resolve their issue in another way)...

mostthingsweb:
> I don't know, this seems like a lot of complexity when we don't really have a good use case right now (the original reporter was able to resolve their issue in another way)...

We can close it for now then if you'd like. I don't feel strongly about it, and can always revisit it if a real need arises.

PR #682, opened 2025-01-12
Remove unnecessary spinner display in multi examples

shuntaka9576:
SSIA

|before|after|
|---|---|
|before|after

### DIFF
--- a/examples/multi.rs
+++ b/examples/multi.rs
@@ -48,8 +48,6 @@ fn main() {
         let m = m.clone();
         let pb2 = pb2.clone();
         threads.push(thread::spawn(move || {
-            let spinner = m.add(ProgressBar::new_spinner().with_message(i.to_string()));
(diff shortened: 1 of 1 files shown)


djc (approved):
I guess it was intentional in some way, but this change seems like an improvement to me.

@chris-laplante what do you think?

(It does look like there is a leftover warning you should fix.)

chris-laplante:
> I guess it was intentional in some way, but this change seems like an improvement to me.
> 
> @chris-laplante what do you think?
> 
> (It does look like there is a leftover warning you should fix.)

I agree it was probably intentional (perhaps to test whether spinners with `enable_steady_tick` behave correctly with `MultiProgress`). I'm OK with removing it.

shuntaka9576:
@djc 
> (It does look like there is a leftover warning you should fix.)

Sorry about that - fixed the leftover warning.

PR #683, opened 2025-01-13
Only get draw_target-width when we actually draw

jaheba:
When calling `BarState::draw`, there is a call to `self.draw_target.width()`.

However, that operation can be expensive and the width is not needed, if there is no draw-target.

Thus, this PR delays the calling `.width()` if there is not target to draw to.

For comparison, I've used this minibenchmark:

```rust
    let bar = ProgressBar::new_spinner();

    for n in 0..1_000_000 {
        bar.set_message(format!("{}", n));
    }
    bar.finish();
```

Using the latest release it runs in about 400ms, whilst the fixed version uses about 100ms.

The issue surfaced with uv, when resolving many packages: https://github.com/astral-sh/uv/issues/10384

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -351,6 +351,14 @@ impl Drawable<'_> {
             } => draw_state.draw_to_term(term_like, last_line_count),
         }
     }
+
+    pub(crate) fn width(&self) -> Option<u16> {
+        match self {
+            Self::Term { term, .. } => Some(term.size().1),
+            Self::Multi { state, .. } => state.width(),
+            Self::TermLike { term_like, .. } => Some(term_like.width()),
+        }
+    }
 }
 
 pub(crate) enum LineAdjust {
--- a/src/state.rs
+++ b/src/state.rs
@@ -183,8 +183,6 @@ impl BarState {
     }
 
     pub(crate) fn draw(&mut self, mut force_draw: bool, now: Instant) -> io::Result<()> {
-        let width = self.draw_target.width();
-
         // `|= self.is_finished()` should not be needed here, but we used to always draw for
         // finished progress bars, so it's kept as to not cause compatibility issues in weird cases.
         force_draw |= self.state.is_finished();
@@ -192,6 +190,9 @@ impl BarState {
             Some(drawable) => drawable,
             None => return Ok(()),
         };
+
+        // Getting the width can be expensive; thus this should happen after checking drawable.
+        let width = drawable.width();
 
         let mut draw_state = drawable.state();
 


src/state.rs
```diff
@@ -183,26 +183,28 @@ impl BarState {
...
         // finished progress bars, so it's kept as to not cause compatibility issues in weird cases.
         force_draw |= self.state.is_finished();
-        let mut drawable = match self.draw_target.drawable(force_draw, now) {
-            Some(drawable) => drawable,
-            None => return Ok(()),
-        };
 
-        let mut draw_state = drawable.state();
+        // First check whether there is something to be drawn.
+        if self.draw_target.drawable(force_draw, now).is_none() {
```
jaheba, line 191: This double call to `drawable` is sort of ugly in my eyes.

However, the returned `drawable` prevents us from calling `self.draw_target.width()`, since the mutable borrow is still active.

jaheba:
@djc Tests pass locally now.

jaheba:
Actually, this is not working, since `self.draw_target.drawable(...)` will set the time of latest update, and then second call to `.drawable` can return `None`.

jaheba:
Should (hopefully) work now.

The contrast in speed is actually bigger when using `hyperfine --show-output`. The fixed version takes ~120ms, while the current version takes ~700ms.

src/state.rs
```diff
@@ -183,16 +183,17 @@ impl BarState {
     }
 
     pub(crate) fn draw(&mut self, mut force_draw: bool, now: Instant) -> io::Result<()> {
-        let width = self.draw_target.width();
-
         // `|= self.is_finished()` should not be needed here, but we used to always draw for
         // finished progress bars, so it's kept as to not cause compatibility issues in weird cases.
         force_draw |= self.state.is_finished();
+
```
djc, line 189: Nit: please avoid unrelated cosmetic changes.

src/draw_target.rs
```diff
@@ -351,6 +351,14 @@ impl Drawable<'_> {
             } => draw_state.draw_to_term(term_like, last_line_count),
         }
     }
+
+    pub(crate) fn width(&self) -> Option<u16> {
+        match self {
+            Drawable::Term { term, .. } => Some(term.size().1),
```
djc, line 357: Nit: please use `Self` instead of `Drawable` here.

djc (approved):
LGTM modulo some small nits.

Would be nice if you can squash your commits.

Thanks for these performance wins. Do you have more in the pipeline? Would you like a release once these are merged?

jaheba:
Thanks for the review!

I've addressed your comments.

I don't have any more changes (besides #684) planned. I wanted to look into some basic benchmarking to see if there are any other possible performance improvements.

PR #684, opened 2025-01-13
Make tab extension lazy

jaheba:
Running the snippet of #683 a big chunk of the time is spent in allocating a new string in `TabExpandedString`. Delaying the expanding of tabs reduces the runtime of ~100ms to ~60ms.

The idea is simple: Rather than replacing `\t` with spaces immediately, this is done once `expanded` is actually called.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::cell::OnceCell;
 use std::io;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


chris-laplante:
Makes sense to me, thanks! Just kicked of CI to verify.

src/state.rs
```diff
@@ -346,21 +347,20 @@ pub(crate) enum TabExpandedString {
     NoTabs(Cow<'static, str>),
     WithTabs {
         original: Cow<'static, str>,
-        expanded: String,
+        expanded: OnceCell<String>,
```
djc, line 350: Can we just use `Option` instead of `OnceCell` for this? Seems simpler.

djc (commented):
Thanks, that looks like a nice win!

Also, would be nice if you can squash your commits.

src/state.rs
jaheba: No, I don't think so: We don't have `mut` access in `expanded`. AFAIK, that's essentially [what `OnceCell` is there for](https://old.reddit.com/r/rust/comments/126qaai/after_years_of_work_and_discussion_once_cell_has/jebnzf9/).

djc (approved):

tgross35:
This change seems to have made `ProgressStyle` non-`Sync`, we are getting some failures like https://github.com/rust-lang/libm/actions/runs/13008267806/job/36279959248?pr=483. I can probably work around this in my application, but if the change wasn't intentional, would it be possible to change to `OnceLock`?

djc:
@tgross35 ah, sorry about that. Would you be able to submit a PR, including a test for `ProgressStyle` being `Sync`?

tgross35:
Sure! Done at https://github.com/console-rs/indicatif/pull/694

PR #685, opened 2025-01-15
Make `ProgressBar:set_tab_with` take `&self`

jaheba:
Currently, `ProgressBar:set_tab_with` is the only method that requires `&mut self`.

However, there is no reason that it needs exclusive access, and thus this can be just `&self`.


djc (approved):

djc:
Thanks!

Issue #688, opened 2025-01-20
Allow to manually redraw the progress bar


ClementNerma:
Hi!

I've been using this crate a lot and it's been amazing, thanks for making it <3

I only have one problem currently, which is that the progress bar doesn't update visually when position changes happen too quickly. Here is an example:

```rust
use indicatif::ProgressBar;

static LEN: u64 = 10000;

fn main() {
    let pb = ProgressBar::new(LEN);

    for _ in 0..LEN {
        pb.inc(1);
        // pb.tick();
    }

    std::thread::sleep(std::time::Duration::from_secs(10000));
}
```

If you run this program, you will see when we reach the `sleep` portion that the program only has a part of the progress bar filled. Every time you run the program, the final position is a little bit different.

Uncommenting the `pb.tick();` line worsens the problem as the bar gets stuck on position 10 / 10000 (I understand the problem with the previous code why not sure why for this one?)

So I'd like to be able to manually refresh the progress bar and redraw it. Something like `pb.redraw()` or ` pb.refresh()`. It would obviously greatly increase latency and this example code would become a lot slower, but that would solve the problem I have.

I didn't find a way to do this currently ; would it be possible to add this feature?

Many thanks :)

jaheba:
Hi,

not 100% sure I understand your issue. But have you had a look at `enable_steady_tick`?

That will decouple rendering of the progress bar from your calls to `pb.inc()`.

ClementNerma:
Hmm seems like [`enable_steady_tick`] works fine. Not really sure why.

But basically I'd like to *manually* redraw the progress bar without having a separate thread. I'd like the output to be refreshed synchronously when the progress bar changes.

ClementNerma:
Would it be possible to add a method that allows to force re-rendering of the progress bar? To ensure it is completely synchronized with its internal state.

jaheba:
Can you talk a bit about your use-case? I'm curious to learn why the intended ways don't match your needs.


---

In my view it should be fine to add a `.force_draw()` method or something akin to it to indicatif, but that's @djc decision to make.

There is also suspend, which is essentially just a force draw:

```rust
    for _ in 0..LEN {
        pb.inc(1);
    }

    pb.suspend(|| {});
```

Alternatively you can use `console::Term` as a draw target to enforce rendering on `pb.tick()`:

```rust
    pb.set_draw_target(ProgressDrawTarget::term_like(Box::new(Term::stderr())));

    for _ in 0..LEN {
[...]

PR #689, opened 2025-01-21
Add ProgressBar::force_draw

jaheba:
Allow users to force a redraw of the progress bar.

This should normally not be needed, since there are more idiomatic ways to ensure that the progress bar is up to date.

Also, I've noticed that other ways to enforce redrawing (`.suspend()`) can lead to flickering, while `.force_draw()` does not.

Adresses: #688


djc (approved):

djc:
Thanks!

PR #690, opened 2025-01-21
Add `dec` and `dec_length` to `ProgressBar`

jaheba:
Add analgous methods for `inc` and `inc_length` which are atomic, in contrast to `set_position` and `set_length`.


src/progress_bar.rs
```diff
@@ -289,6 +298,10 @@ impl ProgressBar {
     pub fn inc_length(&self, delta: u64) {
         self.state().inc_length(Instant::now(), delta);
     }
+    /// Decrease the length of the progress bar
```
djc, line 301: Nit: empty line before this method, please.

src/state.rs
```diff
@@ -107,6 +107,13 @@ impl BarState {
         self.update_estimate_and_draw(now);
     }
 
+    pub(crate) fn dec_length(&mut self, now: Instant, delta: u64) {
```
djc, line 110: Nit: for consistency, let's order `dec_length()` after `inc_length()` to match the ordering on all the related methods.

djc (approved):

jaheba:
Addressed comments.

djc:
Thanks!

Issue #663, opened 2024-10-22
per_sec does not respect width specifier


rlee287:
Reproducer (using indicatif 0.17.8, default feature set):

```rust
use indicatif::{ProgressBar, ProgressStyle};

fn main() {
    let pb_style = ProgressStyle::default_bar()
        .template("{bar:16} | {pos:>6}/{len:6} | {per_sec:>8}")
        .unwrap();

    let pb = ProgressBar::new(100000).with_style(pb_style);

    for _ in 0..1000 {
        std::thread::sleep(std::time::Duration::from_secs_f64(0.01));
        pb.inc(100);
    }
    pb.finish();
}
```

The progress bar that's left after the program runs:

`████████████████ | 100000/100000 | 9,631.6687/s`

Despite `per_sec` specifying a width of 8 characters, the final `per_sec` display is 12 characters wide.

Compressed asciinema recording of the above program running on my machine:

[pb_per_sec.cast.gz](https://github.com/user-attachments/files/17483190/pb_per_sec.cast.gz)

djc:
Have you tried main? IIRC there have been some improvements in this area.

rlee287:
The same problem occurs when I specify a Git dependency for indicatif using the default commit (branch `main`, commit 5396704f8e9fe781d21a5475c8d3af292d0155fe as of time of writing).

PR #696, opened 2025-03-18
Use width to truncate `HumanFloatCount` values

### LINKED ISSUE #485: Unnecessary hard coded precision in `HumanFloatCount`

ReagentX:
- Use `width` to truncate `HumanFloatCount` values 
  - Normally float truncation is done by using `.#`, but in `indicatif` `.` is reserved for specifying a style string.
  - Resolves #485 and #663

Examples:

`"{spinner:.green} [{elapsed}] [{bar:.blue}] {human_pos}/{human_len} ({per_sec:0}, ETA: {eta})"`:

image

`"{spinner:.green} [{elapsed}] [{bar:.blue}] {human_pos}/{human_len} ({per_sec:4}, ETA: {eta})"`:

image

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -373,6 +373,9 @@ mod tests {
         assert_eq!("1,234.1", format!("{:.1}", HumanFloatCount(1234.1234321)));
         assert_eq!("1,234.12", format!("{:.2}", HumanFloatCount(1234.1234321)));
         assert_eq!("1,234.123", format!("{:.3}", HumanFloatCount(1234.1234321)));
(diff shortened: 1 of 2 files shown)


djc:
Thanks!

PR #697, opened 2025-03-19
Feat/cs/human count short

### LINKED ISSUE #637: Feature Req.: Human readable `per_sec` rate
I am using `indicatif` to process lots of data, and it's great. My programme does hundreds of thousands to millions of iterations of second. The `{per_sec}` template will show me the rate to 4 decimal places (e.g. `120,399,714.0565/s`). That's too many numbers.

Can we get a `per_sec_human` template variable which will print something like `120 M/s` in this case?

ReagentX:
- Resolve #637 

Example: 

`"{spinner:.green} [{elapsed}] [{bar:.blue}] {human_pos_short}/{human_len_short} ({per_sec:4}, ETA: {eta})"`

image

`"{spinner:.green} [{elapsed}] [{bar:.blue}] {human_pos_short:3}/{human_len_short:3} ({per_sec:4}, ETA: {eta})"`

image


ReagentX:
This includes changes from #696 but can be implemented without them if required.

ReagentX:
> I'm inclined to think that the human_ prefix gives us a certain amount of latitude to change things

That would break existing templates; this PR retains existing functionality and only adds new features. Old templates will behave the same as how they did before.

> The assertion style with `outcome, input` seems unidiomatic

I agree; I just copied the design of what was already in the repo for both tests and code to maintain consistency.

ReagentX:
I didn't want to change any existing behavior unless that was specifically requested.

djc:
IMO this behavior should become the default for `human_` values, with an easy way to opt out.

ReagentX:
I can just flip the logic for when the alternative formatter is enabled, are there any other design details I should include?

djc:
> I can just flip the logic for when the alternative formatter is enabled, are there any other design details I should include? It sounds like `human_` as a prefix doesn't really describe any of the behavior, it's more like `formatted_` and `shorthand_`.

Flipping the behavior sounds good to me. I don't see much of an issue with the `human_` prefix, and we're kind of stuck with it now, anyway.

chris-laplante:
> > I can just flip the logic for when the alternative formatter is enabled, are there any other design details I should include? It sounds like `human_` as a prefix doesn't really describe any of the behavior, it's more like `formatted_` and `shorthand_`.
> 
> Flipping the behavior sounds good to me. I don't see much of an issue with the `human_` prefix, and we're kind of stuck with it now, anyway.

(sorry for the radio silence - I was on vacation) 

I am also fine with flipping the behavior.

djc:
Please rebase on top of main. Here's what I'm looking for:

- No new template keys
- Default formats for (some) `human_` keys now always show 3 significant digits (if 100+)
- Potentially custom ways to format `human_` keys, which are explained in the documentation

I don't see new formatting options (`$` or `#.`??) explained in the docs.

ReagentX:
I added logic to maintain 3 digits by default; I am not sure what the rest of your comment is referring to.

src/format.rs
```diff
@@ -170,48 +170,122 @@ impl fmt::Display for HumanCount {
...
-        let frac_trimmed = frac_part.trim_end_matches('0');
-        if !frac_trimmed.is_empty() {
-            f.write_char('.')?;
-            f.write_str(frac_trimmed)?;
         }
         Ok(())
     }
 }
 
+const HUMAN_FLOAT_COUNT_THRESHOLDS: [(f64, &str); 5] = [
```
djc, line 281: I don't think we need both this and `HUMAN_COUNT_TRESHOLDS`, presumably we can drive the float value from the integer one?

ReagentX: The types are different, so they have to be defined separately unless I want to cast them every time the formatter runs. I could do something like this:

```rust
macro_rules! define_thresholds {
    ($(($num:expr, $suffix:expr)),*) => {
        const HUMAN_COUNT_THRESHOLDS: [(u64, &str); _] = [
            $(($num, $suffix),)*
        ];

        const HUMAN_FLOAT_COUNT_THRESHOLDS: [(f64, &str); _] = [
            $(($num as f64, $suffix),)*
        ];
    };
}

define_thresholds!(
    (1_000_000_000_000, "T"), // trillion
    (1_000_000_000, "B"),     // billion
    (1_000_000, "M"),         // million
    (1_000, "k"),             // thousand
    (0, "")                   // no suffix
);
```

but that feels a lot less maintainable than just having two arrays.

ReagentX:
I updated the documentation in my PR: https://github.com/console-rs/indicatif/pull/697/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759

I didn't add the `#`, that is a Rust [stdlib feature](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.alternate) that `indicatif` was already using: https://github.com/console-rs/indicatif/pull/697/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759

The `indicatif` template string is not the same as [`fmt`](https://doc.rust-lang.org/std/fmt/index.html); what happens in `indicatif` is it builds a `fmt` based on the provided `TemplatePart`. In cases where we want to enable the alternate formatter, we have to add a new `key` so that `format_state()` knows what format string to build.

PR #699, opened 2025-03-21
`ProgressStyle` enable/disable colors based on draw target

### LINKED ISSUE #698: [Feature Request] `ProgressStyle`: enable/disable colorized output based on draw target
Hello! This is a follow up to #86:

> It seems that the default implementation draws to stderr, but whether the output is colorized is dependent on whether stdout is a tty, so if you do something like `foo > log`, you get progress bars, but they're uncolored.

... which is a bit surprising for programs that intend their stdout to be piped but also use stderr for progress reporting. Right now I'm manually doing a `set_colors_enabled(colors_enabled_stderr())` to convince `indicatif` to print colors.

I'm wondering if you'd be interested in a PR to make it configurable for `ProgressStyle`, now that `console` has features like [`for_stderr`] for differentiating color support on stdout vs. stderr.

[`for_stderr`]: https://docs.rs/console/0.15.11/console/struct.Style.html#method.for_stderr

How I imagine this to work:

1. Have a `ProgressStyle::for_stderr()` that explicitly sets color detection for its `TemplatePart`s; this could be a public API or just `pub(crate)`;
2. Within `ProgressBar::set_style()`, implicitly call `ProgressStyle::for_stderr()` based on whether the bar's `ProgressDrawTarget` is `TermTarget::Stderr` (I'm not sure if doing this makes sense for other types of target though)

Thank you for this crate!

tonywu6:
Resolves #698 

A `ProgressBar` using stderr as the draw target will now implicitly use stderr to determine if colors are enabled (by configuring its `ProgressStyle`).

Previously, this was not configured, so the default behavior was to detect colors using stdout no matter what.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -7,7 +7,7 @@ use std::time::Duration;
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::Instant;
 
-use console::Term;
+use console::{Term, TermTarget};
 #[cfg(target_arch = "wasm32")]
 use web_time::Instant;
 
@@ -125,6 +125,16 @@ impl ProgressDrawTarget {
             TargetKind::Term { ref term, .. } => !term.is_term(),
             TargetKind::Multi { ref state, .. } => state.read().unwrap().is_hidden(),
             _ => false,
+        }
+    }
+
+    /// This is used in progress bars to determine whether to use stdout or stderr
+    /// for detecting color support.
+    pub(crate) fn is_stderr(&self) -> bool {
+        if let TargetKind::Term { ref term, .. } = self.kind {
+            matches!(term.target(), TermTarget::Stderr)
+        } else {
+            false
         }
     }
 
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -157,6 +157,11 @@ impl ProgressBar {
     ///
     /// This does not redraw the bar. Call [`ProgressBar::tick()`] to force it.
     pub fn set_style(&self, style: ProgressStyle) {
+        let style = if self.state().draw_target.is_stderr() {
+            style.for_stderr()
+        } else {
+            style
+        };
         self.state().set_style(style);
     }
 
--- a/src/style.rs
+++ b/src/style.rs
@@ -164,6 +164,15 @@ impl ProgressStyle {
         Ok(self)
     }
 
+    /// Specifies that the progress bar is intended to be printed to stderr
+    ///
+    /// The progress bar will determine whether to enable/disable colors based on stderr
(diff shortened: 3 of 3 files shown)


src/draw_target.rs
```diff
@@ -128,6 +128,16 @@ impl ProgressDrawTarget {
...
     }
 
+    /// This is used in progress bars to determine whether to use stdout or stderr
+    /// for detecting color support.
+    pub(crate) fn is_stderr(&self) -> bool {
+        if let TargetKind::Term { ref term, .. } = self.kind {
+            matches!(term.target(), TermTarget::Stderr)
+        } else {
+            false
+        }
```
djc, line 138: Nit: please write this as a `match` with match ergonomics:

```rust
match &self.kind {
     TargetKind::Term { term, .. } => ..,
     _ => false,
}
```

src/style.rs
```diff
@@ -164,6 +164,15 @@ impl ProgressStyle {
         Ok(self)
     }
 
+    /// Specifies that the progress bar is intended to be printed to stderr
+    ///
+    /// The progress bar will determine whether to enable/disable colors based on stderr
+    /// instead of stdout. Under the hood, this uses [`console::colors_enabled_stderr`].
+    pub(crate) fn for_stderr(mut self) -> Self {
+        self.template.set_for_stderr();
```
djc, line 172: Let's inline `Template::set_for_stderr()` here, and make it take `&mut self` (and return nothing).

src/style.rs
```diff
@@ -623,6 +632,22 @@ impl Template {
             }
         }
     }
+
+    fn set_for_stderr(&mut self) {
+        for part in &mut self.parts {
+            if let TemplatePart::Placeholder {
```
djc, line 638: Please use early continue like this:

```rust
let (style, alt) = match part {
    TemplatePart::PlaceHolder { style, alt_style, .. } => (style, alt_style),
    _ => continue,
};
```

djc (commented):
Thanks, this looks like a nice improvement!

src/progress_bar.rs
```diff
@@ -156,7 +156,10 @@ impl ProgressBar {
     /// Overrides the stored style
     ///
     /// This does not redraw the bar. Call [`ProgressBar::tick()`] to force it.
-    pub fn set_style(&self, style: ProgressStyle) {
+    pub fn set_style(&self, mut style: ProgressStyle) {
+        if self.state().draw_target.is_stderr() {
```
djc, line 160: Sorry, I wrote another comment here which seems to have gotten lost.

Can you avoid the duplicate `self.state()` calls in favor of binding `let state = self.state()`?

tonywu6: ✅

djc (approved):

PR #704, opened 2025-05-20
Swtch number-prefix dep to unit-prefix

kimono-koans:
number-prefix is unmaintained and does not include license information within its crate. This PR patches number_prefix to use the maintained fork.

See: https://github.com/ogham/rust-number-prefix/pull/8

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,9 @@ readme = "README.md"
 exclude = ["screenshots/*"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -13,9 +13,11 @@ readme = "README.md"
 exclude = ["screenshots/*"]
 
 [dependencies]
-console = { version = "0.15", default-features = false, features = ["ansi-parsing"] }
+console = { version = "0.15", default-features = false, features = [
```
djc, line 16: Please avoid unrelated formatting changes.

PR #705, opened 2025-05-21
[style] Improve integer completion percentage calculation for accurate representation (358/359 is NOT 100%)

mitrandir77:
By default rust float formatting doesn't round down - in fact they use
round_ties_even udner the hood: https://doc.rust-lang.org/std/primitive.f32.html#method.round_ties_even

For percentage calculations this is not great - in particular we shouldn't show 100% when the operation is not fully complete.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -1055,4 +1055,21 @@ mod tests {
         assert_eq!(&buf[2], "bar");
         assert_eq!(&buf[3], "baz");
     }
+
+    #[test]
+    fn test_pct_precision() {
+        // seeing something that's not finished as 100% hurts my eyes
+        const WIDTH: u16 = 80;
+        let pos = Arc::new(AtomicPosition::new());
+        pos.set(358);
+        let state = ProgressState::new(Some(359), pos);
+
+        let style = ProgressStyle::default_bar()
+                .template(
+                    &format!("{{pos}}/{{len}} {{percent}}%"),
+                ).unwrap();
+        let mut buf = Vec::new();
+        style.format_state(&state, &mut buf, WIDTH);
+        assert_eq!(&buf[0], "358/359 100%");
+    }
 }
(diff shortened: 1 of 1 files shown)


mitrandir77:
@djc 
> now 1/359 is going to be 0% which seems about equally bad
I don't think it's equally bad: 1/359 means that the work is not 1% complete yet. 0% seems correct.

I believe rounding down consistently is an improvement and de-facto industry standard. Almost every progress reporting logic I've seen before was using integer division for computing percentage - which made rounding down the default behaviour.

djc:
I'm still not convinced. For all past indicatif users, this hasn't been enough of a problem to remark upon. IMO conceptually rounding is the least surprising behavior.

joshkehn:
> I'm still not convinced. For all past indicatif users, this hasn't been enough of a problem to remark upon.

Is this proposal logically incorrect though? 

I've never seen a progress indicator say 100% complete when there is work still on the table. Showing "100% (358/359)" is a bad UX bug at best and will more likely waste lots of time debugging.

edward-shen:
As a workaround, you can use [`percent_precise`](https://docs.rs/indicatif/latest/indicatif/#:~:text=with%203%20fraction%20digits) as a format key if you would rather round to 3 decimal places instead.

mitrandir77:
@edward-shen `percent_precise` exhibits the same confusing behaviour: given enough items it will start showing not fully complete operations as 100%. (I'll prep unit test showcasing that)

Showing 100% on incomplete progress is confusing and bad ux.

chris-laplante:
> > (aside: I also wonder if I should just create an `indicatif-zoo` or `indicatif-extras` crate or something that includes things we don't want to support in the main crate, but that are oft requested?)
> 
> Suggest first making a list in an issue so we can review the evidence of how often different things come up.

Absolutely, will do. I have plenty of things to do before I go creating another crate (like my neglected work on `MultiProgress`) as well :)

djc:
> What if we added `percent_floor` and `percent_precise_floor`? It's yet more format keys, but the case of `floor` vs `round` seems to not have a consensus.

I could live with that if you think it's valuable.

PR #722, opened 2025-07-09
Fixes issue #580

G0rocks:
Here is a new and improved pull request which contains 2 commits.
The first one is from when I was mostly gaining understanding of the code and the second one contains the changes I implemented.

Here is the old pull request:
https://github.com/console-rs/indicatif/pull/721

This should fix issue #580

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -288,17 +288,20 @@ impl ProgressState {
 
     /// The expected ETA
     pub fn eta(&self) -> Duration {
+        // If done set ETA to 0
         if self.is_finished() {
             return Duration::new(0, 0);
         }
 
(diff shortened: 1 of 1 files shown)


G0rocks:
Great, yeah no stress from my part.
Meanwhile, since you started the workflow (I have very little experience with those) and there was a failed run from `cargo fmt --all -- --check` I looked up that it suggests formatting changes and I implemented the suggested changes so now hopefully the PR should pass the workflow.

G0rocks:
Apparently the `cargo fmt` workflow does not suggest all the suggestions the first time around, since it had different suggestions this time but I implemented those.

PR #724, opened 2025-07-27
Make ProgressBar::index public

strowk:
This could be useful for some user crates, for example potentially can help solve the issue with order of child progress bars that exists in this crate - https://github.com/emersonford/tracing-indicatif/blob/main/src/pb_manager.rs#L217-L221

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -619,7 +619,7 @@ impl ProgressBar {
     }
 
     /// Index in the `MultiState`
-    pub(crate) fn index(&self) -> Option<usize> {
+    pub fn index(&self) -> Option<usize> {
         self.state().draw_target.remote().map(|(_, idx)| idx)
     }
 


strowk:
Thanks for your response, @chris-laplante .

Let me give you a bit of background why I thought this might be useful.

There is a crate that I have found convenient to use - https://github.com/emersonford/tracing-indicatif .
It allows to reduce some boilerplate around managing progress bars. It does have, however, one bug, that I demonstrate in this test - https://github.com/strowk/tracing-indicatif/commit/cf0cb0a8a3a9a693f4f555c0638d452f1bfe0237#diff-a0262fea108af44a98cc49e5eb72641963dd816513f2d0a22eb738fcfed55ebeR161-R181

Without access to `index()` method, this test would fail and results would look like this:
```
foo{}
--> child2{}
--> child1{}
```

I.e the order of progress bars would be wrong (child 2 would be before child 1). This is a known issue in this crate, which I have managed to fix (test passes in that linked revision) by using custom version of indicatif (from which this PR is created) and by calling index() - https://github.com/strowk/tracing-indicatif/commit/cf0cb0a8a3a9a693f4f555c0638d452f1bfe0237#diff-a646224308d682586cd49b1953007c405ab373195b6e3c08ed2e1ae0cbe17a52R239-R241

Now, as you can see in that bit above, this works as a combination of index() and insert:
```rust
let parent_index = parent_pb.index().unwrap_or(0);
self.mp.insert(
    parent_index + siblings_offset,
    pb_span_ctx.progress_bar.take().unwrap(),
)
```

, which in theory could've work like this:

```rust
self.mp.insert_after_with_offset(
    parent_pb,
    siblings_offset,
    pb_span_ctx.progress_bar.take().unwrap(),
)
```

, given if indicatif supported a method `insert_after_with_offset`, that would be similar to this -  https://github.com/console-rs/indicatif/blob/e1f410d3a29f1a96a373377ad7e377df04e285c8/src/multi.rs#L140-L142

```rust
    pub fn insert_after(&self, after: &ProgressBar, pb: ProgressBar) -> ProgressBar {
        self.internalize(InsertLocation::After(after.index().unwrap()), pb)
    }
```

, but also take offset to add to index. This could be alternative approach that would also fix the issue that tracing-indicatif.

Now, I do not actually know whether author of tracing-indicatif would agree with either of those honestly :) 
The author of that crate did note that tracking index of progress bars is not trivial in this comment - https://github.com/emersonford/tracing-indicatif/pull/21#issuecomment-3131759367
I have found that the index is already available in indicatif and thought maybe it can be utilized and it did work in practice. I wouldn't claim that the tests I have done are exhaustive enough to say that it works in absolutely all cases, but the tool for which I needed this, hasn't been showing any weird order of progress bars in those several days that I use it (about dozen times a day).

I also can in principle live without either of these being fixed in upstreams, my forks of both indicatif and indicatif-tracing kind of work for me at the moment. So you are free to close this as well, no big.

chris-laplante:
@djc While I'm here... I'd be fine with merging this, if maybe we tweaked the docs for `ProgressBar::index` to make it clear that it is an implementation detail, not actually the real "index". What do you think?

djc:
> @djc While I'm here... I'd be fine with merging this, if maybe we tweaked the docs for `ProgressBar::index` to make it clear that it is an implementation detail, not actually the real "index". What do you think?

Do you want to submit another PR that tries to document any caveats around this? Maybe it could/should be `#[doc(hidden)]`?

chris-laplante:
> > @djc While I'm here... I'd be fine with merging this, if maybe we tweaked the docs for `ProgressBar::index` to make it clear that it is an implementation detail, not actually the real "index". What do you think?
> 
> Do you want to submit another PR that tries to document any caveats around this? Maybe it could/should be `#[doc(hidden)]`?

Yes, I'd be happy to do that.

chris-laplante:
@djc should I just merge this one in first? Or should I just supersede it in the new PR? I have no preference personally.

djc:
> @djc should I just merge this one in first? Or should I just supersede it in the new PR? I have no preference personally.

I'd like to read your documentation proposal before merging anything. But you could submit a PR that has this branch as a base maybe?

PR #742, opened 2025-10-31
Add ProgressBar::set_elapsed

sunshowers:
Similar to the other `set_` methods -- it's possible to do this as `bar.clone().with_elapsed` but not very intuitive what that means.


djc (commented):
Can you explain your use case?

sunshowers:
The lint issue looks like something new in Rust 1.91, unrelated to the PR.

djc (approved):
Thanks!

Would you like a release with this?

PR #748, opened 2025-11-18
Fix duration after finish

### LINKED ISSUE #557: Duration becomes 0 after calling `finish()`
If I have got a `ProgressBar` with `ProgressStyle` that uses a template like `"… {elapsed}/{duration} …"`. As usual, I repeatedly call `.inc(…)` and finally `.finish()`. However, I am typically getting this kind of printout:

- 0s/4s
- 1s/4s
- 2s/4s
- 3s/4s
- 4s/4s
- 4s/0s

That is, the estimated total duration makes plenty of sense, except that after calling `finish` it drops to 0. I would expect that `duration` would reflect the real final duration when the progress bar is finished.

I guess the problem is around lines 292–297 of `src/state.rs`:

    pub fn duration(&self) -> Duration {
        if self.len.is_none() || self.is_finished() {
            return Duration::new(0, 0);
        }
        self.started.elapsed() + self.eta()
    }

It seems to be explicitly returning 0 when `is_finished()`. Is this intentional?

luolong:
When initializing progress bar with style that exposes template variable `{duration}`, the duration is set to 0 after progress bar finishes.

This is is unexpected and makes progress bar styles other than 'ProgressFinish::AndClear' display wrong duration.

This will add Duration parameter to `crate::state::State::Done*` variants and uses it to display duration if curret state of the progress bar is one of the Done states.

Fixes #557

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -842,7 +842,7 @@ mod tests {
         );
     }
 
-    fn state_started_at(start_at: Instant) -> BarState {
+    fn state_started_at(started: Instant) -> BarState {
         BarState {
             draw_target: ProgressDrawTarget::hidden(),
             on_finish: ProgressFinish::default(),
@@ -853,8 +853,8 @@ mod tests {
                     len: None,
                     tick: 0,
                     status: Status::InProgress,
-                    started: start_at,
-                    est: Estimator::new(start_at),
+                    started,
+                    est: Estimator::new(started),
                     message: TabExpandedString::NoTabs("".into()),
                     prefix: TabExpandedString::NoTabs("".into()),
                 }


src/state.rs
```diff
@@ -689,6 +697,30 @@ mod tests {
...
+                        started: start_at,
+                        est: Estimator::new(start_at),
+                        message: TabExpandedString::NoTabs("".into()),
+                        prefix: TabExpandedString::NoTabs("".into()),
+                    }
+                },
+                tab_width: DEFAULT_TAB_WIDTH,
+            }
+        }
+    }
```
djc: This seems pretty unidiomatic. Please turn it into a `state_started_at()` helper function that is defined just below the tests that use it.

src/state.rs
```diff
@@ -312,10 +313,11 @@ impl ProgressState {
 
     /// The expected total duration (that is, elapsed time + expected ETA)
     pub fn duration(&self) -> Duration {
-        if self.len.is_none() || self.is_finished() {
-            return Duration::new(0, 0);
+        match (&self.status, self.len) {
```
djc: Derive `Copy` for `Status` so that we don't need the `&` and `*`.

src/state.rs
```diff
@@ -806,6 +808,61 @@ mod tests {
...
+        assert_eq!(
+            state.state.duration(),
+            duration,
+            "Expected duration: {}, actual: {}",
+            duration_to_secs(duration),
+            duration_to_secs(state.state.duration())
+        );
+    }
+
+    fn state_started_at(start_at: Instant) -> BarState {
```
djc: Nit: rename the argument to `started` so that it doesn't need to be renamed in the `ProgressState` initializer.

PR #754, opened 2025-11-29
Expose the current tab width

alexisfontaine:
I'm developing a build workflow for my project, which includes several components that each spawn multiple sub-tasks. I want to increase the tab width to represent the depth of each task:

```rust
fn new_sub_task (parent: &ProgressBar) {
  let child = ProgressBar::new(0).set_tab_width(parent.tab_width() + 2);

  /* ... */
}
```


djc (approved):
Seems fine, thanks.

PR #760, opened 2026-01-23
More reactive updates with `enable_steady_tick()`

MoreDelay:
I have a use case where I create two progress bars where the first bar shows the overall progress whereas the second bar shows the progress of the current task, each also showing the elapsed time. A task can take a few seconds or up to a few minutes. This is exactly the kind of situation where one would use `enable_steady_tick()` to let the overall progress bar update its timer.

However I noticed that on task switches, the overall progress bar lags behind ever so slightly as the tick gets ignored on its increment, and thus waits until the specified interval has passed. I would like to have the overall progress bar update at the same time as the task progress bar. In this situation, I would either have to make due with a stale elapsed timer, or implement my own ticker thread logic around the bar.

So instead, I propose to add a configuration option to specify how the steady ticker should react to calls to `tick()`.

The first commit changes the behavior such that the bar gets updated immediately on a call to `tick()`, even with a running ticker. The second commit adds a configuration option to get back the old behavior if preferred. This is currently a breaking API change on `enable_steady_tick()`, so I'm not sure if you may want this to be a new method instead.

Feel free to leave suggestions, or whether or not these changes are in scope for this project.

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -175,9 +175,6 @@ impl ProgressBar {
     ///
     /// When this is enabled a background thread will regularly tick the progress bar in the given
     /// interval. This is useful to advance progress bars that are very slow by themselves.
-    ///
-    /// When steady ticks are enabled, calling [`ProgressBar::tick()`] on a progress bar does not
-    /// have any effect.
     pub fn enable_steady_tick(&self, interval: Duration) {
         // The way we test for ticker termination is with a single static `AtomicBool`. Since cargo
         // runs tests concurrently, we have a `TICKER_TEST` lock to make sure tests using ticker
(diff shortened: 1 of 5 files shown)


src/progress_bar.rs
```diff
@@ -732,32 +783,60 @@ struct TickerControl {
...
+                        let passed = now - last_tick;
+                        passed >= interval
+                    }
+                },
+            };
+
+            if do_tick {
+                state.tick(now);
+                last_tick = Some(now);
+            }
```
djc: I think this is confused about what it wants to be. If you want `.tick()` to be responsive, IMO it should not affect a running ticker at all.

PR #766, opened 2026-02-05
Bump version to 0.18.4

djc:
In preparation for a release.


tgross35:
It doesn't look like this ever got released, there is no tag and it is missing on 0.18.4. Would it be possible to do that?

djc:
Yeah, I was waiting for other stuff to land. I've published it now:

- https://github.com/console-rs/indicatif/releases/tag/0.18.4

PR #784, opened 2026-04-03
Add `ProgressBar::visual_index` and clarify meaning of `ProgressBar::index`.

chris-laplante:
`MultiProgress::insert` lets users insert a progress bar at a specified position within the `MultiProgress`. Before this PR, there was no way (besides keeping track manually) to know the position within the `MultiProgress` of a `ProgressBar`.

The `ProgressBar::index` method would be tempting if `pub`, but not appropriate because it refers to the opaque index within the `MultiState`'s data structures. This PR also clarifies its documentation.

Something else that this PR enables is that users can now write their own `insert_after_with_offset` (as suggested in https://github.com/console-rs/indicatif/pull/724#issuecomment-3145629928), i.e. `mp.insert(other_pb.visual_index() + offset, pb)`

Supersedes #782 and #724

### DIFF
--- a/src/progress_bar.rs
+++ b/src/progress_bar.rs
@@ -633,7 +633,13 @@ impl ProgressBar {
         self.state().tab_width
     }
 
-    /// Index in the `MultiState`
+    /// Opaque index in the `MultiState`
+    ///
+    /// This value is an implementation detail and does not necessarily correspond to visual
+    /// position on the screen.
+    ///
+    /// It is an index into `MultiState::members`, which is not in any particular order (due to
+    /// reclaiming of indices from the `MultiState::free_set`).
     pub(crate) fn index(&self) -> Option<usize> {
         self.state().draw_target.remote().map(|(_, idx)| idx)
     }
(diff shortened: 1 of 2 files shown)


src/progress_bar.rs
```diff
@@ -644,6 +644,15 @@ impl ProgressBar {
         self.state().draw_target.remote().map(|(_, idx)| idx)
     }
 
+    /// If this [`ProgressBar`] is a member of a [`MultiProgress`](crate::MultiProgress), then return visual position
+    /// on screen. Otherwise, `None`.
+    pub fn visual_index(&self) -> Option<usize> {
```
djc, line 649: Do you think there's value in wrapping this up in a newtype `VisualIndex`? In some future we might be able to then use that in `InsertLocation` variants...

Or maybe, should this just be named `index` and we rename the private method to something else (`members_index()` maybe)?

chris-laplante: I like the second option - I'll work on it later this weekend/Monday. In that case, I would still suggest we newtype the "members index" type, so that we don't accidentally get them confused internally either.

chris-laplante:
Refactored, now uses newtypes to represent the opaque index (`MultiStateIndex`). There is also a small newtype vec utilizing the index, `MultiStateMembers`.

chris-laplante:
@djc gentle ping :)

Issue #595, opened 2023-09-26
Document behavior of `MultiProgress`, `suspend`, and `ProgressBar` scopes


chris-laplante:
The way `suspend` works with `MultiProgress` is not super obvious, which leads to issues like #594. Document how it works and include some common pitfalls & solutions.

chris-laplante:
Documentation has been added to `MultiProgress`.

PR #786, opened 2026-04-03
multi: document common MultiProgress bugs + behavior (#595)

chris-laplante:
I'm not super happy about adding the embed-doc-image dep, but I wasn't sure if there was another way to add the images.


Cargo.toml
```diff
@@ -14,6 +14,7 @@ exclude = ["screenshots/*"]
 
 [dependencies]
 console = { version = "0.16", default-features = false, features = ["ansi-parsing", "std"] }
+embed-doc-image = "0.1.4"
```
djc: Yeah, I definitely think we don't want to add this as a library dependency. Not sure whether dev-dependencies are sufficient for rustdoc stuff?

Maybe this could be embedded as an SVG, with an `include_bytes!()` or similar?

chris-laplante: I tried dev-dependencies but it doesn't work, unfortunately. Looks like we can do it with a build.rs (either manually, or with a tool like https://docs.rs/rustdoc-assets/latest/rustdoc_assets/). 

I'm leaning towards just using a normal `` tag and pull from github directly, i.e. https://raw.githubusercontent.com/console-rs/indicatif/refs/heads/cpl/multi-progress-docs/screenshots/mp-drop-after.png.

djc: `build.rs` doesn't seem great either. Have you checked out inline SVG options?

chris-laplante: I'll give it a try now.

chris-laplante:
@djc svg works well, thanks!

djc (approved):
Nice work! The BAD/GOOD with emoji looks a little LLM-generated -- not sure if it is, but maybe write that a little differently to make it more humane?

mostthingsweb:
> Nice work! The BAD/GOOD with emoji looks a little LLM-generated -- not sure if it is, but maybe write that a little differently to make it more humane?

It’s not lol, although I totally get that it looks like that way. I’ll rewrite that part tomorrow/Monday :)

chris-laplante:
I'd say this is ready now

djc (approved):

src/multi.rs
```diff
@@ -13,7 +13,133 @@ use crate::progress_bar::ProgressBar;
...
+/// ```
+///
+/// Running the example, we see that it is broken:
+///
+#[doc = include_str!("../screenshots/mp-drop-before.svg")]
+///
+/// The issue is that at the end of each loop iteration, `pb` is dropped. Conceptually `MultiProgress` only maintains
+/// weak references to `ProgressBar`s. At the next loop iteration, `suspend` causes `MultiProgress` to clear the screen.
+/// `MultiProgress`' "zombie" algorithm ensures the dropped (zombie) bar is not left behind on the screen. But
+/// `MultiProgress` can't reconstitute the 'finish' state (i.e. "DONE!" text), since the bar no longer exists.
```
djc, line 109: I wonder how hard it would be to improve on this. Maybe we could use an enum to hold on to a strong reference if the finish state requires it?

chris-laplante: Yeah, I think it would be relatively straightforward. I probably won't have time to do it though :/.

djc (approved):

Issue #677, opened 2024-12-04
RFC: Redesign how progress bars are added to `MultiProgress`


chris-laplante:
I believe the current `MultiProgress` API is error-prone and too easy to "hold wrong", and would like to suggest a small-ish redesign.

`MultiProgress` exists primarily to coordinate draws of multiple `ProgressBar`s. 

With the current API, you create a `ProgressBar` and add it to the `MultiProgress`. These are two discrete steps:

```rust
let mp = MultiProgress::new();

// possibly other pbs added here too...

let pb = ProgressBar::new(100);
// <--- (1)  danger!
mp.add(pb);
```

The problem is that it is very easy to accidentally cause the `ProgressBar` to be drawn before it is added to the `MultiProgress`, i.e. at the spot marked `(1)`. The most common offender is `enable_steady_tick` (for spinners). But lots of other methods (e.g. `set_message`) can cause a draw too. 

If the `MultiProgress` has been drawn by the time we get to `(1)`, then it is already game over: `pb` has likely corrupted the screen. 

Off the top of my head, the most recent occurrences I've seen are listed below (I could have sworn there are more, but I can't find them):
- https://github.com/console-rs/indicatif/issues/502
- https://github.com/console-rs/indicatif/issues/645

The safe "fix" is to ensure `ProgressBar`s are only mutated/customized after being added to `MultiProgress`. For example:

```rust
let mp = MultiProgress::new();

// possibly other pbs added here too...

let pb = mp.add(ProgressBar::new(100));
// no problem touching `pb` here
```

However, I think we can do better by ensuring the `ProgressBar`s destined for a `MultiProgress` are never prematurely touched in the first place. We could replace `MultiProgress::add` with `add_pb`/`add_progress` and `add_spinner` methods to `MultiProgress`. (Similarly, `insert_spinner_before`, `insert_pb_after`, etc.)

```rust
let mp = MultiProgress::new();

// possibly other pbs added here too...

let pb: ProgressBar = mp.add_pb(100);

// no possibility of touching pb before we should
```

Obviously, this breaks backwards compatibility. But I think it will save users from a lot of troubleshooting and annoyance in the long term.

Another option would be to create some kind of proxy with a `ProgressBar`-like API, but which defers customizations/draws until added to a `MultiProgress`. Then change `MultiProgress` to only take those instead of `ProgressBar`. Example:

```rust
// other name options: DeferredProgressBar, ProgressBarBuilder, ???
[...]

PR #787, opened 2026-04-03
Add MultiBar for safe deferred progress bar configuration

jgrund:
## Summary

Implements the design from #677: introduces `MultiBar`, a builder type that captures progress bar configuration without triggering any draws. This prevents the common footgun where a `ProgressBar` draws to stderr before being added to a `MultiProgress`, corrupting the terminal.

- **`MultiBar`** — plain data struct with `with_*` builder methods mirroring `ProgressBar`. No draws until added to `MultiProgress`.
- **`MultiProgressInput`** — `#[doc(hidden)]` `#[non_exhaustive]` enum with `From<ProgressBar>` and `From<MultiBar>`, enabling backwards-compatible `impl Into<MultiProgressInput>` on all `MultiProgress::add/insert` methods.
- **Materialization** — when added to `MultiProgress`, creates a hidden `ProgressBar`, applies config via `with_*` (no draws), sets the remote draw target, then starts steady tick last.

### Before (footgun)
```rust
let pb = ProgressBar::new(100);
pb.set_message("downloading"); // draws to stderr before mp knows about it!
pb.enable_steady_tick(Duration::from_millis(100)); // spawns thread drawing to stderr!
mp.add(pb); // too late, screen already corrupted
```

### After (safe)
```rust
let pb = mp.add(
    MultiBar::new(100)
        .with_message("downloading")
        .with_steady_tick(Duration::from_millis(100))
);
// pb is a fully configured ProgressBar, safely managed by mp
```

Existing code using `mp.add(ProgressBar::new(...))` continues to work unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -9,6 +9,7 @@ use crate::draw_target::{
     visual_line_count, DrawState, DrawStateWrapper, LineAdjust, LineType, ProgressDrawTarget,
     VisualLines,
 };
+use crate::multi_bar::MultiProgressInput;
 use crate::progress_bar::ProgressBar;
 #[cfg(all(target_arch = "wasm32", feature = "wasmbind"))]
 use web_time::Instant;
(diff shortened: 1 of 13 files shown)


djc:
I think the core of this idea is still good. I'm not convinced we want to make semver-compatible changes to do it, and I'm not sure I want to invest much time in reviewing a PR that it looks like is mostly LLM-generated without much human oversight. Separately, not sure if the name is right since the general concept seems applicable to other ways of building up a `ProgressBar`, too.

jgrund:
Hi, 

> I think the core of this idea is still good. I'm not convinced we want to make semver-compatible changes to do it.

Do you want to make API breaking changes instead here?

> I'm not sure I want to invest much time in reviewing a PR that it looks like is mostly LLM-generated without much human oversight.

Yes, this code was mostly generated with LLM, but I designed the overall solution and am  manually reviewing it / testing it / updating it etc...

>  Separately, not sure if the name is right since the general concept seems applicable to other ways of building up a ProgressBar, too.

Would you prefer this become a builder / default API for `ProgressBar` overall?

djc:
> > I think the core of this idea is still good. I'm not convinced we want to make semver-compatible changes to do it.
> 
> Do you want to make API breaking changes instead here?

Sorry, I mistyped. I don't want to make semver-**in**compatible changes for this.

> > I'm not sure I want to invest much time in reviewing a PR that it looks like is mostly LLM-generated without much human oversight.
> 
> Yes, this code was mostly generated with LLM, but I designed the overall solution and am manually reviewing it / testing it / updating it etc...

Okay, that seems fine. Suggest you write the PR description by hand in the future.

> > Separately, not sure if the name is right since the general concept seems applicable to other ways of building up a ProgressBar, too.
> 
> Would you prefer this become a builder / default API for `ProgressBar` overall?

Yes, I think that might be better.

jgrund:
> Yes, I think that might be better.

Looking over this a bit, would there be any functional difference between a builder and the 

ProgressBar::with_* API (if we fill in the bits that don't currently have a ProgressBar::set_* equivalent)?

I.E. we could have a `ProgressBarBuilder` (which would be more idiomatic) but that then seems redundant against the existing with_* methods. 

One thing I could think of:

- Create `ProgressBarBuilder`
- Have the with_* methods be marked as deprecated
- Create a MultiProgress::add_bar method (which only takes builder)
- Have the MultiProgress::add method be marked as deprecated

That keeps things semver compatible and gives a clear migration path to users. Thoughts?

djc:
> One thing I could think of:
> 
> * Create `ProgressBarBuilder`
> * Have the with_* methods be marked as deprecated
> * Create a MultiProgress::add_bar method (which only takes builder)
> * Have the MultiProgress::add method be marked as deprecated
> 
> That keeps things semver compatible and gives a clear migration path to users. Thoughts?

Given the prevalence of the old `with_` methods and the lack of utility for anyone not using `MultiProgress` I'd prefer to avoid explicit deprecation of the old `with_` methods. But otherwise your plan sounds good to me, especially deprecating `MultiProgress::add()` in favor of a new method that adds the builder directly (not sure `add_builder()` is the best name though).

jgrund:
FYI: Latest push adds `ProgressBarBuilder` and `MultiProgress::register{,_at,_from_back,_before,_after}` as the way to use it. Previous ways are marked as deprecated.

src/multi.rs
```diff
@@ -162,12 +221,38 @@ impl MultiProgress {
...
         let idx = state.insert(location);
         drop(state);
 
-        pb.set_draw_target(ProgressDrawTarget::new_remote(self.state.clone(), idx));
+        let draw_target = ProgressDrawTarget::new_remote(self.state.clone(), idx);
+        pb.set_draw_target(draw_target);
+        pb
+    }
+
+    fn internalize_builder(
```
djc, line 234: Nit: let's call this `internalize()`.

src/progress_bar_builder.rs
```diff
@@ -0,0 +1,436 @@
+use std::borrow::Cow;
```
djc, line 1: Let's call this module `builder`.

src/progress_bar_builder.rs
```diff
@@ -0,0 +1,436 @@
...
+    prefix: Option<Cow<'static, str>>,
+    position: Option<u64>,
+    elapsed: Option<Duration>,
+    tab_width: Option<usize>,
+    on_finish: Option<ProgressFinish>,
+    steady_tick: Option<Duration>,
+}
+
+// ProgressStyle doesn't implement Debug, so we print all other fields
+impl fmt::Debug for ProgressBarBuilder {
```
djc, line 97: Nit: this should go below the inherent `impl` block.

src/progress_bar_builder.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ProgressBarBuilder")
```
djc, line 99: Nit: use `let Self { .. } = self;` to make sure we keep track of all the fields.

src/progress_bar_builder.rs
```diff
@@ -0,0 +1,436 @@
...
+            .field("elapsed", &self.elapsed)
+            .field("tab_width", &self.tab_width)
+            .field("on_finish", &self.on_finish)
+            .field("steady_tick", &self.steady_tick)
+            .finish_non_exhaustive()
+    }
+}
+
+impl ProgressBarBuilder {
+    fn base() -> Self {
```
djc, line 113: This should be `Default` impl instead (presumably it can be derived).

src/progress_bar_builder.rs
```diff
@@ -0,0 +1,436 @@
...
+        let pb = mp.register(ProgressBarBuilder::new(42));
+        assert_eq!(pb.length(), Some(42));
+    }
+
+    #[test]
+    fn builder_no_length() {
+        let mp = MultiProgress::with_draw_target(ProgressDrawTarget::hidden());
+        let pb = mp.register(ProgressBarBuilder::no_length());
+        assert_eq!(pb.length(), None);
+    }
```
djc, line 323: This looks like a lot of tests for a fairly shallow API wrapper. Please exercise some judgement in which tests add value/coverage.

tests/render.rs
```diff
@@ -515,13 +512,52 @@ fn ticker_drop() {
     );
 }
 
+#[test]
+fn builder_with_steady_tick() {
```
djc, line 516: Why do we need a separate test for this?

tests/render.rs
```diff
@@ -791,6 +827,37 @@ fn progress_style_tab_width_unification() {
     assert_eq!(in_mem.contents(), "OK    OK");
 }
 
+#[test]
+fn builder_tab_width_propagates_to_style() {
```
djc, line 831: Why do we need a separate test for this?

src/multi.rs
```diff
@@ -71,6 +72,51 @@ impl MultiProgress {
...
+    /// The builder is materialized into a [`ProgressBar`] whose draw target is
+    /// a remote target intercepted by this [`MultiProgress`]. The resulting bar
+    /// is positioned below all other bars currently in the [`MultiProgress`].
+    ///
+    /// Using a [`ProgressBarBuilder`] avoids the footgun of configuring a
+    /// [`ProgressBar`] (which triggers draws to stderr) before adding it to a
+    /// [`MultiProgress`]. See [#677] for details.
+    ///
+    /// [#677]: https://github.com/console-rs/indicatif/issues/677
+    pub fn register(&self, builder: ProgressBarBuilder) -> ProgressBar {
```
djc, line 86: Not sure I love the `register()` name/prefix. How about using `push()` and/or `insert()` like the std collections types?

@chris-laplante thoughts?

chris-laplante: What about `make`, `make_after`, `make_before`, etc.? or alternatively, `realize`, `realize_after`, and so on,

djc: Those don't really make sense to me? The question is what relation makes sense between a `MultiProgress` and a `ProgressBar` (builder). The `MultiProgress` doesn't `make` a `ProgressBar`, right, nor does it really `realize` it?

chris-laplante: The thought was that a `MultiProgress` makes a `ProgressBar` from a builder. Or alternatively, a builder is 'realized' to produce the `ProgressBar`.

chris-laplante: Maybe it makes sense to use `build`, `build_after`, etc.:

```rust
let builder = ProgressBarBuilder::/* whatever */;
let pb = mp.build(builder);

// or
let pb = mp.build_after(builder, &existing_pb);
```

djc: Ah, I understand what're you getting at now. I think I'd still prefer to emphasize the relation between the `MultiProgress` and the resulting `ProgressBar` over the "realization" from builder to `ProgressBar`.

chris-laplante: I see - the issue is that we already have `insert` and friends (https://docs.rs/indicatif/latest/indicatif/struct.MultiProgress.html#method.insert_after). For the builder flavors, what about `insert_with`, `insert_after_with`, etc.?

jgrund: Yes, I had to choose a new verb as all the others were overloaded with some other meaning (even `_with` has a specific connotation in this codebase).

Let me know what you land on and I'll make the change.

Issue #759, opened 2026-01-12
set_move_cursor plays badly with println


Hawk777:
[The documentation for `set_move_cursor`](https://docs.rs/indicatif/0.18.3/indicatif/struct.MultiProgress.html#method.set_move_cursor) says:

> This can reduce flickering, but do not enable it if you intend to change the number of progress bars.

It does not say anything about other restrictions, only not to use it when changing the number of bars. Consider the following code:

```rust
use indicatif::{MultiProgress, ProgressBar};
use std::thread::sleep;

fn main() {
	const SEC: std::time::Duration = std::time::Duration::from_secs(1);
	let pb = ProgressBar::new_spinner();
	let mp = MultiProgress::new();
	mp.add(pb.clone());
	mp.set_move_cursor(true);
	pb.set_message("This is quite a long message");
	sleep(SEC);
	pb.tick();
	sleep(SEC);
	mp.println("foo");
	sleep(SEC);
	pb.tick();
	sleep(SEC);
}
```

This code does not, at any point, change the number of progress bars—it creates exactly one progress bar that lasts for the program’s whole lifetime, and adds it to the `MultiProgress` as early as it possibly can, before setting the message or doing anything else with either object. However, the output of this program looks like this, once the `println` runs:

```
foohis is quite a long message                                                                                                                                                                
⠉ This is quite a long message                                                                                                                                                                
```

I’m using urxvt 9.31.

Hawk777:
Using `set_move_cursor(true)` also seems to cause `MultiProgress::clear` to not clear all the lines properly. This code:

```rust
use indicatif::{MultiProgress, ProgressBar};
use std::thread::sleep;

fn main() {
	const SEC: std::time::Duration = std::time::Duration::from_secs(1);
	let pb = ProgressBar::new_spinner();
	let pb2 = ProgressBar::new_spinner();
	let pb3 = ProgressBar::new_spinner();
	let pb4 = ProgressBar::new_spinner();
	let mp = MultiProgress::new();
	mp.add(pb.clone());
	mp.add(pb2.clone());
	mp.add(pb3.clone());
	mp.add(pb4.clone());
	mp.set_move_cursor(true);
	pb.set_message("This is quite a long message");
	pb2.set_message("This is quite a long message");
	pb3.set_message("This is quite a long message");
	pb4.set_message("This is quite a long message");
	sleep(SEC);
	pb.tick();
	pb2.tick();
	pb3.tick();
	pb4.tick();
[...]

PR #790, opened 2026-04-12
Clear the rest of line after drawing to the terminal

### LINKED ISSUE #4808: Progress bar UI corruption with `.set_move_cursor()`
Sometimes this will happen on my Mac (macOS v26, Ghostty v1.3, fish v4.6):

Image

Note that:
- Ideally the progress bar should be cleaned up after the "installed" message appears (see the final line).
- Ideally the progress bar should update normally even when the input shortens (see the second line and the second-to-last line).

I'm pretty sure this surfaced after https://github.com/rust-lang/rustup/pull/4786, otherwise we actually haven't changed much about the progress bar (modulo the extra alignment #4768 but should be irrelevant).

I would like to know if anyone else has experienced this and esp. @ChrisDenton if you have seen anything broken on Windows since.

ChrisDenton:
EDIT: Updated to avoid changing the `TermLike` trait.

This will help fix https://github.com/rust-lang/rustup/issues/4808


djc (commented):
Could we add a default implementation to keep this semver-compatible?

ChrisDenton:
Sure. Although the only compatible default would be a no-op, which may be slightly surprising.

djc:
> Sure. Although the only compatible default would be a no-op, which may be slightly surprising.

What would the behavior be for downstream impls that get the default impl?

ChrisDenton:
The behaviour for downstream impls that use the default will be the same as in the current version (which doesn't clear the end of the line). The only difference is if they make use of the `TermLike` trait themselves and then only if they use the new method and expect it to work.

Btw, in the future I would love it if `move_cursor` was set to `true` by default (or always be `true` and remove the option) but that would likely warrant a breaking version change (and some confidence that we've found all the bugs).

djc:
> The behaviour for downstream impls that use the default will be the same as in the current version (which doesn't clear the end of the line). The only difference is if they make use of the `TermLike` trait themselves and then only if they use the new method and expect it to work.

Yes, I meant, how would downstreams that have their own `TermLike` impl would be impacted?

> Btw, in the future I would love it if `move_cursor` was set to `true` by default (or always be `true` and remove the option) but that would likely warrant a breaking version change (and some confidence that we've found all the bugs).

See discussion here: https://github.com/console-rs/indicatif/issues/759#issuecomment-3745083502. In my mind the cost to doing semver-incompatible releases in this crate is relatively high, so if at all possible I'd really prefer to improve it without making semver-incompatible changes.

ChrisDenton:
> Yes, I meant, how would downstreams that have their own `TermLike` impl would be impacted?

The only place this crate makes use of the implementation is in this PR where the default no-op implementation would be the same as the old behaviour. I.e. if `move_cursor` is set to `true` they will not have this bug fix but it will otherwise work.

src/term_like.rs
```diff
@@ -73,6 +77,10 @@ impl TermLike for Term {
         self.clear_line()
     }
 
+    fn clear_to_end_of_line(&self) -> io::Result<()> {
+        self.clear_chars(0)
```
djc: Why is this the right thing to be doing? `clear_chars()` is documented to "[c]lear the last `n` characters of the current line", why does passing 0 here have the desired effect? Maybe add a comment to explain this a little?

ChrisDenton:
After thinking about this some more, and considering the desire to not have a breaking release, I think this PR should be changed to do this unconditionally:

https://github.com/console-rs/indicatif/blob/3b0294a0d65befc63af106a4b59a97ab6993760b/src/draw_target.rs#L581-L582

That would entirely avoid the need for changing `TermLike`. It does move the cursor to the end but that's fine so long as we also avoid writing a new line character on the next write. So remove this:

https://github.com/console-rs/indicatif/blob/3b0294a0d65befc63af106a4b59a97ab6993760b/src/draw_target.rs#L572-L574

djc:
Sounds good!

ChrisDenton:
Ok, done and also tested via rustup's use of this crate.

djc (approved):

djc:
Good stuff, thanks!

PR #791, opened 2026-04-12
Upgrade rand to 0.10.1

ChrisDenton:
Should address https://rustsec.org/advisories/RUSTSEC-2026-0097


djc:
Only in the lockfile please.

djc (approved):

PR #799, opened 2026-04-29
Fix height and filler when wrapped at the middle of wide characters

### LINKED ISSUE #798: Rendering failure when wrapped at wide characters such as CJK
For example, "Aあいうえお" is 11 columns, one ASCII and 5 CJK wide characters. When displaying this text to 10 columns terminal, since CJK wide characters cannot be broken, it will be broken before the last wide character and result in "Aあいうえ\nお". The first line can contain only 9 columns.

It looks like `DrawState.draw_to_term` computes the width correctly, but it doesn't take this additional space caused by line wrapping into account.

Test:
```rust
#[test]
fn cjk_msg_at_the_end_wrap() {
    let in_mem = InMemoryTerm::new(10, 10); // Width 10
    let pb = ProgressBar::with_draw_target(
        Some(10),
        ProgressDrawTarget::term_like(Box::new(in_mem.clone())),
    )
    .with_style(ProgressStyle::with_template(" {msg}").unwrap());

    pb.set_message("こんにちは"); // 5 CJK characters, width 10
    pb.tick();

    println!("{}", in_mem.moves_since_last_check());

    // Formatted string is " こんにちは" (11 columns).
    // Line 1: " こんにち" (9 columns)
    // Line 2: "は" (2 columns)
    assert_eq!(
        in_mem.contents(),
        " こんにち\nは"
    );
}
```

kojiishi:
When a wide character such as CJK appears at the end of wrap with only 1 column available, the line wraps before the character, leaving an empty column.

This patch fixes `wrapped_height` to take them into account, by computing each wrap instead of dividing by the width.

Fixes #798.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -7,7 +7,11 @@ use std::time::Duration;
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::Instant;
 
+#[cfg(feature = "unicode-width")]
+use console::AnsiCodeIterator;
 use console::{Term, TermTarget};
+#[cfg(feature = "unicode-width")]
+use unicode_width::UnicodeWidthChar;
 #[cfg(all(target_arch = "wasm32", feature = "wasmbind"))]
 use web_time::Instant;
 
@@ -556,23 +560,23 @@ impl DrawState {
         let mut real_height = VisualLines::default();
 
         for line in self.lines.iter() {
-            let line_height = line.wrapped_height(term_width);
+            let metrics = line.wrapped_metrics(term_width);
 
             // Check here for bar lines that exceed the terminal height
             if matches!(line, LineType::Bar(_)) {
                 // Stop here if printing this bar would exceed the terminal height
-                if real_height + line_height > term.height().into() {
+                if real_height + metrics.height > term.height().into() {
                     break;
                 }
 
-                real_height += line_height;
+                real_height += metrics.height;
             }
 
             term.write_str(line.as_ref())?;
 
             // clear the line and keep the cursor on the right terminal side so that
             // future writes/prints will happen on the next line
-            let line_filler = line_height.as_usize() * term_width - line.console_width();
+            let line_filler = term_width - metrics.last_line_width;
             term.write_str(&" ".repeat(line_filler))?;
         }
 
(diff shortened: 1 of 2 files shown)


kojiishi:
Updated the commit message to be more descriptive.

src/draw_target.rs
```diff
@@ -656,20 +659,53 @@ pub(crate) enum LineType {
...
+            for ch in substr.chars() {
+                let ch_width = unicode_width::UnicodeWidthChar::width(ch).unwrap_or(0);
+                column += ch_width;
+                if column > width {
+                    num_lines += 1;
+                    column = ch_width;
+                }
+            }
+        }
+        let effective_width = (num_lines - 1) * width + column;
```
djc: I'm not sure `effective_width` is a good name for this? At least it seems confusing that we're multiplying `num_lines` with a `width` and also calling the result a `width`.

src/draw_target.rs
```diff
@@ -656,20 +659,53 @@ pub(crate) enum LineType {
...
+        // The `effective_width` takes such empty columns into account.
+        let str = self.as_ref();
+        let mut num_lines = 1;
+        let mut column = 0;
+        for (substr, is_ansi) in AnsiCodeIterator::new(str) {
+            if is_ansi {
+                continue;
+            }
+            for ch in substr.chars() {
+                let ch_width = unicode_width::UnicodeWidthChar::width(ch).unwrap_or(0);
```
djc: Please import `UnicodeWidthChar` at the top of the module. I think the `unwrap_or(0)` here is a little weird. I think we can do `let Some(ch_width) = .. else { continue };` instead?

src/draw_target.rs
```diff
@@ -656,20 +659,53 @@ pub(crate) enum LineType {
...
-        usize::max(terminal_len, 1).into()
+        (usize::max(terminal_len, 1).into(), unwrapped_width)
     }
 
+    #[cfg(not(feature = "unicode-width"))]
     fn console_width(&self) -> usize {
         console::measure_text_width(self.as_ref())
     }
+
+    #[cfg(feature = "unicode-width")]
```
djc, line 667: I would like to have this function definition be next to (and before) the non-`unicode-width` definition.

src/draw_target.rs
```diff
@@ -656,20 +659,53 @@ pub(crate) enum LineType {
 
 impl LineType {
     fn wrapped_height(&self, width: usize) -> VisualLines {
+        self.wrapped_height_and_effective_width(width).0
+    }
+
+    #[cfg(not(feature = "unicode-width"))]
+    fn wrapped_height_and_effective_width(&self, width: usize) -> (VisualLines, usize) {
```
djc: I find this name a little verbose. Maybe something like `metrics`? Could even introduce a `Metrics` struct for the return type, maybe.

djc:
What are your uses for? Is indicatif being used in Chromium?

src/draw_target.rs
kojiishi: Thank you for the suggestion, I like that. Introduced a `Metrics` struct.

src/draw_target.rs
kojiishi: Done.

src/draw_target.rs
kojiishi: Done, and thank you for the good suggestion again, done.

src/draw_target.rs
kojiishi: Renamed to `width_sum`. How about this?

kojiishi:
> What are your uses for? Is indicatif being used in Chromium?

No, this is for my personal project https://github.com/kojiishi/compare-dir, half for own use and half for learning Rust.

src/draw_target.rs
kojiishi: I got another idea; I made it `last_line_width`, and modified `draw_to_term` accordingly. This saves some CPU cycles with `unicode-width`, adds some cycles when only height is needed without `unicode-width`, but subtle. Does this look clearer?

djc (approved):

djc:
Thanks!

kojiishi:
Thanks!

PR #803, opened 2026-05-05
Fix `ProgressDrawTarget.is_stderr()` for `MultiProgress`

### LINKED ISSUE #802: Colors are not applied to `MultiProgress` if stdout is redirected
Say, you build a command using `MultiProgress`.
```
mycmd
```
shows the progress bars in colors, but:
```
mycmd > /out
```
prints the progress bars without colors.

kojiishi:
When configuring the style, `draw_target.is_stderr()` is used to determine if `stderr` should be used for color capability detection instead of defaulting to `stdout`.

This patch adds support for `MultiProgress` to `draw_target.is_stderr()`.

Fixes #802.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -325,7 +325,7 @@ impl MultiProgress {
     }
 
     pub fn is_hidden(&self) -> bool {
-        self.state.read().unwrap().is_hidden()
+        self.state.read().unwrap().draw_target.is_hidden()
     }
 }
 
@@ -339,7 +339,7 @@ pub(crate) struct MultiState {
     /// Indices to the `draw_states` to maintain correct visual order
     ordering: Vec<usize>,
     /// Target for draw operation for MultiProgress
-    draw_target: ProgressDrawTarget,
+    pub(crate) draw_target: ProgressDrawTarget,
     /// Controls how the multi progress is aligned if some of its progress bars get removed, default is `Top`
     alignment: MultiProgressAlignment,
     /// Lines to be drawn above everything else in the MultiProgress. These specifically come from
@@ -363,7 +363,7 @@ impl MultiState {
     }
 
     pub(crate) fn mark_zombie(&mut self, index: usize) {
-        let width = self.width().map(usize::from);
+        let width = self.draw_target.width().map(usize::from);
 
         let member = &mut self.members[index];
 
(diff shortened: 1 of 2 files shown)


src/multi.rs
```diff
@@ -511,6 +511,10 @@ impl MultiState {
         self.draw_target.is_hidden()
     }
 
+    pub(crate) fn is_stderr(&self) -> bool {
```
djc: Can we make the `draw_target` field `pub(crate)` and drop indirections around it?

djc (commented):
This mostly makes sense to me, thanks.

src/multi.rs
chris-laplante: Agreed, if possible.

kojiishi: Done.

kojiishi:
If you want me to remove other similar indirections, happy to replace them too, please let me know. It looks like `is_hidden()` and `width()` are the candidates. As a follow up patch if you want to separate fix and refactoring, or put into this patch if you prefer.

djc:
> If you want me to remove other similar indirections, happy to replace them too, please let me know. It looks like `is_hidden()` and `width()` are the candidates. As a follow up patch if you want to separate fix and refactoring, or put into this patch if you prefer.

Yes, that would be nice. Can you do two commits, in this order:

- Make `draw_target` public, inline existing methods
- Add the new `is_stderr` check forwarding

kojiishi:
@djc
> Yes, that would be nice. Can you do two commits, in this order:
> 
> * Make `draw_target` public, inline existing methods

Done at https://github.com/console-rs/indicatif/pull/804

> * Add the new `is_stderr` check forwarding

I'm now rebasing this PR.

kojiishi:
@djc Rebased. I think I did it right, but this my first time to send dependent PRs, sorry if something went wrong.

It looks like the diff tab shows both changes, but commits tab looks as expected, and the diff of the 2nd commit is the fix. Once #804 is merged, I think rebasing should remove the 1st commit.

Thank you for your support.

kojiishi:
Ooh, I found you wrote "two commits", not "two PRs". Sorry if doing two PRs wasn't your intention.

* If you're not squashing, this PR should be fine then, please disregard #804.
* If you're squashing, please take #804 first, then this one.

djc (approved):

djc:
Thanks!

kojiishi:
Thank you for your support!

PR #806, opened 2026-05-16
Introduce `Width::set_ambiguous_wide`

kojiishi:
Provides a solution for #497 by allowing callers to opt-in when they're in ambiguous-wide environment.
    
It doesn't cover grapheme cluster issues also discussed in #497.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -2,6 +2,8 @@ use std::collections::HashMap;
 use std::fmt::{self, Formatter, Write};
 use std::mem;
 use std::str::FromStr;
+#[cfg(feature = "unicode-width")]
+use std::sync::atomic::{AtomicBool, Ordering};
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::Instant;
 #[cfg(feature = "unicode-width")]
@@ -844,16 +846,31 @@ pub(crate) static WIDTH: Width = Width::new();
 ///
 /// This structure encapsulates width calculation logic, allowing configuration
 /// of how ambiguous-width characters are treated.
-pub struct Width {}
+pub struct Width {
+    #[cfg(feature = "unicode-width")]
+    is_cjk: AtomicBool,
+}
 
 impl Width {
     const fn new() -> Self {
-        Self {}
+        Self {
+            #[cfg(feature = "unicode-width")]
+            is_cjk: AtomicBool::new(false),
+        }
+    }
+
+    #[cfg(feature = "unicode-width")]
+    fn is_cjk(&self) -> bool {
+        self.is_cjk.load(Ordering::Relaxed) || WIDTH.is_cjk.load(Ordering::Relaxed)
     }
 
     #[cfg(feature = "unicode-width")]
     pub(crate) fn char(&self, c: char) -> usize {
-        c.width().unwrap_or(0) // Make control characters zero-width.
+        match self.is_cjk() {
+            true => c.width_cjk(),
+            false => c.width(),
+        }
+        .unwrap_or(0) // Make control characters zero-width.
     }
 
     #[cfg(not(feature = "unicode-width"))]
@@ -864,7 +881,10 @@ impl Width {
 
     #[cfg(feature = "unicode-width")]
     pub(crate) fn str(&self, s: &str) -> usize {
-        UnicodeWidthStr::width(s)
+        match self.is_cjk() {
+            true => UnicodeWidthStr::width_cjk(s),
+            false => UnicodeWidthStr::width(s),
+        }
     }
 
     #[cfg(not(feature = "unicode-width"))]
@@ -878,7 +898,10 @@ impl Width {
         let mut width = 0;
         for (chunk, is_ansi) in AnsiCodeIterator::new(s) {
             if !is_ansi {
-                width += UnicodeWidthStr::width(chunk);
+                width += match self.is_cjk() {
+                    true => UnicodeWidthStr::width_cjk(chunk),
+                    false => UnicodeWidthStr::width(chunk),
+                };
             }
         }
         width
@@ -889,7 +912,24 @@ impl Width {
     pub(crate) fn ansi_str(&self, s: &str) -> usize {
         measure_text_width(s)
     }
+
+    /// Set to use the alternate width calculation more suited to CJK contexts.
+    /// This is primarily about the [UAX#11] ambiguous characters being treated
+    /// as wide (CJK), but not limited to.
+    /// Please also see the ["cjk" feature flag of the `unicode-width`].
+    ///
+    /// ["cjk" feature flag of the `unicode-width`]: https://docs.rs/unicode-width/latest/unicode_width/#cjk-feature-flag
+    /// [UAX#11]: https://www.unicode.org/reports/tr11/
+    #[cfg(feature = "unicode-width")]
+    pub fn set_default_cjk(is_cjk: bool) {
+        WIDTH.is_cjk.store(is_cjk, Ordering::Relaxed);
+    }
 }
+
+// Serializes tests that modify or rely on the global CJK ambiguous width setting
+// (via `Width::set_cjk`) to avoid flakiness in parallel execution.
+#[cfg(test)]
+pub(crate) static WIDTH_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
 
 #[cfg(test)]
 mod tests {
@@ -1153,6 +1193,7 @@ mod tests {
 
     #[test]
     fn wide_element_style() {
+        let _guard = WIDTH_TEST_LOCK.lock().unwrap();
         set_colors_enabled(true);
 
         const CHARS: &str = "=>-";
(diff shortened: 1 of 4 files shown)


kojiishi:
Note, this PR is on top of #799, and thus if you go to the diff tab, you'll see both changes. You probably want either look at only the 2nd commit in the commits tab, or merge #799 and review the diff tab.

src/width.rs
```diff
@@ -0,0 +1,148 @@
+#[cfg(feature = "unicode-width")]
+use console::AnsiCodeIterator;
+use portable_atomic::AtomicBool;
+#[cfg(feature = "unicode-width")]
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
+
+static IS_AMBIGUOUS_WIDE: AtomicBool = AtomicBool::new(false);
```
djc: I don't love having this as a global `static`. Maybe it could be attached to the draw target?

kojiishi: Yeah, I originally thought to add a property to draw target. But then passing draw targets to all the code paths in `style.rs` seems to need a bit of plumbing. I also found another benefit, a global flag allows apps to change the settings even when they're using `indicatif` indirectly (i.e., using libraries that use `indicatif`).

I haven't tried passing to `style.rs` though. I can try it and see how much plumbing it would need.

kojiishi: I tried this locally. `ProgressStyle.char_width` is problematic. Currently  it's computed in `new`, but it doesn't know draw target yet, until `format_bar`.

Assuming the flag doesn't change often, `format_bar` can compute the `char_width` and cache it, but since `format_bar` needs to be thread-safe, we'll need some tricks there.

The easiest one is `AtomicUsize`, but since it's not clonable, we need to implement `Clone` manually. Also since it's not really a lock, first  a few `format_bar`s may compute the widths until one of them finished computing it and store to the `AtomicUsize`.

Mutex can fully avoid multiple computations, but it adds some complexity too, and is slower than `AtomicUsize`.

Not caching at all (compute width each time in `format_bar`) is also possible, but I suspect you don't want this for the performance reasons, do you?

Which way would you like?

kojiishi: Another idea is to udpate `char_width` when the style is set to `BarState` (via `ProgressBar`). This would mostly work, except when the draw target of `MultiProgress` is changed after progress bars are added. `MultiProgress` doesn't hold references to progress bars, right?

Maybe you have better ideas what is the best way to solve this?

djc:
How would callers decide to set this value, anyway? Are there ways to query the font metrics, or does the caller basically just have to guess what environment it's running in?

kojiishi:
> How would callers decide to set this value, anyway? Are there ways to query the font metrics, or does the caller basically just have to guess what environment it's running in?

It depends on the callers.
* If they know the fonts, they can query the width of one of such characters.
* If they're console programs (most likely for `indicatif` I suppose?), unfortunately console API doesn't provide font information. It should be a guess, or a user setting.

Either way, the bad news is that there's no very easy way to do it. If there's a concrete way, `unicode-width` might have already done it. They just delegated to callers, and we're delegating to our callers, until it reaches someone who knows the font or the end user environment.

That's where a global setting might win. We could check an env var to initialize the default, then users (not callers) can change the settings without modifying their programs.

A few new thoughts from the discussion:
* It is the cached `char_width` that makes it complicated. Moving it to the state and compute it on-demand should be the cleanest way, though the change becomes a bit bigger.
* Even if we make it non-global, it may be a good idea to check an env var to setup the default value, in case callers of `indicatif` didn't do anything, but users want to switch. Maybe a separate, additional enhancement.
* Further, if we're doing this "delegate" chains, it may be a good idea for `unicode-width` itself to check an env var and switch to the `cjk` variant.
  If `unicode-width` likes the idea, then no changes are needed to `indicatif`, unless you like this refactoring (to gather all width logic to one place).

kojiishi:
> * Further, if we're doing this "delegate" chains, it may be a good idea for `unicode-width` itself to check an env var and switch to the `cjk` variant.

...it may not be a good idea, because `unicode-width` is so low-level that branching by a global var might regress the performance. Better to branch at higher-level.

Anyway, I updated the patch, with 2 commits:
1. The 1st commit moves `ProgressStyle.char_width` cache to `BarState`. It's a pure-refactoring without behavior changes.
2. The 2nd commit adds the new flag, to switch between `width` and `width_cjk`.

djc:
> That's where a global setting might win. We could check an env var to initialize the default, then users (not callers) can change the settings without modifying their programs.

Okay, I think I now agree that this should be a process-global (maybe a `static WIDTH_CJK: AtomicBool`?). Sorry for flipflopping...

kojiishi:
> > That's where a global setting might win. We could check an env var to initialize the default, then users (not callers) can change the settings without modifying their programs.
> 
> Okay, I think I now agree that this should be a process-global (maybe a `static WIDTH_CJK: AtomicBool`?). Sorry for flipflopping...

Thanks for reviewing. I just rebased before I saw your reply. I'll revert to the global version.

kojiishi:
Reverted to the global version, and renamed to `WIDTH_CJK`.

src/width.rs
```diff
@@ -0,0 +1,155 @@
+#[cfg(feature = "unicode-width")]
+use console::AnsiCodeIterator;
+use portable_atomic::AtomicBool;
+#[cfg(feature = "unicode-width")]
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
+
+static WIDTH_CJK: AtomicBool = AtomicBool::new(false);
```
djc: Could we have `static WIDTH: Width` and hide the `AtomicBool` inside it as a private field?

src/width.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Helper for calculating text width.
+///
+/// This structure encapsulates width calculation logic, allowing configuration
+/// of how ambiguous-width characters are treated.
+pub struct Width;
+
+impl Width {
+    #[cfg(feature = "unicode-width")]
+    fn is_cjk() -> bool {
+        WIDTH_CJK.load(portable_atomic::Ordering::Relaxed)
```
djc: Nit: please import `Ordering` from `std` (should be the same one from core as used in portable_atomic).

src/width.rs
```diff
@@ -0,0 +1,155 @@
+#[cfg(feature = "unicode-width")]
```
djc: I'd prefer not to introduce a new module for this, just stick this in `style` for now, somewhere near the bottom (but before `tests`).

src/width.rs
```diff
@@ -0,0 +1,155 @@
...
+            if !is_ansi {
+                width += if is_cjk {
+                    UnicodeWidthStr::width_cjk(chunk)
+                } else {
+                    UnicodeWidthStr::width(chunk)
+                };
+            }
+        }
+        width
+    }
```
djc: Let's move all of these methods before `set_cjk()` and `is_cjk()`, and have `set_cjk()` before `is_cjk()`.

src/style.rs
```diff
@@ -824,21 +820,11 @@ fn segment(s: &str) -> Vec<Box<str>> {
     s.chars().map(|x| x.to_string().into()).collect()
 }
 
-#[cfg(feature = "unicode-width")]
```
djc, line 827: I think I'd still like to have (at least) two commits:

- One that introduces the `Width` wrapper
- Another that introduces the `AtomicBool` state inside it

src/width.rs
kojiishi: Sure, so you want to make it like `Width::default().str("....")`, correct? Allow me to confirm:
* That'll involve a lock for each `default()` or `new()`, as `Atomic*` is only for primitives. Locks are a bit slower than `Atomic*`. Is that ok?
* If we're making the measure functions instance functions (not static functions), do you want it to be thread-safe (i.e. do we need to use `AtomicBool` instead of `bool`)? Callers can always create a new instance or clone it for multi-thread access, no?

src/style.rs
kojiishi: Done.

src/width.rs
kojiishi: Done.

src/width.rs
kojiishi: Sounds good, done.

src/width.rs
kojiishi: `AtomicBool` is no longer used, see the previous comment.

src/width.rs
kojiishi: Ok, so I took `default()`/`new()` style, and used `bool` instead of `AtomicBool` assuming that it's not necessary for instances to be thread-safe. Please let me know if you prefer it to be thread-safe.

djc: No, that's not what I had in mind. Have a look at the approach I just pushed to your branch.

(And please have a look at the test failures.)

kojiishi: I see, thanks, sorry for misunderstanding.

I uploaded a new one, hope this time is right.

djc:
You changed a bunch of stuff back from my push (merged the test file, method order/names). Any particular reason?

kojiishi:
> You changed a bunch of stuff back from my push (merged the test file, method order/names). Any particular reason?

Not intentional, I think I merged your changes to my checkout manually to fix the test failures (not knowing how to download yours from github), but it sounds like I missed some, sorry about that. Please go ahead with your own.

djc:
> > You changed a bunch of stuff back from my push (merged the test file, method order/names). Any particular reason?
> 
> Not intentional, I think I merged your changes to my checkout manually to fix the test failures (not knowing how to download yours from github), but it sounds like I missed some, sorry about that. Please go ahead with your own.

Sorry, can you try that again? You should be able to download my commits from GitHub with `git fetch git@github.com:console-rs/indicatif.git ea_ambiguous`, then `git reset --hard FETCH_HEAD` while you have your own branch checked out.

kojiishi:
As I ivnestigated the correct truncations and line wrapping for my own personal projects, I hit some more edge cases, such as Unicode segmentations (I see `indicatif` has an optional feature). So I built my own width calculator for those projects, [unicode-width-utils](https://crates.io/crates/unicode-width-utils).

Instead of maintaining two copies, what do you think about using the crate instead? I tried it to be as lightweight as possible. If you're happy with it, I can update this PR.

djc:
Where else do you want to use that logic? Not sure I love the idea of depending on a new crate with no long-term maintenance experience. I think we could move a bunch of this into the console crate if that would make sense?

I definitely don't want regex as a dependency, even an optional one, though -- that seems way to heavy.

kojiishi:
That makes sense. regex isn't necessary for indicatif, but I guess it's not the only reasons, right?

Sorry for the noise, please move on with your modified version.

djc (approved):

djc:
Thanks for all the iterations!

PR #764, opened 2026-02-03
fix: respect NO_COLOR and TERM=dumb environment variables

shaanmajid:
Fixes #763

Changes `is_hidden()` and `drawable()` to check `term.features().colors_supported()` instead of `term.is_term()`. The `console` crate's `colors_supported()` already checks for `NO_COLOR` and `TERM=dumb`, so this change makes indicatif respect those signals.

## Test plan

Use the MRE from #763 with a path dependency:
```toml
indicatif = { path = "path/to/fix/branch/indicatif" }
```

`script` creates a pseudo-TTY so `isatty()` returns true. Grepping for `1b 5b` (hex `ESC[`) detects ANSI escape sequences. With the fix, no output should appear when `TERM=dumb` or `NO_COLOR=1`.

```console
# Normal TTY - escape sequences expected
script -q /dev/null ./target/release/mre 2>&1 | xxd | grep '1b 5b'
# Output: 00000050: ... 0d1b 5b32 4b ...  (ESC[2K present)

# With TERM=dumb - escape sequences should be suppressed
TERM=dumb script -q /dev/null ./target/release/mre 2>&1 | xxd | grep '1b 5b'
# No output expected

# With NO_COLOR=1 - escape sequences should be suppressed
NO_COLOR=1 script -q /dev/null ./target/release/mre 2>&1 | xxd | grep '1b 5b'
# No output expected
```

PR #808, opened 2026-05-20
fix: respect NO_COLOR and TERM=dumb environment variable

### LINKED ISSUE #807: Unexpected behavior : NO_COLOR hides progress bar
In #764 NO_COLOR and TERM=dumb environment variables are both treated as "I do not want any progress bar".

However, NO_COLOR just means "do not try to display colors" which was already supported thanks to console-rs/console.

**Current behavior :**

Setting NO_COLOR environment variable hides progress bars.

**Expected behavior :**

Setting NO_COLOR environment variable displays monochrome progress bars and texts.
(Worked on version 0.18.3)

ValentinLeTallec:
A rewrite of #764 to fix #807.

I gave a look at the [implementation](https://github.com/console-rs/console/blob/d2d89ad0332e03383e6fe9b4bf0f8612f3c64389/src/unix_term.rs#L24) of [colors_supported](https://github.com/console-rs/console/blob/main/src/term.rs#L97) and extracted only the part about `TERM=dumb`.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -1,4 +1,3 @@
-use std::io;
 use std::ops::{Add, AddAssign, Sub};
 use std::slice::SliceIndex;
 use std::sync::{Arc, RwLock, RwLockWriteGuard};
@@ -6,6 +5,7 @@ use std::thread::panicking;
 use std::time::Duration;
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::Instant;
+use std::{env, io};
 
 #[cfg(feature = "unicode-width")]
 use console::AnsiCodeIterator;
@@ -73,11 +73,16 @@ impl ProgressDrawTarget {
     /// hidden.  This is done so that piping to a file will not produce
     /// useless escape codes in that file.
     ///
-    /// Progress bars will also be hidden if `NO_COLOR` is set or `TERM` is unset/`dumb`.
+    /// Progress bars will also be hidden if `TERM` is unset/`dumb`.
     ///
     /// Will panic if `refresh_rate` is `0`.
     pub fn term(term: Term, refresh_rate: u8) -> Self {
-        if !term.features().colors_supported() {
+        let term_is_dumb_or_unset = match env::var("TERM") {
+            Ok(term) => term == "dumb",
+            Err(_) => true,
+        };
+
+        if !term.is_term() || term_is_dumb_or_unset {
             return Self::hidden();
         }
         Self {


djc (approved):
Seems like an improvement, thanks!

ugur-eren:
Hi @ValentinLeTallec and @djc,

May I ask the reasoning behind returning a hidden `ProgressDrawTarget` when the `TERM` environment variable is not set?

On Windows, native terminals such as Command Prompt, PowerShell, and Windows Terminal typically do not set TERM, even though they fully support and can display progress bars correctly.

In my case, `console::Term::is_term()` returns true and the terminal is detected as attended, but `ProgressDrawTarget::term()` still returns a hidden draw target solely because `TERM` is unset. Was this behavior intentional?

ValentinLeTallec:
It was to match [console's implementation](https://github.com/console-rs/console/blob/d2d89ad0332e03383e6fe9b4bf0f8612f3c64389/src/unix_term.rs#L24). Upon closer inspection they do seem to differentiate unix shells from [windows shells](https://github.com/console-rs/console/blob/d2d89ad0332e03383e6fe9b4bf0f8612f3c64389/src/windows_term/mod.rs#L62).

You could probably fix your issue through a quick PR that change the behavior when not finding ``TERM according to the platform.

djc:
Happy to review/merge something like that! Sorry for the regression -- please add some comments to document this behavior.

ugur-eren:
I can put together a fix, but wanted to check how much of this platform-specific logic should live in `indicatif` vs. `console` before I do.

## Option 1: keep the fix in indicatif
```diff
- let term_is_dumb_or_unset = match env::var("TERM") {
-     Ok(term) => term == "dumb",
-     Err(_) => true,
- };
+ #[cfg(unix)]
+ let term_is_dumb = match env::var("TERM") {
+     Ok(term) => term == "dumb",
+     Err(_) => true,
+ };
+
+ #[cfg(windows)]
+ let term_is_dumb = if term.features().is_msys_tty() {
+     match env::var("TERM") {
+         Ok(term) => term == "dumb",
+         Err(_) => false,
+     }
+ } else {
+     false
+ };
+
+ #[cfg(not(any(unix, windows)))]
+ let term_is_dumb = false;
```

This duplicates platform detection logic that console already owns internally (it does something similar in its own is_a_color_terminal), but it's self-contained and doesn't require any upstream change.

## Option 2: add is_dumb() to console, use it here
Cleaner from indicatif's side and keeps platform logic where it belongs, but means coordinating a console PR/release first.

```diff
- if !term.is_term() || term_is_dumb_or_unset {
-     return Self::hidden();
- }
+ if !term.is_term() || term.features().is_dumb() {
+     return Self::hidden();
+ }
```

Happy to send either :) let me know which you'd prefer.

djc:
I like the idea of doing this in console! Overhead should be fairly minimal.

PR #811, opened 2026-06-18
Make Template public, move from_str to impl FromStr.

### LINKED ISSUE #810: Ability to create template ahead of using it
There's multiple situations when a program would prefer to parse user input as soon as possible including progress bar templates.

it would be beneficial if one can create template ahead of time, store it and then pass to a progress bar later on in comparison to fail somewhere in the middle of the process.

eirnym:
closes #810

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -75,7 +75,7 @@ impl ProgressStyle {
         }
     }
 
-    fn new(template: Template) -> Self {
+    pub fn new(template: Template) -> Self {
         let progress_chars = segment("█░");
         let char_width = width(&progress_chars);
         Self {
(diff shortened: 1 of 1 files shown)


djc:
I made some changes here which I'd like to merge first:

- https://github.com/console-rs/indicatif/pull/813

eirnym:
good, do you have any comments on what's need to be included/excluded, taking in account that your PR doesn't change much?

djc (commented):
Making `ProgressStyle::new()`, `Template` and `Template::set_tab_with()` `pub` seems okay to me.

djc:
Want to rebase this?

eirnym:
Thank you for heads up, I'll do it asap

djc (approved):

PR #813, opened 2026-06-19
style module cleanup

djc:
Some cleanup before we consider merging

- #811

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -472,7 +472,21 @@ struct Template {
 }
 
 impl Template {
-    fn from_str_with_tab_width(s: &str, tab_width: usize) -> Result<Self, TemplateError> {
+    fn set_tab_width(&mut self, new_tab_width: usize) {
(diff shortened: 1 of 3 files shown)


src/style.rs
```diff
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::fmt::{self, Formatter, Write};
 use std::mem;
+use std::str::FromStr;
```
eirnym, line 4: it's in core since Rust 1.6.0

djc: We don't import anything from `core::` so far, so this doesn't seem relevant right now.

djc:
@chris-laplante friendly ping?

djc:
Added a version bump to prepare for a release.

chris-laplante (approved):

chris-laplante:
> @chris-laplante friendly ping?

Sorry I missed this! Looks good to me.

PR #815, opened 2026-06-24
Fix stray comma after minus sign in HumanFloatCount

### LINKED ISSUE #814: HumanFloatCount prints a stray comma after the minus sign for some negatives
`HumanFloatCount`'s `Display` groups the integer digits, but it includes the leading `-` in the grouping, so the sign gets counted as a digit and a comma ends up right after it whenever the digit count is a multiple of 3.

```rust
use indicatif::HumanFloatCount;
println!("{}", HumanFloatCount(-100.0));             // "-,100"      (want "-100")
println!("{}", HumanFloatCount(-100000.0));          // "-,100,000"  (want "-100,000")
println!("{}", HumanFloatCount(f64::NEG_INFINITY));  // "-,inf"      (want "-inf")
```

`-1000.0` happens to look right ("-1,000") only because the sign offset lines up with a real group boundary, which hides it.

the cause is in the `Display` impl in `src/format.rs` - `int_part` still has the `-` when the grouping loop runs. stripping the sign before grouping fixes it. opening a PR with that plus a regression test.

sueun-dev:
Fixes #814

`HumanFloatCount`'s `Display` impl grouped the integer digits over the whole int part including the leading `-`, so the sign was counted as a digit and a comma landed right after it for sign-aligned lengths:

- `HumanFloatCount(-100.0)` → `"-,100"`
- `HumanFloatCount(-100000.0)` → `"-,100,000"`
- `HumanFloatCount(f64::NEG_INFINITY)` → `"-,inf"`

(`-1000.0` only looked right because the sign offset coincides with a real group boundary.)

The fix splits the optional sign off before grouping and writes it first; positive numbers are unchanged.

Added `human_float_count_negative` covering negatives and `-inf`. It fails on the current code (`left "-100"` / `right "-,100"`) and passes with the change. `cargo test`, `cargo fmt --check`, and `cargo clippy` are clean.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -194,8 +194,16 @@ impl fmt::Display for HumanFloatCount {
             Some((int_str, fract_str)) => (int_str.to_string(), fract_str),
             None => (self.0.trunc().to_string(), ""),
         };
-        let len = int_part.len();
-        for (idx, c) in int_part.chars().enumerate() {
+        // Keep the optional sign out of the digit grouping, otherwise the '-'
+        // is counted as a digit and a stray comma lands right after it
+        // (e.g. -100 would render as "-,100").
+        let (sign, digits) = match int_part.strip_prefix('-') {
+            Some(digits) => ("-", digits),
+            None => ("", int_part.as_str()),
+        };
+        f.write_str(sign)?;
+        let len = digits.len();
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #816, opened 2026-06-26
Fix HumanFloatCount dropping the fractional part

mahirhir:
`HumanFloatCount`'s `Display` rounds with `{:.0}` and drops the fractional part (e.g. 1234.9 → "1,234"); for a fraction ≥ 0.5 it's also off by one. This mirrors the precision handling a recent external PR (#815) added to the sibling counter. 58 tests green, 0 regression.


djc (approved):

djc:
Thanks!

PR #817, opened 2026-06-29
Drop screenshots from MultiProgress documentation

djc:
@chris-laplante documented some `MultiProgress` footguns in

- #595 

These added SVG screenshots referenced from docstrings. These screenshots are 61k and 75k respectively; indicatif 0.18.4 came in at a total size of 67k. On reflection, I feel like including these is probably not the right trade-off.

(We currently exclude the `screenshots` directory from packaging, meaning that this is broken in its current state anyway.)


chris-laplante (approved):
That's fine by me. Maybe (after this release) we can just host the images elsewhere and hot-link them into the documentation.

PR #818, opened 2026-07-01
fix: correctly detect dumb terminal on Windows

ugur-eren:
# Fix incorrect dumb-terminal detection on Windows

The current check for whether to hide the progress bar relies on reading the TERM environment variable directly:
```rs
let term_is_dumb_or_unset = match env::var("TERM") {
    Ok(term) => term == "dumb",
    Err(_) => true,
};
```

This treats an unset `TERM` as "dumb" and hides the progress bar.

On Windows, `TERM` is typically not set at all, even in perfectly capable terminals (Windows Terminal, PowerShell, cmd.exe, etc.), so this logic incorrectly hides progress bars on Windows in cases where they should be shown.

## Behavior

Unset `TERM` environment variable is now treated as "not dumb" on Windows.

## Note

This depends on `is_dumb()` being exposed from the console crate. See console-rs/console#290 for the corresponding change adding a public `is_dumb()` function.

- console-rs/console#290
- #808

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -305,7 +305,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "indicatif"
-version = "0.18.5"
+version = "0.18.6"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,9 +135,9 @@ checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 
 [[package]]
 name = "console"
-version = "0.16.3"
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -135,9 +135,9 @@ checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 
 [[package]]
 name = "console"
-version = "0.16.3"
+version = "0.16.4"
```
djc, line 138: We'll need to update the requirements in `Cargo.toml` for this version.

djc:
(And please add a version bump commit here, too.)

djc (approved):

djc:
Published [0.18.6](https://github.com/console-rs/indicatif/releases/tag/0.18.6), thanks!

PR #827, opened 2026-07-28
Migrate in_memory terminal from vt100 to rio-vt

raphamorim:
Swaps the emulator behind the `in_memory` feature (`InMemoryTerm`) from [vt100](https://crates.io/crates/vt100) to [rio-vt](https://crates.io/crates/rio-vt), Rio's terminal engine. `InMemoryTermState` now holds a rio-vt `Crosswords` + `Processor` instead of a `vt100::Parser`; `write_str` feeds bytes through the parser, and `contents()` reads the grid back.

`rio-vt` is a safe-Rust, dependency-light terminal core (VT state machine, grid, parser), actively maintained, and it drops the transitive **`net2` (RUSTSEC-2020-0016, unmaintained)** vt100 pulled in — so `cargo deny check advisories` stays clean.

**All `render.rs` integration tests (37, ~126 `contents()` assertions) and the `in_memory` unit tests pass.** Two behavior notes:

1. **`contents()`** is now built from *physical* screen rows via `visible_rows()` rather than rio-vt's own plain formatter (which rejoins soft-wrapped rows into logical lines). This reproduces vt100's per-row output exactly — wrapped lines split across rows as before.
2. **Cursor pending-wrap**: the two `line_wrapping` unit-test cursor assertions were updated (`(0,5)`→`(0,4)`, `(1,5)`→`(1,4)`). rio-vt (alacritty-style) keeps the cursor on the last column with a pending-wrap flag (`col == width - 1`) rather than reporting `col == width`. This only affects the internal `cursor_pos` test helper; `InMemoryTerm`'s public output is unchanged.

`contents_formatted()` now delegates to rio-vt's `contents_formatted()` (no in-tree test covers it). Also pinned one pre-existing type-inference site in `iter.rs` (`u8::MAX.into()` → `usize::from(u8::MAX)`) that became ambiguous once rio-vt adds `impl PartialOrd<Column> for usize`.

- Benchmark vs vt100 & alacritty_terminal: https://github.com/raphamorim/rio-vt-benchmark
- Engine background: https://rioterm.com/blog/2026/07/27/rio-vt-and-librio

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
+name = "aho-corasick"
+version = "1.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


djc:
rio-vt has 488 recent downloads on crates.io, which means it's very unpopular (maybe due to it being very new).

With +1,112/-65 in `Cargo.lock`, I'm not sure how you can maintain that this is "dependency-light".

raphamorim:
Hey @djc thanks for the reply!

I did it as out of will to experiment the library please feel free to close if you feel like. Thank you for your time and reply, I appreciate!

PR #831, opened 2026-08-15
Fix HumanFloatCount printing "-0" for values that round to zero

vjymisal0:
`HumanFloatCount` extracts the sign separately from the integer digits so it stays out of the digit grouping (added in "Fix stray comma after minus sign in HumanFloatCount"), but nothing checks whether the value actually survives rounding at the requested precision.

Because `format!("{:.4}", -0.00001f64)` is `"-0.0000"` and the all-zero fraction is then trimmed away, the sign was printed in front of an all-zero result:

- `format!("{}", HumanFloatCount(-0.00001))` gave `"-0"`, expected `"0"`
- `format!("{:.0}", HumanFloatCount(-0.4))` gave `"-0"`, expected `"0"`

The fix suppresses the sign when the grouped digits are all zeros and no non-zero fractional part remains. Genuinely negative values (`-1.5`, `-0.5`, `-0.0001`) and `-inf` are unaffected.

Added `human_float_count_negative_rounding_to_zero`, which fails on main with `left: "0", right: "-0"` and passes with this change; it also covers the sign-preserving cases so the previous fix does not regress.

`cargo test --all-features`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo fmt --all -- --check` all pass locally.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -201,10 +201,17 @@ impl fmt::Display for HumanFloatCount {
         // Keep the optional sign out of the digit grouping, otherwise the '-'
         // is counted as a digit and a stray comma lands right after it
         // (e.g. -100 would render as "-,100").
-        let (sign, digits) = match int_part.strip_prefix('-') {
+        let (mut sign, digits) = match int_part.strip_prefix('-') {
             Some(digits) => ("-", digits),
             None => ("", int_part),
         };
+        let frac_trimmed = frac_part.trim_end_matches('0');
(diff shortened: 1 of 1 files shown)


vjymisal0:
Hello maintainers! Reaching out gently to see if someone might have time to review this small fix preventing `-0` formatting in `HumanFloatCount`. The test suite passes cleanly. Happy to tweak anything to match preferred formatting conventions!

djc (approved):

PR #832, opened 2026-08-18
Fix SGR attribute bleed from filled bar style into empty segment (#727)

### LINKED ISSUE #727: Some attributes are preserved when drawing progress bar
When I draw progress bar, some attributes are not cleared.

Let's take template `Processing [{wide_bar:.blink.red/dim.white}]`. In this case whole bar will blink instead of only filled part

youdie006:
Fixes #727.

A bar style like `{wide_bar:.blink.red/dim.white}` makes the empty part of the bar blink too.
The filled style is applied as a wrapper around the *whole* bar, with the empty part's style
nested inside it. Colours override across that boundary, but cumulative SGR attributes -- bold,
dim, blink, underline, reverse -- do not; they keep accumulating until the single reset at the
very end. So `\x1b[5m` from the filled style is still in effect while the empty `---` is drawn.

Before:

```
\x1b[31m\x1b[5m====>\x1b[37m\x1b[2m---\x1b[0m\x1b[0m
```

After:

```
\x1b[31m\x1b[5m====>\x1b[0m\x1b[37m\x1b[2m---\x1b[0m
```

This is not a console bug; each segment styled on its own resets correctly. It is the nesting in
`format_bar` and `BarDisplay::fmt`.

The fix styles the filled and empty segments as separate `StyledObject`s so each emits its own
reset. Most of the diff is mechanical: pulling the filled-portion rendering out into a
`FilledDisplay` struct so it can be wrapped. Two smaller pieces go with it -- when no empty style
is given the empty segment falls back to the filled style, so single-colour bars still colour the
whole bar, and the generic outer style-wrap is skipped for `bar`/`wide_bar` since the styling now
happens inside `format_bar`. Each segment still asks console whether colours are enabled and
whether it is writing to stderr, so nothing is hand-emitted.

Two existing byte-level tests change: `wide_element_style` and `multicolor_without_current_style`
now have a reset between the segments instead of two at the tail. For colour-only styles the
rendered result is the same; only the escape bytes moved.

The new test in `src/style.rs` fails on `main` and passes here. `cargo fmt --check`, `cargo
clippy --all-features -- -D warnings` and the 59 lib tests are clean.

I used AI assistance to prepare this, and I have read and verified everything in it myself.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -209,14 +209,7 @@ impl ProgressStyle {
             num: bg,
         };
 
-        // Style the filled and empty segments independently, so each emits its
-        // own reset. Wrapping the whole bar in the filled style with the empty
-        // style nested inside leaks cumulative SGR attributes (bold, dim, blink,
-        // underline, reverse) from the filled style into the empty segment: the
-        // empty style's codes override colors but attributes keep accumulating
-        // until the single outer reset (see #727). When no explicit empty style
-        // is given, fall back to the filled style so single-color bars keep
-        // coloring the empty segment as before.
+        // Each segment is styled on its own so it emits its own reset.
         BarDisplay {
             filled: style.unwrap_or(&Style::new()).apply_to(FilledDisplay {
                 chars: &self.progress_chars,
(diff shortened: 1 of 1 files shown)


eirnym (approved):

djc:
Please rewrite the PR description in your own words, and edit down the overly verbose comments in the code. I'm not opposed to LLM usage but I expect you to take responsibility and communicate with me yourself.

youdie006:
Fair, thanks for saying so directly.

Rewritten the description, and cut the code comments down: they were 11 of 79 added lines, now 5 of 73, one line each at the site rather than a paragraph restating the PR body. The only prose left is a one-line doc comment on `FilledDisplay` and a one-line note on the regression test.

No behaviour change from that commit; `cargo fmt --check`, `cargo clippy --all-features -- -D warnings` and the 59 lib tests still pass.

To be explicit about the process, since that is the part you are asking about: I use AI assistance to find and draft these, and I read and run everything before it goes out. The description above is mine. If anything in it does not hold up under review, that is on me to answer, and I will.

djc (approved):

djc:
This looks okay, thanks.

PR #833, opened 2026-08-18
Add compact short_pos/short_len/short_per_sec template keys (SI prefixes, 3 sig figs)

### LINKED ISSUE #637: Feature Req.: Human readable `per_sec` rate

ChrisJr404:
Closes #637.

For high-throughput bars, `{per_sec}` (and the raw `{pos}`/`{len}`) can print
huge numbers that jitter in width from frame to frame. This adds a compact,
fixed-width alternative using SI prefixes and three significant digits, matching
the format sketched in the issue discussion (`143`, `7.21k`, `39.4M`).

**What's added**

- `HumanShortCount(pub f64)` in `format.rs` — a new public formatter, alongside
  the existing `HumanCount` / `HumanFloatCount`. It reuses `unit_prefix` (already
  a dependency, used by the byte formatters) for the decimal prefix, keeps three
  significant figures, and trims trailing zeros so `1000` renders as `1k` rather
  than `1.00k`.
- Three template keys wired through `style.rs`:
  - `short_pos`, `short_len` — compact counterparts to `human_pos`/`human_len`
  - `short_per_sec` — compact counterpart to `per_sec` (appends `/s`)
- Docs for the new keys in the template-key list, a re-export from the crate
  root, and a doctest on the struct.

**Behavior is opt-in** — no default template or existing key changes; you only
get the new format if you reference one of the new keys.

**Examples**

```
143     -> 143
999     -> 999
1000    -> 1k
7210    -> 7.21k
39400000 -> 39.4M
```

Tested with a unit test over `HumanShortCount`, the struct doctest, and a
`format_state` render test exercising `{short_pos}/{short_len}`. `cargo test`,
`cargo fmt --check`, and `cargo clippy --all-targets` are all clean.

I went with new keys (mirroring `human_pos`/`human_len`) rather than a `:short`
format suffix on the parser since it's smaller and matches the existing
convention, but happy to adjust naming or scope if you'd prefer something else.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -71,6 +71,28 @@ pub struct HumanCount(pub u64);
 #[derive(Debug)]
 pub struct HumanFloatCount(pub f64);
 
+/// Formats a count compactly using SI (decimal) prefixes and three
+/// significant digits.
+///
+/// Unlike [`HumanCount`], which keeps every digit and groups them with commas,
+/// this renders the order of magnitude in a fixed width. It is handy for rates
+/// or large counts where the exact value matters less than the magnitude, e.g.
+/// a `{per_sec}`-style field that would otherwise jitter in width.
+///
+/// # Examples
+/// ```rust
+/// # use indicatif::HumanShortCount;
+/// assert_eq!("0",     format!("{}", HumanShortCount(0.0)));
+/// assert_eq!("143",   format!("{}", HumanShortCount(143.0)));
+/// assert_eq!("999",   format!("{}", HumanShortCount(999.0)));
+/// assert_eq!("1k",    format!("{}", HumanShortCount(1_000.0)));
+/// assert_eq!("7.21k", format!("{}", HumanShortCount(7_210.0)));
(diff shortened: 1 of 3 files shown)


djc:
I don't think I want to add more niche template keys.

You can either implement these downstream by implementing the [`ProgressTracker`](https://docs.rs/indicatif/latest/indicatif/style/trait.ProgressTracker.html) trait, or we can make improvements to the default implementations for `pos`/`len`/`per_sec` that are improvements for everyone/most users.

(Also don't copy/paste LLM-generated text into PR descriptions/comments if you want me to engage.)

ChrisJr404:
That is fair. I would rather not pile on niche template keys most users will not reach for either. The ProgressTracker trait covers this downstream, so I will close it out. Thanks for taking a look.

PR #835, opened 2026-08-26
tests: replace useless format!() with .to_string()

tangaac:
## Summary

Newer clippy versions flag `format!("{}", expr)` as `useless_format`
when the expression already implements `Display`. With `-D warnings` in
CI, this breaks the build.

Replace with `.to_string()` in `tests/render.rs`.

## Verification

`cargo clippy --all-targets --all-features -- -D warnings` passes.


djc (approved):
Thanks!

PR #836, opened 2026-08-30
fix: avoid underflow in draw_to_term when the last line is wider than the terminal

lenamonj:
`DrawState::draw_to_term` computes the trailing filler as `term_width - metrics.last_line_width`. When the last visual row is wider than the terminal - a terminal reporting zero columns, or a two-column character with one column left - the subtraction underflows: a panic (`attempt to subtract with overflow`) in debug builds, and in release a wrap to ~1.8e19 that aborts in `" ".repeat()` with a capacity overflow.

The cursor is already at or past the end of the row in that state, so no filler is due: `saturating_sub` is the whole fix.

The added test drives `draw_to_term` through a minimal `TermLike` at width 0 and at width 1 with a two-column CJK character; both panic before this change and pass after it, and the test also asserts no filler run ever exceeds the terminal width.

### DIFF
--- a/src/draw_target.rs
+++ b/src/draw_target.rs
@@ -576,7 +576,14 @@ impl DrawState {
 
             // clear the line and keep the cursor on the right terminal side so that
             // future writes/prints will happen on the next line
-            let line_filler = term_width - metrics.last_line_width;
+            //
+            // Saturating: `last_line_width` can exceed `term_width` when the last
+            // visual row cannot fit even one character - a zero-width terminal, or a
+            // two-column character on a one-column terminal. The cursor is then
+            // already at or past the end of the row, so no filler is due. A plain
+            // subtraction underflows there, which panics in debug and wraps to a
+            // ~1.8e19 `" ".repeat()` allocation in release.
+            let line_filler = term_width.saturating_sub(metrics.last_line_width);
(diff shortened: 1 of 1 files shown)


src/draw_target.rs
```diff
@@ -576,7 +576,14 @@ impl DrawState {
...
             // clear the line and keep the cursor on the right terminal side so that
             // future writes/prints will happen on the next line
-            let line_filler = term_width - metrics.last_line_width;
+            //
+            // Saturating: `last_line_width` can exceed `term_width` when the last
+            // visual row cannot fit even one character - a zero-width terminal, or a
+            // two-column character on a one-column terminal. The cursor is then
+            // already at or past the end of the row, so no filler is due. A plain
+            // subtraction underflows there, which panics in debug and wraps to a
+            // ~1.8e19 `" ".repeat()` allocation in release.
```
djc: This reads like slop. Please tighten it up (or maybe remove it entirely).

djc (approved):

djc:
Thanks!

src/draw_target.rs
lenamonj: Removed in d4c7d81, and trimmed the test comment to two lines.

djc (approved):

djc:
Thanks!

